PR #1, opened 2017-09-09
OSS!

mbrandonw:


mbrandonw:
> Yes! Should we add CI before merging?

will do in a separate PR!

PR #5, opened 2017-09-09
Redirect related domains

mbrandonw:
redirects hosts that are not in an allow list to `www.pointfree.co`

### DIFF
--- a/Sources/PointFree/Routes.swift
+++ b/Sources/PointFree/Routes.swift
@@ -6,7 +6,8 @@ import Prelude
 import Styleguide
 
 public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data?> =
-  route(router: PointFree.router)
+  redirectUnrelatedDomains
+    <<< route(router: PointFree.router)
     <| (render(conn:) >>> perform)
 
 public enum Route {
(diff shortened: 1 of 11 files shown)


Sources/PointFree/Routes.swift
```diff
@@ -55,6 +56,38 @@ private func route<I, A, Route>(
     }
 }
 
+private let relatedDomains = [
+  "pointfree.co",
```
stephencelis: What if we did the reverse and redirect anything that doesn't match to www.pointfree.co?

``` swift
private let hosts = [
  "www.pointfree.co",
  "localhost",
  "0.0.0.0",
  "127.0.0.1",
]
```

mbrandonw: Good idea! Will do

mbrandonw:
@stephencelis ok done! give it another look!

Tests/PointFreeTests/LaunchSignupTests.swift
```diff
@@ -19,7 +19,7 @@ class LaunchSignupTests: TestCase {
   }
 
   func testHome_SuccessfulSignup() {
-    let request = URLRequest(url: URL(string: "/?success=true")!)
+    let request = URLRequest(url: URL(string: "http://localhost/?success=true")!)
```
stephencelis: Might be good to hardcode a port in here to be thorough! Otherwise good!

PR #7, opened 2017-09-10
redirect to https

mbrandonw:


Sources/PointFree/Routes.swift
```diff
@@ -6,9 +6,10 @@ import Prelude
 import Styleguide
 
 public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data?> =
-  redirectUnrelatedDomains
+  requireHttps
+    <<< redirectUnrelatedDomains
     <<< route(router: PointFree.router)
-    <| (render(conn:) >>> perform)
```
mbrandonw: I just moved `perform` down into `render`

Sources/PointFree/Routes.swift
```diff
@@ -86,11 +90,47 @@ private func redirectUnrelatedDomains<A>(
...
+// TODO: move to httppipeline?
+private func makeHttps(url: URL) -> URL? {
+  var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+  components?.scheme = "https"
+  return components?.url
+}
+
+// TODO: move to prelude
+extension Optional {
+  fileprivate func filterOptional(_ p: (Wrapped) -> Bool) -> Optional {
```
mbrandonw: we should prob move this to prelude at some point. it's an optional specific version of `mfilter: (a -> Bool) -> m a -> m a`

stephencelis: Definitely! I was tempted to just call it `filter` since we have it on `Event` and `Array` that way, but `filterOptional` was a lot easier to read/understand here on GitHub.

Sources/PointFree/Routes.swift
```diff
@@ -86,11 +90,47 @@ private func redirectUnrelatedDomains<A>(
...
+            |> end
+        }
+        ?? middleware(conn)
     }
 }
 
+// TODO: move to httppipeline?
+private func makeHttps(url: URL) -> URL? {
+  var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+  components?.scheme = "https"
```
stephencelis: This'd be fun to lens/prism about.

Sources/PointFree/Routes.swift
mbrandonw: yeah def, I'm down to do just `filter` too.

or to align with `mfilter` maybe it should be `optionalFilter`.

Sources/PointFree/Routes.swift
mbrandonw: oh jeez, totally! check it here! https://github.com/pointfreeco/pointfreeco/pull/7/commits/30f8d780f38eeed001bfcce168c59fe1b6504538

Sources/PointFree/Routes.swift
```diff
@@ -86,11 +91,56 @@ private func redirectUnrelatedDomains<A>(
...
+
+// TODO: move to HttpPipeline
+private func requireHttps<A>(
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, A, Data?>
+  )
+  -> Middleware<StatusLineOpen, ResponseEnded, A, Data?> {
+
+    return { conn in
+      conn.request.url
+        .filterOptional { (url: URL) -> Bool in
```
mbrandonw: oh also this type sig was needed to help the compiler...

PR #8, opened 2017-09-11
proper redirects for heroku

mbrandonw:
it seems that checking the url scheme for `https` cannot be trusted in heroku and instead we must check a header: https://jaketrent.com/post/https-redirect-node-heroku/

### DIFF
--- a/Sources/PointFree/Routes.swift
+++ b/Sources/PointFree/Routes.swift
@@ -7,7 +7,7 @@ import Prelude
 import Styleguide
 
 public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data?> =
-  requireHttps
+  requireHerokuHttps
     <<< redirectUnrelatedDomains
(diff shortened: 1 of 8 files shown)


Sources/PointFree/Routes.swift
```diff
@@ -91,28 +92,29 @@ private func redirectUnrelatedDomains<A>(
...
-private let httpAllowedHosts = [
+private let allowedInsecureHosts: [String] = [
   "127.0.0.1",
   "0.0.0.0",
   "localhost"
 ]
 
 // TODO: move to HttpPipeline
-private func requireHttps<A>(
+private func requireHerokuHttps<A>(
```
mbrandonw: if we move this to `HttpPipeline` I think it could be made into general redirect middleware that takes a predicate `(URLRequest) -> Bool` to determine if to redirect, and a request rewriting function `(URLRequest) -> URL` for the redirect. it should also handle any non-`GET` requests to redirect to a 400 page.

PR #9, opened 2017-09-11
GA!

mbrandonw:


Sources/PointFree/LaunchSignup/Views.swift
```diff
@@ -233,8 +233,8 @@ private let view: View<Bool?> = View { success in
           ),
           body(
             success == .some(true)
-              ? [ successSectionNode, headerNode ]
-              : [ headerNode, defaultSectionNode ]
+              ? [ successSectionNode, headerNode, googleAnalytics ]
+              : [ headerNode, defaultSectionNode, googleAnalytics ]
```
mbrandonw: this can of course be better but just knocking stuff out!

PR #15, opened 2017-09-13
Added twitter avatar meta data.

mbrandonw:
uploaded an image to s3 and updated the twitter meta tags


Sources/PointFree/MetaLayout.swift
```diff
@@ -6,16 +6,20 @@ import Prelude
 
 public struct Metadata<A> {
   public let description: String?
+  public let image: String?
   public let rest: A
   public let title: String?
+  public let twitterCard: String?
```
mbrandonw: we'll be able to factor this struct into a better type that properly enumerates all the options, but this will do for now

PR #16, opened 2017-09-13
Fix Styleguide playground

stephencelis:
This playground was out of date! It should build now.

Would be nice to automate testing around these things, but it's not an easy thing by default.

### DIFF
--- a/Styleguide.playground/Contents.swift
+++ b/Styleguide.playground/Contents.swift
@@ -1,53 +1,55 @@
+import Css
+import CssReset
+import Html
+import HtmlCssSupport
+import PlaygroundSupport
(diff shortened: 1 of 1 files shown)


Styleguide.playground/Contents.swift
```diff
@@ -68,56 +70,57 @@ let styleguideHtml = html(
...
-
-    div(
+    body(
       [
-        button(
-          [ HTML.class("btn") ],
-          [ "I’m a button!" ]
+        div(
+          [
+            Html.button(
```
stephencelis: This qualification was a bit surprising. Was clashing with `Css`.

PR #23, opened 2017-10-20
Update swift web to use new router

mbrandonw:
Updates the lib to use the new router in `swift-web`. pretty straight forward.


Tests/PointFreeTests/AuthTests.swift
```diff
@@ -16,7 +16,7 @@ class AuthTests: TestCase {
     let conn = connection(from: request)
     let result = conn |> siteMiddleware
 
-    assertSnapshot(matching: result)
+    assertSnapshot(matching: result.perform())
```
mbrandonw: once we do the snapshotting function (as opposed to protocol) we could prob get rid of these...

PR #24, opened 2017-10-21
Move env vars to .env

mbrandonw:
This will first try to load env vars from `.env` locally, and if that fails fallback to `ProcessInfo.processInfo.environment`. It should allow us to control the environment locally while keeping our true secrets only on the server.

### DIFF
--- a/Sources/PointFree/EnvVars.swift
+++ b/Sources/PointFree/EnvVars.swift
@@ -7,8 +7,7 @@ enum EnvVars {
       .appendingPathComponent(".env")
 
     let localEnvVars = (try? Data(contentsOf: envFilePath))
-      .flatMap { try? JSONSerialization.jsonObject(with: $0) }
-      .flatMap { $0 as? [String: String] }
+      .flatMap { try? JSONDecoder().decode([String: String].self, from: $0) }
 
     return localEnvVars ?? ProcessInfo.processInfo.environment
   }()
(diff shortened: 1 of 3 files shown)


Sources/PointFree/EnvVars.swift
```diff
@@ -1,16 +1,19 @@
...
-    return ProcessInfo.processInfo.environment
-  }
+  private static let env: [String: String] = {
+    let envFilePath = URL(fileURLWithPath: #file)
+      .deletingLastPathComponent()
+      .appendingPathComponent(".env")
 
-  static let appSecret = env["APP_SECRET"] ?? "deadbeefdeadbeefdeadbeefdeadbeef"
+    let localEnvVars = (try? Data(contentsOf: envFilePath))
+      .flatMap { try? JSONSerialization.jsonObject(with: $0) }
```
stephencelis: Could use the new codable stuff here:

``` swift
JSONDecoder().decode([String: String].self, from: $0)
```

stephencelis:
CI failures aren't loading for me :/

Sources/PointFree/EnvVars.swift
mbrandonw: oh jeez, duh! why do I keep reaching for `JSONSerialization`...

mbrandonw:
tests are passing and addressed yer feedback

PR #25, opened 2017-10-21
Fun test stub cleanup using function composition

stephencelis:
Not sure if you prefer the explicitness of the mock names. If so we could still simplify the implementations, _e.g._:

``` swift
func mockFetchAuthToken(
  result: Either<Prelude.Unit, GitHubAccessToken>
  )
  -> (String)
  -> EitherIO<Prelude.Unit, GitHubAccessToken> {
    return result |> pure >>> const
}
```

They'd both have the same implementation, though!

### DIFF
--- a/Tests/PointFreeTests/TestCase.swift
+++ b/Tests/PointFreeTests/TestCase.swift
@@ -1,3 +1,4 @@
+import Either
 @testable import PointFree
 import Prelude
 import SnapshotTesting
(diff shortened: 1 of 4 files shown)


mbrandonw:
I just pushed a fix for tests on this branch.

Tests/PointFreeTests/TestCase.swift
```diff
@@ -9,11 +10,9 @@ class TestCase: XCTestCase {
 
     AppEnvironment.push(
       env: .init(
-        airtableStuff: mockCreateRow(result: .right(unit)),
-        fetchAuthToken: mockFetchAuthToken(result: .right(.init(accessToken: "deadbeef"))),
-        fetchGitHubUser: mockFetchGithubUser(
-          result: .right(.init(email: "hello@pointfree.co", id: 1, name: "Blob"))
-        )
+        airtableStuff: const(const(pure(unit))),
```
mbrandonw: lol, `mockCreateRow` was curried, so gotta `const >>> const` to stub it out. whatev, we will prob find better ways for all this some day.

stephencelis:
Oh weird I swear I ran tests.

PR #29, opened 2017-10-29
Error Handling Spike

stephencelis:
This shouldn't get merged in before we figure some things out, but wanted to get the discussion going about one way of handling errors.

This PR allows endpoints to lift their response data into `Either` to return potential errors. It's a little messy/noisy, though, and I'd like to figure out a better way of composing and working with these nested types.

### DIFF
--- a/Sources/PointFree/Auth.swift
+++ b/Sources/PointFree/Auth.swift
@@ -17,7 +17,7 @@ let githubCallbackResponse =
 
 /// Redirects to GitHub authorization and attaches the redirect specified in the connection data.
 let loginResponse: Middleware<StatusLineOpen, ResponseEnded, String?, Data> =
-  { $0 |> redirect(to: githubAuthorizationUrl(withRedirect: $0.data)) }
+  { $0 |> redirect(to: gitHubAuthorizationUrl(withRedirect: $0.data)) }
 
 let logoutResponse: (Conn<StatusLineOpen, Prelude.Unit>) -> IO<Conn<ResponseEnded, Data>> =
   redirect(
@@ -68,7 +68,7 @@ private func readGitHubSessionCookieMiddleware(
         conn.request.cookies[githubSessionCookieName]
           .flatMap { ResponseHeader.verifiedValue(signedCookieValue: $0, secret: EnvVars.appSecret) }
           .map(Either.right)
-          ?? Either.left(unit)
+          ?? .left(unit)
       )
     )
 }
@@ -95,7 +95,7 @@ private func writeGitHubSessionCookieMiddleware(
 private func authTokenMiddleware(
   _ conn: Conn<StatusLineOpen, (code: String, redirect: String?)>
   )
-  -> IO<Conn<ResponseEnded, Data>> {
+  -> IO<Conn<ResponseEnded, Either<Error, Data>>> {
 
     return AppEnvironment.current.fetchAuthToken(conn.data.code)
       .flatMap { token in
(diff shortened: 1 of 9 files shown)


Sources/PointFree/Auth.swift
```diff
@@ -95,34 +95,34 @@ private func writeGitHubSessionCookieMiddleware(
...
-            |> redirect(to: path(to: .secretHome))
-
+      .flatMap { gitHubUserEnvelope in
+
+        switch gitHubUserEnvelope {
+        case let .left(error):
+          return conn.map(const(Either<Error, Data>.left(error)))
+            |> writeStatus(.internalServerError)
+            >-> end
+            >>> map(map(pure))
```
stephencelis: This is the ugly part. Our middleware combinators always expect to end on `Data`, so both of these branches have to move the monad index forward and re-wrap the result in `Either`.

Sources/PointFree/GitHubApi.swift
```diff
@@ -39,43 +39,58 @@ func fetchAuthToken(forCode code: String) -> EitherIO<Prelude.Unit, GitHubAccess
...
+    do {
+      return .right(try f())
+    } catch let error {
+      return .left(error)
+    }
+  }
+}
+
+// TODO: Move to PreludeFoundation?
+public func jsonDataTask<A>(_ session: URLSession)
```
stephencelis: Some cleanup so that we're not dealing with some of the messier parts of `URLSession`.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -13,35 +14,69 @@ public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Uni
...
+      return conn.map(const((code, redirect)))
+        |> githubCallbackResponse
 
-  case let .home(signedUpSuccessfully):
-    return conn.map(const(signedUpSuccessfully))
-      |> homeResponse
+    case let .home(signedUpSuccessfully):
+      return conn.map(const(signedUpSuccessfully))
+        |> homeResponse
+        >>> map(map(pure))
```
stephencelis: Cool to be able to look at our main routing function and see which endpoints are pure and which may have errors!

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -13,35 +14,69 @@ public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Uni
...
-      |> secretHomeResponse
-  }
+    case .secretHome:
+      return conn.map(const(unit))
+        |> secretHomeResponse
+        >>> map(map(pure))
+    }
+}
+
+public func handleErrors<A>(
```
stephencelis: This could be generalized with a higher-order error handling function.

Sources/PointFree/Auth.swift
```diff
@@ -95,34 +95,34 @@ private func writeGitHubSessionCookieMiddleware(
...
-            |> redirect(to: path(to: .secretHome))
-
+      .flatMap { gitHubUserEnvelope in
+
+        switch gitHubUserEnvelope {
+        case let .left(error):
+          return conn.map(const(Either<Error, Data>.left(error)))
+            |> writeStatus(.internalServerError)
+            >-> end
+            >>> map(map(const(.left(error))))
```
stephencelis: This is the ugly part. Our middleware combinators always expect to end on `Data`, so both of these branches have to move the monad index forward and re-wrap the result in `Either`.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -13,35 +14,69 @@ public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Uni
       realm: "Point-Free",
       protect: isProtected
     )
+    <<< handleErrors
```
mbrandonw: v cool!

stephencelis:
This was fun to think about but not to be merged! We may revisit a unification of error handling in the future...and we may not!

PR #31, opened 2017-11-12
Move EnvVars to the Environment

mbrandonw:
### What

Moves the `EnvVars` values to a proper struct, and put it inside the `Environment` so that we can control it for tests. Also added a make task for pull configs from heroku.


Sources/PointFree/Auth.swift
```diff
@@ -121,21 +121,17 @@ private func authTokenMiddleware(
...
     "redirect_uri": url(to: .githubCallback(code: "", redirect: redirect))
   ]
 
-  let query = params.map { key, value in
-    "\(key)=" + (value.addingPercentEncoding(withAllowedCharacters: .urlQueryParamAllowed) ?? "")
-    }
-    .joined(separator: "&")
-
-  return "https://github.com/login/oauth/authorize?\(query)"
+  return "https://github.com/login/oauth/authorize?\(urlFormEncode(value: params))"
```
mbrandonw: just cleaning this up now that we have a url encoder

Sources/PointFree/EnvVars.swift
```diff
@@ -1,43 +1,116 @@
...
-
-  enum Airtable {
-    static let base1 = env["AIRTABLE_BASE_1"] ?? "deadbeef-base-1"
-    static let base2 = env["AIRTABLE_BASE_2"] ?? "deadbeef-base-2"
-    static let base3 = env["AIRTABLE_BASE_3"] ?? "deadbeef-base-3"
-    static let bearer = env["AIRTABLE_BEARER"] ?? "deadbeef-bearer"
-  }
+    return (try? JSONSerialization.data(withJSONObject: envVars))
+      .flatMap { try? JSONDecoder().decode(EnvVars.self, from: $0) }
+      ?? EnvVars()
```
stephencelis: This dance is a lil unfortunate.

mbrandonw: I guess another way to do this is to take the json config file that heroku gives us an then codegen EnvVars from it. then we'd have the compile time guarantees

PR #32, opened 2017-11-12
Fire and forget the send email work.

mbrandonw:
When I thought this _didn't_ work I was just confused how breakpoints work lol. So this does work!

### DIFF
--- a/Sources/PointFree/LaunchSignup/LaunchSignup.swift
+++ b/Sources/PointFree/LaunchSignup/LaunchSignup.swift
@@ -33,14 +33,16 @@ func notifyUsOfNewSignup<I>(_ conn: Conn<I, String>) -> IO<Conn<I, String>> {
   return IO {
 
     // Fire-and-forget to notify us that someone signed up
-    _ = sendEmail(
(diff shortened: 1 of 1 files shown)


Sources/PointFree/LaunchSignup/LaunchSignup.swift
```diff
@@ -33,14 +33,16 @@ func notifyUsOfNewSignup<I>(_ conn: Conn<I, String>) -> IO<Conn<I, String>> {
...
+        from: "Point-Free <support@pointfree.co>",
+        to: ["brandon@pointfree.co", "stephen@pointfree.co"],
+        subject: "New signup for Point-Free!",
+        content: inj2(notifyUsView.view(conn.data))
+        )
+        .run
       )
-      .run
-      .perform()
+      .run({ _ in })
```
mbrandonw: lil funny to see this as `.run).run({_ in })`. we could maybe have a convenience `run` on parallel specifically for fire-and-forget that doesnt take a callback. then it could maybe even be something like:

```swift 
sendEmail(...)
  |> parallelize 
  |> runAndDiscard
```

PR #33, opened 2017-11-13
Database Signup

stephencelis:
Basics for signup! Saving GitHub auth details for login!


Sources/PointFree/Database.swift
```diff
@@ -0,0 +1,165 @@
...
+  return conn.flatMap { conn in
+    .wrap { try conn.execute(query, representable) }
+  }
+}
+
+public func createSubscription(from stripeSubscription: StripeSubscription, for user: User)
+  -> EitherIO<Error, Prelude.Unit> {
+    return execute(
+      """
+      INSERT INTO "subscriptions" ("stripe_subscription_id", "user_id")
```
stephencelis: Will be fun to explore some type-safe query generation! (Maybe with `Codable`?)

Sources/PointFree/Database.swift
```diff
@@ -0,0 +1,169 @@
...
+      """
+    )))
+    .flatMap(const(execute(
+      """
+      CREATE EXTENSION IF NOT EXISTS "citext" WITH SCHEMA "public"
+      """
+    )))
+    .flatMap(const(execute(
+      """
+      CREATE TABLE IF NOT EXISTS "users" (
```
stephencelis: We'll want proper migrations (using a schema version) in the future, but this works for now!

Sources/PointFree/Auth.swift
```diff
@@ -105,12 +111,17 @@ private func authTokenMiddleware(
         AppEnvironment.current.fetchGitHubUser(token)
           .map { user in GitHubUserEnvelope(accessToken: token, gitHubUser: user) }
       }
+      .flatMap { env in
+        AppEnvironment.current.fetchUser(env.accessToken).bimap(const(unit), const(env))
+          <|> AppEnvironment.current.createUser(env).bimap(const(unit), const(env))
```
stephencelis: This was fun but maybe unnecessary with the upsert!

PR #34, opened 2017-11-28
Episode(s) page, and lots of stubs

mbrandonw:
What started as @amotion and me jamming on the episode page has turned into a whole bunch of crap! it's too much to review, but I'll add some notes...

screenshot 2017-11-27 21 05 57

screenshot 2017-11-27 21 06 00


Sources/PointFree/About.swift
```diff
@@ -0,0 +1,39 @@
...
+let aboutResponse: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
+  writeStatus(.ok)
+    >-> respond(aboutView)
+
+private let aboutView = View<Prelude.Unit> { _ in
+  document([
+    html([
+      head([
+        style(renderedNormalizeCss),
+        style(styleguide),
```
mbrandonw: need to come up with nice, reusable, composable layouts for abstracting away this stuff. I think there's a few directions we can go, and not sure of the best one...

Sources/PointFree/Auth.swift
```diff
@@ -35,6 +35,11 @@ private let secretHomeView = View<Either<Prelude.Unit, GitHubUserEnvelope>> { da
       )
       ]),
 
+    a(
+      [href(path(to: .episodes(tag: nil)))],
```
mbrandonw: these type safe url helpers are so nice to use

Sources/PointFree/Episode/Episode.swift
```diff
@@ -0,0 +1,212 @@
...
+
+private func responseEpisode(_ conn: Conn<HeadersOpen, Episode?>) -> IO<Conn<ResponseEnded, Data>> {
+
+  switch conn.data {
+  case .none:
+    return conn.map(const(unit))
+      |> respond(notFoundView)
+  case let .some(ep):
+    return conn.map(const(ep))
+      |> respond(episodeView.map(addHighlightJs >>> addGoogleAnalytics))
```
mbrandonw: right now views can opt into adding Highlight.js and google analytics by `map`ing with these lil helper functions. I like this a lot and think we should build more off of it.

Sources/PointFree/Episode/Episode.swift
```diff
@@ -0,0 +1,212 @@
...
+}
+
+public let episodeView = View<Episode> { ep in
+  document([
+    html([
+      head([
+        style(renderedNormalizeCss),
+        style(styleguide),
+        title("Episode #\(ep.sequence): \(ep.title)"),
+        meta(viewport: .width(.deviceWidth), .initialScale(1)),
```
mbrandonw: this is also something that needs to go into every view...

Sources/PointFree/Episode/Episodes.swift
```diff
@@ -0,0 +1,147 @@
...
+        p([`class`([Class.pf.type.callout])], [.text(encode(ep.blurb))]),
+
+        div(pillTagsView.view(ep.tags))
+        ])
+      ])
+    ])
+}
+
+extension Array {
+  func sorted<A: Comparable>(by f: (Element) -> A) -> Array {
```
mbrandonw: @stephencelis we got anything like this in the prelude?

Sources/PointFree/Pricing.swift
```diff
@@ -0,0 +1,190 @@
+import Css
```
mbrandonw: most of the stuff in this file can be ignored. it was mostly me experimenting with doing purely css tabs, and kinda snowballed from there. we'll come back to this soon.

Sources/PointFree/Episode/Episodes.swift
stephencelis: Our Prelude sorting story isn't great yet. We have a `Comparator` protocol that doesn't bridge to `Comparable` (and maybe it shouldn't exist at all). With a bridge, though, we could define:

``` swift
func comparing<A, B: Comparable>(_ f: @escaping (A) -> B) -> (A, A) -> Comparator {
  return { compare(f($0), f($1)) } // I think `compare` is curried right now, but maybe it shouldn't be?
}
```

Sources/PointFree/Episode/Episode.swift
```diff
@@ -0,0 +1,216 @@
...
+
+private func fetchEpisode(_ conn: Conn<StatusLineOpen, Either<String, Int>>) -> IO<Conn<HeadersOpen, Episode?>> {
+
+  let possibleEpisode = episodes.first(where: {
+    conn.data.left == .some($0.slug)
+      || conn.data.right == .some($0.id)
+  })
+
+  return conn.map(const(possibleEpisode))
+    |> writeStatus(possibleEpisode == nil ? .notFound : .ok)
```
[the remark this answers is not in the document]
mbrandonw: > Not a blocker but found it a lil tough to read writing status here and then responding below.

yeah I agree, hoping a better error handling sitch will help out

PR #35, opened 2017-11-28
Remove .env logic from PointFree

stephencelis:
We don't really need it here. We want it in pointfree-server.


stephencelis:
See: https://github.com/pointfreeco/pointfreeco-server/pull/4

PR #37, opened 2017-11-30
Testing on Linux

mbrandonw:
Opening this PR a bit early cause there's still two failing Linux tests, but just wanna get this out there.

This gets all of our tests running on linux, and I included the change to make the GH auth code optional in the route.

The tests that are failing have to do with encryption hashes differing for some reason, and query parameters being rendered in the wrong order :/


Sources/PointFree/Routes.swift
```diff
@@ -29,7 +29,7 @@ private let routers: [Router<Route>] = [
 
   Route.iso.githubCallback
     <¢> get %> lit("github-auth")
-    %> queryParam("code", .string) <%> queryParam("redirect", opt(.string))
+    %> queryParam("code", opt(.string)) <%> queryParam("redirect", opt(.string))
```
mbrandonw: I went ahead and made this change, and updated the code to handle it... some interesting things popped out.

Sources/PointFree/Auth.swift
```diff
@@ -147,7 +180,7 @@ private func githubAuthorizationUrl(withRedirect redirect: String?) -> String {
   let params: [String: String] = [
     "scope": "user:email",
     "client_id": AppEnvironment.current.envVars.gitHub.clientId,
-    "redirect_uri": url(to: .githubCallback(code: "", redirect: redirect))
+    "redirect_uri": url(to: .githubCallback(code: nil, redirect: redirect))
```
mbrandonw: now we can use `nil` for the redirect and it will exclude the `code` from the query params.

Tests/PointFreeTests/__Snapshots__/AuthTests/testLogin.1.Conn.txt
```diff
@@ -9,4 +9,4 @@
 
 ▿ Response
   Status 302 FOUND
-  Location: https://github.com/login/oauth/authorize?scope=user:email&redirect_uri=http://localhost:8080/github-auth%3Fcode%3D&client_id=deadbeef-client-id
\ No newline at end of file
+  Location: https://github.com/login/oauth/authorize?scope=user:email&redirect_uri=http://localhost:8080/github-auth&client_id=deadbeef-client-id
```
mbrandonw: as seen here!

Tests/PointFreeTests/EnvVarTests.swift
```diff
@@ -3,14 +3,6 @@ import SnapshotTesting
 import XCTest
 
 class EnvVarTests: TestCase {
-  func testEncoding() {
```
mbrandonw: not a good test, so deleting

Sources/PointFree/Auth.swift
```diff
@@ -5,14 +5,47 @@ import HttpPipeline
...
 
 let secretHomeResponse: (Conn<StatusLineOpen, Prelude.Unit>) -> IO<Conn<ResponseEnded, Data>> =
   writeStatus(.ok)
     >-> readGitHubSessionCookieMiddleware
     >-> respond(secretHomeView)
 
 let githubCallbackResponse =
-  authTokenMiddleware
+  extractGitHubAuthCode
+    <| authTokenMiddleware
```
mbrandonw: this is how we alter the `githubCallbackResponse` middleware to handle the optional code. we use a middleware transformer to convert the optional `String` to a non-optional, and in the case of `nil` we render a 400 bad request page.

Sources/PointFree/Auth.swift
```diff
@@ -5,14 +5,47 @@ import HttpPipeline
...
+
+/// Middleware transformer to convert the optional GitHub code to a non-optional. In the `nil` case we show
+/// a 400 Bad Request page.
+private func extractGitHubAuthCode(
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, (code: String, redirect: String?), Data>
+  )
+  -> Middleware<StatusLineOpen, ResponseEnded, (code: String?, redirect: String?), Data> {
```
mbrandonw: this is the transformer...

Sources/PointFree/Auth.swift
```diff
@@ -5,14 +5,47 @@ import HttpPipeline
...
+    
+    return { conn in
+      conn.data.code
+        .map { (code: $0, redirect: conn.data.redirect) }
+        .map { conn.map(const($0)) }
+        .map(middleware)
+        ?? (conn |> const(unit) >¢< missingGitHubAuthCodeMiddleware)
```
mbrandonw: I defined `contramap` on middleware so that I could plug in `missingGitHubAuthCodeMiddleware`, which is defined on `Unit`, into middleware that wants `(String, String?)`.

Sources/PointFree/Auth.swift
```diff
@@ -5,14 +5,47 @@ import HttpPipeline
...
+  )
+  -> Middleware<I, J, A, B> {
+
+    return map(lhs) >>> rhs
+}
+
+/// Middleware to run when the GitHub auth code is missing.
+private let missingGitHubAuthCodeMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
+  writeStatus(.badRequest)
+    >-> respond(text: "GitHub code wasn't found :(")
```
mbrandonw: can make this response better later. prob some kind of error specific view or somethin

mbrandonw:
Looks like we only have one failure, this snapshot:

```diff 
▿ Response
   Status 302 FOUND
-  Location: https://github.com/login/oauth/authorize?scope=user:email&redirect_uri=http://localhost:8080/github-auth&client_id=deadbeef-client-id
+  Location: https://github.com/login/oauth/authorize?redirect_uri=http://localhost:8080/github-auth&client_id=deadbeef-client-id&scope=user:email
```

struggling to figure out how to get a well-defined order printed in the query params...

Tests/PointFreeTests/AuthTests.swift
```diff
@@ -9,15 +9,20 @@ import XCTest
 
 class AuthTests: TestCase {
   func testAuth() {
-    let request = URLRequest(url: URL(string: "http://localhost:8080/github-auth?code=deadbeef")!)
-      |> \.allHTTPHeaderFields .~ [
-        "Authorization": "Basic " + Data("hello:world".utf8).base64EncodedString()
-    ]
+    // NB: It seems that the result of OpenSSL crypto on Linux is not deterministic, although its decryption
+    //     is, so we cannot do snapshot tests on encrypted values :/ We will still run these tests on
+    //     macOS at least.
```
mbrandonw: so I think maybe we just cant take snapshot tests on Linux that contain encrypted data cause it changes in the OpenSSL implementation of common crypto functions. I think maybe that's fine as long as we continue running these tests on Mac.

ok maybe not _fine_, but perhaps the best we can do...

Tests/PointFreeTests/__Snapshots__/AuthTests/testLogin.1.Conn.txt
```diff
@@ -9,4 +9,4 @@
 
 ▿ Response
   Status 302 FOUND
-  Location: https://github.com/login/oauth/authorize?scope=user:email&redirect_uri=http://localhost:8080/github-auth%3Fcode%3D&client_id=deadbeef-client-id
\ No newline at end of file
+  Location: https://github.com/login/oauth/authorize?scope=user:email&client_id=deadbeef-client-id&redirect_uri=http://localhost:8080/github-auth
```
stephencelis: I forget, are we supposed to sort params?

mbrandonw: ugh well, if it's just plain key/value pairs that are strings, then that's the best way to ensure well-defined rendering. but for more complicating query strings, like `a[]=1&a[]=2` we can't just sort.

so it's complicated...

PR #38, opened 2017-12-01
Code gen the route partial isos

mbrandonw:
We've already got sourcery in the build process, might as well use it!

### DIFF
--- a/.sourcery-templates/RoutePartialIsos.stencil
+++ b/.sourcery-templates/RoutePartialIsos.stencil
@@ -5,21 +5,21 @@ import Prelude
   extension {{ enum.name }} {
     enum iso {
       {% for case in enum.cases %}
-          {% if case.hasAssociatedValue %}
-            {{ enum.accessLevel }} static let {{ case.name }} = parenthesize <| PartialIso(
-              apply: Route.{{ case.name }},
-              unapply: {
-                guard case let .{{ case.name }}(result) = $0 else { return nil }
-                return result
-            })
-          {% else %}
-            {{ enum.accessLevel }} static let {{ case.name }} = parenthesize <| PartialIso<Prelude.Unit, Route>(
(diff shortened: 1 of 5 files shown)


mbrandonw:
> I like it! It'll make the template messier, but what do ya think of formatting it in such a way that the output is indented normally?

yahhh, was on the fence. it seems most people style their stencils so that outputs are optimal, but the stencil is the thing we actually have to edit, so I preferred styling it instead :/

we could prob run `swiftlint autocorrect` on the file to clean up some things, and also lib/Syntax will be coming soon and fix all of our problems!

stephencelis:
> yahhh, was on the fence. it seems most people style their stencils so that outputs are optimal, but the stencil is the thing we actually have to edit, so I preferred styling it instead :/

I was probably overthinking, but I considered if folks decided to use this lib and codegen we'd optimize for readability. I don't think we need to worry about this right now.

.sourcery-templates/RoutePartialIsos.stencil
```diff
@@ -0,0 +1,26 @@
+import ApplicativeRouter
+import Prelude
+
+{% for enum in types.implementing.DerivePartialIsos|enum %}
+  extension {{ enum.name }} {
+    enum iso {
+      {% for case in enum.cases %}
+          {% if case.hasAssociatedValue %}
```
stephencelis: This is a level off tho!

mbrandonw: lol dangit

mbrandonw:
> I was probably overthinking, but I considered if folks decided to use this lib and codegen we'd optimize for readability. I don't think we need to worry about this right now.

well right now we're in `pointfreeco` not `swift-web`! if we extract out to `swift-web` we can generalize and clean up a bit fer sure.

PR #39, opened 2017-12-01
App env

stephencelis:
Adds an "app" env to track if we're running in dev, staging, or prod.

### DIFF
--- a/Sources/PointFree/Environment.swift
+++ b/Sources/PointFree/Environment.swift
@@ -43,30 +43,12 @@ public struct AppEnvironment {
   private static var stack: [Environment] = [Environment()]
   public static var current: Environment { return stack.last! }
 
-  public static func push(env: Environment) {
+  public static func push(_ env: Environment) {
     self.stack.append(env)
   }
 
-  public static func with(
-    airtableStuff: @escaping AirtableCreateRow = AppEnvironment.current.airtableStuff,
-    createUser: @escaping CreateUser = AppEnvironment.current.createUser,
-    envVars: EnvVars = AppEnvironment.current.envVars,
-    fetchAuthToken: @escaping FetchAuthToken = AppEnvironment.current.fetchAuthToken,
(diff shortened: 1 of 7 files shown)


Sources/PointFree/Mailgun/SendEmail.swift
```diff
@@ -25,7 +25,9 @@ func sendEmail(
         to: to,
         cc: nil,
         bcc: nil,
-        subject: subject,
+        subject: AppEnvironment.current.envVars.appEnv == .production
+          ? subject
+          : "[\(AppEnvironment.current.envVars.appEnv)] " + subject,
```
stephencelis: We can now prefix all emails with the env!

Sources/PointFree/Environment.swift
```diff
@@ -49,18 +49,22 @@ public struct AppEnvironment {
 
   public static func with(
     airtableStuff: @escaping AirtableCreateRow = AppEnvironment.current.airtableStuff,
+    createUser: @escaping CreateUser = AppEnvironment.current.createUser,
```
mbrandonw: do we wanna add `appEnv` to this? 

since it's so easy to lens into stuff I wonder if we stop maintaining arguments for every field in `with` and just require a whole new environment:

```swift
AppEnvironment.with(AppEnvironment.current |> \.appEnv .~ .development) {
  // stuff
}
```

mbrandonw: oh ha, `appEnv` is in `envVars` not the environment 👀 

so never mind the first part, but the second part stands...

stephencelis: `appEnv` is on `envVars`, so it's more like:

``` swift
AppEnvironment.with(AppEnvironment.current |> \.envVars.appEnv .~ development) {
  // stuff
}
```

stephencelis: Yup. Will remove!

PR #40, opened 2017-12-02
Modularize side effect interfaces

stephencelis:
This is a stab at reorganizing how the app environment holds onto side-effecting logic.

### DIFF
--- a/Sources/PointFree/Auth.swift
+++ b/Sources/PointFree/Auth.swift
@@ -12,7 +12,7 @@ let secretHomeResponse: (Conn<StatusLineOpen, Prelude.Unit>) -> IO<Conn<Response
     >-> readGitHubSessionCookieMiddleware
     >-> respond(secretHomeView)
 
-let githubCallbackResponse =
+let gitHubCallbackResponse =
   extractGitHubAuthCode
     <| authTokenMiddleware
 
@@ -49,15 +49,15 @@ private let missingGitHubAuthCodeMiddleware: Middleware<StatusLineOpen, Response
 
 /// Redirects to GitHub authorization and attaches the redirect specified in the connection data.
 let loginResponse: Middleware<StatusLineOpen, ResponseEnded, String?, Data> =
-  { $0 |> redirect(to: githubAuthorizationUrl(withRedirect: $0.data)) }
+  { $0 |> redirect(to: gitHubAuthorizationUrl(withRedirect: $0.data)) }
 
 let logoutResponse: (Conn<StatusLineOpen, Prelude.Unit>) -> IO<Conn<ResponseEnded, Data>> =
   redirect(
     to: path(to: .secretHome),
-    headersMiddleware: writeHeader(.clearCookie(key: githubSessionCookieName))
+    headersMiddleware: writeHeader(.clearCookie(key: gitHubSessionCookieName))
     )
 
-private let secretHomeView = View<Either<Prelude.Unit, GitHubUserEnvelope>> { data in
+private let secretHomeView = View<Either<Prelude.Unit, GitHub.UserEnvelope>> { data in
   [
     p(["welcome home"]),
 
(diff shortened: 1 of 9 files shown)


Sources/PointFree/Auth.swift
```diff
@@ -144,14 +144,14 @@ private func authTokenMiddleware(
   )
   -> IO<Conn<ResponseEnded, Data>> {
 
-    return AppEnvironment.current.fetchAuthToken(conn.data.code)
+    return AppEnvironment.current.gitHub.fetchAuthToken(conn.data.code)
       .flatMap { token in
-        AppEnvironment.current.fetchGitHubUser(token)
-          .map { user in GitHubUserEnvelope(accessToken: token, gitHubUser: user) }
+        AppEnvironment.current.gitHub.fetchUser(token)
```
stephencelis: We now scope our environment calls through these bundles, which means we can have both `gitHub.fetchUser` and `database.fetchUser`.

Sources/PointFree/Database.swift
```diff
@@ -2,88 +2,36 @@ import Either
...
-    public typealias NodeRepresentable = Any
-    struct Node {
-      subscript(_: Int, _: String) -> Node? {
-        return nil
-      }
-      var int: Int?
-      var string: String?
-    }
+public struct Database {
+  var createSubscription: (StripeSubscription, User) -> EitherIO<Error, Prelude.Unit>
```
stephencelis: I thought about bundling these functions in concrete protocols and uninhabited types, _e.g._:

``` swift
protocol Database {
  static func createSubscription: (StripeSubscription, User)
    -> EitherIO<Error, Prelude.Unit>
}

enum Postgres: Database {
  static func createSubscription: (StripeSubscription, User)
    -> EitherIO<Error, Prelude.Unit> { … }
}

enum MockDatabase: Database {
  static func createSubscription: (StripeSubscription, User)
    -> EitherIO<Error, Prelude.Unit> { … }
}

AppEnvironment.with(\.database .~ MockDatabase.self) { … }
```

But this doesn't allow for more granular stubbing in tests. By using structs, we get some nice lensing:

``` swift
AppEnvironment.with(\.database.migrate .~ { throwE(Database.Error()) }) { … }
```

Sources/PointFree/Database.swift
```diff
@@ -2,88 +2,36 @@ import Either
...
+  var createUser: (GitHub.UserEnvelope) -> EitherIO<Error, Prelude.Unit>
+  var fetchUser: (GitHub.AccessToken) -> EitherIO<Error, User?>
+  var migrate: () -> EitherIO<Error, Prelude.Unit>
+
+  static let `default` = Database(
```
stephencelis: Using a `default` constructor instead of defaulting assignments to avoid accidentally calling live side effects in our test environment.

Sources/PointFree/GitHub.swift
```diff
@@ -3,27 +3,39 @@ import Foundation
...
+    }
   }
-}
 
-public struct GitHubUser: Codable {
-  let email: String
-  let id: Int
-  let name: String
-}
+  public struct User: Codable {
```
stephencelis: Nice to be able to nest these types and have both `Database.User` and `GitHub.User`.

Sources/PointFree/Database.swift
mbrandonw: could maybe call it `live` then? `Database.live` vs `Database.mock`?

Sources/PointFree/Database.swift
```diff
@@ -201,3 +149,69 @@ public func migrate() -> EitherIO<Error, Prelude.Unit> {
     )))
     .map(const(unit))
 }
+
+#if os(iOS)
```
mbrandonw: all of this can be removed since we dont run tests for this repo on ios anymore! I've done it in my branch, but I'll resolve the conflicts.

Sources/PointFree/GitHub.swift
mbrandonw: yes!

PR #41, opened 2017-12-02
Stripe

stephencelis:
This is the basics of Stripe integration (branched from #40).

- [ ] It needs to be connected to the logged-in user
- [ ] The plan shouldn't be hard-coded.


Sources/PointFree/GitHub.swift
```diff
@@ -51,19 +51,8 @@ private func fetchAuthToken(with code: String) -> EitherIO<Prelude.Unit, GitHub.
...
-          callback(
-            data.flatMap { try? JSONDecoder().decode(GitHub.AccessToken.self, from: $0) }
-              .map(Either.right)
-              ?? Either.left(unit)
-          )
-        }
-        .resume()
-    }
-  )
+  return jsonDataTask(with: request)
```
stephencelis: Fun to clean these up!

Sources/PointFree/Pricing.swift
```diff
@@ -71,14 +71,71 @@ private let pricingView = View<Prelude.Unit> { _ in
...
           ]),
         ]
-        <> footerView.view(unit))
-      ])
+        <> footerView.view(unit)
+        <> [
+          script([src("https://js.stripe.com/v3/")]),
+          script(
+            """
+            var apiKey = document.getElementById('card-element').dataset.stripeKey;
```
stephencelis: Out script tag takes a `StaticString`, so things like the API key, here, can't be interpolated. That's potentially a good thing, though was a gotcha!

Sources/PointFree/Pricing.swift
```diff
@@ -71,14 +71,71 @@ private let pricingView = View<Prelude.Unit> { _ in
...
+          )
+        ]
+      )
     ])
+  ])
+}
+
+// FIXME: Move to swift-web
+public func data<T>(_ name: StaticString, _ value: String) -> Attribute<T> {
+  return .init("data-\(name)", value)
```
stephencelis: Also couldn't use `attribute("data-" + name, value)` here because `attribute` also takes `StaticString`.

Sources/PointFree/Routes.swift
```diff
@@ -52,11 +53,24 @@ private let routers: [Router<Route>] = [
   Route.iso.secretHome
     <¢> get %> lit("home") <% end,
 
+  Route.iso.subscribe
+    <¢> post %> lit("subscribe") %> formField("plan", .rawRepresentable) <%> formField("token") <% end,
```
stephencelis: This makes it easy to use "primitive" new-types with our router.

Sources/PointFree/Routes.swift
```diff
@@ -71,7 +85,7 @@ extension PartialIso where A == String, B == Tag {
   public static var tag: PartialIso<String, Tag> {
     return PartialIso<String, Tag>(
       apply: Tag.init(slug:),
-      unapply: get(\.name)
+      unapply: ^\.name
```
stephencelis: Trying to get in the habit :)

Sources/PointFree/Stripe.swift
```diff
@@ -0,0 +1,220 @@
...
+      request.httpMethod = "GET"
+    case let .post(params):
+      request.httpMethod = "POST"
+      request.httpBody = Data(urlFormEncode(value: params).utf8)
+    case .delete:
+      request.httpMethod = "DELETE"
+    }
+
+    return pure(request)
+      .flatMap { jsonDataTask(with: auth <| $0) }
```
stephencelis: This was a bit of a gotcha! Tests push/pop the environment and may access/instantiate environment data for the first time while being mutated. `AppEnvironment` is side-effecting, so we should never call it outside IO (`auth` here accesses `AppEnvironment`).

Sources/PointFree/Pricing.swift
mbrandonw: lol yeah. what do you think about having an `unsafeString` function that allows unencoded data through? seems no matter what we need an escape hatch unless we attempt Swift.js :/

Sources/PointFree/Routes.swift
mbrandonw: lol thanks!

Sources/PointFree/Stripe.swift
```diff
@@ -0,0 +1,220 @@
...
+  public struct Token: Codable {
+    let id: String
+  }
+}
+
+//private let subscriptionPlans = fetchPlans
+//  .map(^\.data >>> filter(StripeSubscriptionPlan.Id.all.contains <<< ^\.id))
+
+private func cancelSubscription(id: String) -> EitherIO<Prelude.Unit, Stripe.Subscription> {
+  return stripeDataTask("https://api.stripe.com/v1/subscriptions/\(id)", .delete)
```
mbrandonw: I forget, is stripe sandbox api entirely determined by the api keys, or is there another URL to hit?

Sources/PointFree/Stripe.swift
mbrandonw: ahhh yes yes, we ran into the exact same thing at kickstarter. we would have to close over it with a `viewDidLoad.flatMap { }` just like you've done here :/

Sources/PointFree/Stripe.swift
```diff
@@ -0,0 +1,220 @@
...
+      request.httpMethod = "DELETE"
+    }
+
+    return pure(request)
+      .flatMap { jsonDataTask(with: auth <| $0) }
+      .withExcept(const(unit))
+}
+
+private func auth(_ request: URLRequest) -> URLRequest {
+  return request |> \.allHTTPHeaderFields %~ attachStripeAuthorization
```
mbrandonw: couldn't this be `let auth = \.allHTTPHeaderFields %~ attachStripeAuthorization` :D

mbrandonw: oh also doesnt seem this is actually used...

Sources/PointFree/Stripe.swift
```diff
@@ -0,0 +1,220 @@
...
+  let secret = Data("\(AppEnvironment.current.envVars.stripe.secretKey):".utf8).base64EncodedString()
+  return (headers ?? [:])
+    |> key("Authorization") .~ ("Basic " + secret) // TODO: Use key path subscript
+}
+
+public protocol SingleValueCodable: Codable, RawRepresentable {}
+
+extension SingleValueCodable where RawValue: Codable {
+  public init(from decoder: Decoder) throws {
+    self.init(rawValue: try decoder.singleValueContainer().decode(RawValue.self))!
```
mbrandonw: the force unwrap is safe here? should we throw instead?

Sources/PointFree/Pricing.swift
mbrandonw: also whatdya think about moving this into its own string (and maybe file) like we did for svg? to keep this view light.

Sources/PointFree/Pricing.swift
```diff
@@ -71,14 +71,71 @@ private let pricingView = View<Prelude.Unit> { _ in
                   ])
                 ]),
 
+              ]),
+            form([action("/subscribe"), id("payment-form"), method(.post)], [
+              input([name("plan"), type(.hidden), value(Stripe.Plan.Id.monthly.rawValue)]),
+              input([name("token"), type(.hidden)]),
+              div([id("card-element"), data("stripe-key", AppEnvironment.current.envVars.stripe.publishableKey)], []),
+              div([id("card-errors"), role(.alert)], []),
+              button(["Submit Payment"])
```
mbrandonw: does it matter this isn't an `<input type="submit">` kind of button?

stephencelis:
Merging to make conflicts easier! Will tend to the above TODOs in another PR.

PR #42, opened 2017-12-03
User nav

mbrandonw:
### What

Adds some data types and views for handling user nav. Quite a bit going on here, so will make inline notes...


Sources/PointFree/Database.swift
```diff
@@ -91,17 +92,15 @@ private func fetchUser(with token: GitHub.AccessToken) -> EitherIO<Error, Databa
     [token.accessToken]
     )
     .map { result -> Database.User? in
-      let uuid = result[3, "id"].flatMap(get(\.string) >>> flatMap(UUID.init(uuidString:)))
-
-      let subscriptionId = result[5, "subscription_id"]
-        .flatMap(get(\.string) >>> flatMap(UUID.init(uuidString:)))
+      let uuid = result["id"]?.array?.first?.wrapped.string.flatMap(UUID.init(uuidString:))
```
mbrandonw: lol I really hope there is a better way but this seems to be the way to get this data from the `result`?!

Sources/PointFree/Database.swift
```diff
@@ -150,36 +149,6 @@ private func migrate() -> EitherIO<Error, Prelude.Unit> {
     .map(const(unit))
 }
 
-#if os(iOS)
```
mbrandonw: we dont run tests on ios anymore (now that snapshots work on macOS) so removing this

Sources/PointFree/RequestContext.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import HttpPipeline
+import Prelude
+
+/// A value that wraps any given data with additional context that is useful for completing the
+/// request-to-response lifecycle.
+struct RequestContext<A> {
+  private(set) var currentUser: Database.User? = nil
+  private(set) var currentRequest: URLRequest
```
mbrandonw: Maybe the request context should carry around the current route instead of bare request. not sure if there might be a situation in which you need the actual request...

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -5,7 +5,8 @@ import Prelude
 import Styleguide
 
 public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
-  requireHerokuHttps(allowedInsecureHosts: allowedInsecureHosts)
+  requestLogger
```
mbrandonw: just for yucks

Sources/PointFree/TODO.swift
```diff
@@ -69,3 +72,36 @@ extension CssSelector {
...
 }
+
+// TODO: Move to HttpPipeline
+public func >¢< <A, B, C, I, J>(
+  lhs: @escaping (A) -> C,
+  rhs: @escaping Middleware<I, J, C, B>
+  )
+  -> Middleware<I, J, A, B> {
+
+    return map(lhs) >>> rhs
```
mbrandonw: arg so technically this is contramapping down _two_ levels... first level is the function arrow `->` and second level is `Conn` :/ which we could compose these dang things

Sources/Styleguide/DesignSystems/Layout.swift
```diff
@@ -1,30 +1,40 @@
...
-    block: CssSelector.class("inline"),
-    inlineBlock: CssSelector.class("inline"),
-    overflowHidden: CssSelector.class("overflow-hidden"),
-    overflowScroll: CssSelector.class("overflow-scroll"),
-    overflowAuto: CssSelector.class("overflow-auto"),
-    clearFix: CssSelector.class("clear-fix"),
-    left: CssSelector.class("left"),
-    right: CssSelector.class("right")
-  )
+  public enum layout {
```
mbrandonw: I just converted this back to an `enum` namespace from a tuple. for a moment there I was like the tuple style, but it can be a lil annoying and you can't document each field.

Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -40,6 +46,24 @@ extension Class {
       public static let footnote = CssSelector.class("pf-footnote")
       public static let callout = CssSelector.class("pf-callout")
     }
+
+    public enum components {
+      /// The standard nav bar style.
+      public static let navBar =
```
mbrandonw: this is cool! we have a single `navBar` selector that can be added to a dom element, i.e. `div([`class`([Class.pf.components.navBar])])`, and since `navBar` is just a union of a whole bunch of other selectors, all of those names get rendered into the class attribute, i.e.`class="nav-bar bg-purple px-2 lh-4 h-4"`, etc.

makes it nice to build components from the single styles!

Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -40,6 +46,24 @@ extension Class {
...
+        _navBar
+          | Class.pf.colors.bg.purple
+          | Class.padding.leftRight(2)
+          | Class.type.lineHeight(rem: 4)
+          | Class.size.height(rem: 4)
+
+      /// A minimal nav bar style.
+      public static let minimalNavBar =
```
mbrandonw: this is the version of the nav bar we use on the 404 page

Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -192,12 +215,6 @@ private let typeStyles =
...
 private let navBarStyles =
-  Class.pf.navBar % (
-    height(.px(64))
-      <> lineHeight(.px(64))
-      <> backgroundColor(Colors.purple)
-      <> color(.other("#fff"))
-      <> padding(topBottom: 0, leftRight: .rem(2))
-  )
-  <> ((Class.pf.navBar ** a) | (Class.pf.navBar ** a & .pseudo(.link))) % color(.other("#fff"))
+  ((_navBar ** a) | (_navBar ** a & .pseudo(.link))) % color(.other("#fff"))
```
mbrandonw: still using cascading for this, tho there may be a way to remove it

Tests/PointFreeTests/EpisodeTests.swift
```diff
@@ -8,25 +8,48 @@ import XCTest
...
+        "Authorization": "Basic " + Data("hello:world".utf8).base64EncodedString()
+    ]
+
+    let conn = connection(from: request)
+    let result = conn |> siteMiddleware
+
+    assertSnapshot(matching: result.perform())
+
+    #if !os(Linux)
+      if #available(OSX 10.13, *) {
```
mbrandonw: we have to double `if` this :/ once for linux, and once for macOS API check. once circleci upgrades to 10.13 we can at least remove one of em

Tests/StyleguideTests/StyleguideTests.swift
```diff
@@ -5,10 +5,41 @@ import XCTest
 
 class StyleguideTests: XCTestCase {
   func testStyleguide() {
```
mbrandonw: so just to remind you of some terminology I've settled on. we have 

* design systems: this is the lib of atomic styles that are mostly ported from basscss. this could be split into it's own lib. we could even have a function `(Config) -> Stylesheet` where a user of the lib gets to specify all of their preferences and then out comes a full lib of atomic styles suited for them

* point free base styles: this is everything that is PF specific and should be (mostly) derived by combining things in the design systems

* styleguide: the union of those two ^ things

Sources/PointFree/Episode/Episode.swift
```diff
@@ -9,50 +9,40 @@ import Optics
...
         style(styleguide),
-        title("Episode #\(ep.sequence): \(ep.title)"),
+        title("Episode #\(context.data.sequence): \(context.data.title)"),
         meta(viewport: .width(.deviceWidth), .initialScale(1)),
         ]),
 
       body(
         [`class`([Class.pf.colors.bg.dark])],
-        navView.view(unit) <> [
+        navView.view(context.map(const(unit))) <> [
```
mbrandonw: will soon figure out how to make a layout view that takes care of header and footer automatically.

Sources/PointFree/Database.swift
stephencelis: 😨

Sources/PointFree/Database.swift
```diff
@@ -1,12 +1,13 @@
...
 import Foundation
+import PostgreSQL
 import Prelude
 
 public struct Database {
   var createSubscription: (StripeSubscription, User) -> EitherIO<Error, Prelude.Unit>
   var createUser: (GitHub.UserEnvelope) -> EitherIO<Error, Prelude.Unit>
   var fetchUser: (GitHub.AccessToken) -> EitherIO<Error, User?>
-  var migrate: () -> EitherIO<Error, Prelude.Unit>
+  public var migrate: () -> EitherIO<Error, Prelude.Unit>
```
stephencelis: Thanks!

Sources/PointFree/Database.swift
```diff
@@ -91,17 +92,15 @@ private func fetchUser(with token: GitHub.AccessToken) -> EitherIO<Error, Databa
...
-      let subscriptionId = result[5, "subscription_id"]
-        .flatMap(get(\.string) >>> flatMap(UUID.init(uuidString:)))
+      let uuid = result["id"]?.array?.first?.wrapped.string.flatMap(UUID.init(uuidString:))
+      let subscriptionId = result["subscription_id"]?.array?.first?.wrapped.string.flatMap(UUID.init(uuidString:))
 
       return curry(Database.User.init)
-        <¢> result[0, "email"]?.string
-        <*> result[1, "github_user_id"]?.int
-        <*> result[2, "github_access_token"]?.string
+        <¢> result["email"]?.array?.first?.wrapped.string
```
stephencelis: This does seem bonkers, but it's isolated enough I guess.

Sources/PointFree/Episode/Episode.swift
```diff
@@ -9,50 +9,40 @@ import Optics
...
-    return conn.map(const(unit))
-      |> respond(notFoundView)
-  case let .some(ep):
-    return conn.map(const(ep))
-      |> respond(episodeView.map(addHighlightJs >>> addGoogleAnalytics))
-  }
-}
-
-public let episodeView = View<Episode> { ep in
+let episodeView = View<RequestContext<Episode>> { context in
```
stephencelis: I'm cool calling this `ctx` wherever needed to minimize chars.

Sources/PointFree/Episode/Episode.swift
```diff
@@ -197,20 +187,34 @@ private let transcriptBlockView = View<Episode.TranscriptBlock> { block -> Node
...
-        ])
+      body(
+        minimalNavView.view(unit) <> [
+        gridRow([`class`([Class.grid.center(.xs)])], [
+          gridColumn(sizes: [:], [
+            div([`class`([Class.padding.all(4)])], [
+              h5([`class`([Class.h5])], ["Episode not found :("]),
+              pre([
+                code([`class`([Class.pf.code(lang: "swift")])], [
+                  "f: (Episode) -> Never"
```
stephencelis: 😄

Sources/PointFree/Episode/Episodes.swift
```diff
@@ -8,27 +8,23 @@ import Prelude
...
+    >>> requestContextMiddleware
+    >-> writeStatus(.ok)
     >-> respond(episodesDocumentView.map(addGoogleAnalytics))
 
-private func fetchEpisodes(_ conn: Conn<StatusLineOpen, Tag?>) -> IO<Conn<HeadersOpen, ([Episode], Tag?)>> {
+private func episodes(for tag: Tag?) -> (episodes: [Episode], selectedTag: Tag?) {
+  return (
+    episodes: episodes
+      .filter { tag.map($0.tags.contains) != .some(false) }
+      .sorted(by: get(\.sequence))
```
stephencelis:
``` swift
.sorted(by: ^\.sequence)
```

Sources/PointFree/Nav.swift
```diff
@@ -1,22 +1,86 @@
...
-        ]),
-      li([`class`([Class.layout.inline])], [
-        a([href(path(to: .about)), `class`([Class.padding.leftRight(1)])], ["About"])
-        ]),
-      ])
+let navView = View<RequestContext<Prelude.Unit>> { context in
+  [
+    gridRow([`class`([Class.pf.components.navBar, Class.grid.between(.xs)])], [
+      gridColumn(
+        sizes: [:],
```
stephencelis: Should this have an empty default?

Sources/PointFree/RequestContext.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import HttpPipeline
+import Prelude
+
+/// A value that wraps any given data with additional context that is useful for completing the
+/// request-to-response lifecycle.
+struct RequestContext<A> {
```
stephencelis: This is nice! Seems like `RequestContext` is also `LayoutContext`: things that the page depends on no matter what. With different layouts it's a bit more amorphous. In the end a band-aid to approach row polymorphism? Wonder if revisiting some of that protocol union stuff you explored with GraphQL would pay off?

Sources/PointFree/RequestContext.swift
stephencelis: 🤔 Seems like maybe the route would be tougher to _use_?

Sources/PointFree/RequestContext.swift
```diff
@@ -0,0 +1,59 @@
...
+          currentUser: $0,
+          currentRequest: conn.request,
+          data: conn.data
+        )
+      )
+    )
+  }
+}
+
+///
```
stephencelis: 🤔

Sources/Styleguide/DesignSystems/Layout.swift
```diff
@@ -1,30 +1,40 @@
...
+  Class.layout.inline % display(.inline)
     <> Class.layout.block % display(.block)
     <> Class.layout.inlineBlock % display(.inlineBlock)
     <> ".table" % display(.table)
     <> ".table-cell" % display(.tableCell)
     <> Class.layout.overflowHidden % overflow(.hidden)
     <> Class.layout.overflowScroll % overflow(.scroll)
     <> Class.layout.overflowAuto % overflow(.auto)
+    <> Class.layout.overflowAuto(.x) % overflow(x: .auto)
+    <> Class.layout.overflowAuto(.y) % overflow(x: .auto)
```
stephencelis: How does this work with the `y` and `x` mismatch?

Sources/PointFree/Auth.swift
```diff
@@ -150,6 +140,7 @@ private func authTokenMiddleware(
           .map { user in GitHub.UserEnvelope(accessToken: token, gitHubUser: user) }
       }
       .flatMap { env in
+        // todo: fetch or create aint working
```
stephencelis: 🤭

Sources/PointFree/RequestContext.swift
mbrandonw: ahhhh layout context! yeah with the lil I'm played around with layouts I was making wrapper types and the `contramap`ing on em to glue em together. we could rename later if that starts to make more sense

Sources/Styleguide/DesignSystems/Layout.swift
mbrandonw: whoops!

Tests/PointFreeTests/NavViewTests.swift
```diff
@@ -0,0 +1,77 @@
...
+}
+
+private let loggedOutRequestContext = RequestContext(
+  currentUser: nil,
+  currentRequest: URLRequest(url: URL(string: "/")!),
+  data: unit
+)
+
+private let loggedInRequestContext = RequestContext(
+  currentUser: Database.User(
```
stephencelis: We gotta mock for this. But it should be updated with that better UUID.

PR #43, opened 2017-12-03
Modulemap for CPostgreSQL so that playgrounds work again

mbrandonw:
Basically the same thing as we did for https://github.com/pointfreeco/swift-web/pull/82, except now for posgres. This allows us to load the `PointFree` package into playgrounds!

There is one final thing to fix tho... when importing `SnapshotTesting` we get:

```
Playground execution failed:

error: pointfreeco.playground:7:8: error: missing required module 'WKSnapshotConfigurationShim'
import SnapshotTesting
       ^
```

It can be useful to import this into playgrounds cause it gives us an easy way to capture the output of middleware and print it, which is nice for requests that don't render a page but instead do redirects or something else. Hopefully we can fix that soon.


Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -0,0 +1,111 @@
+import Foundation
```
mbrandonw: I did something I wanted to do at kickstarter too... move all the mocks to its own package so that we can use them in tests _and_ playgrounds!

pointfreeco.playground/Pages/Episodes Page.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,28 @@
...
+import HtmlCssSupport
+import HttpPipeline
+import PlaygroundSupport
+@testable import PointFree
+@testable import PointFreeTestSupport
+import Prelude
+import WebKit
+import Optics
+
+AppEnvironment.push(.mock)
```
mbrandonw: now it's super easy to set up a playground with mocks.

stephencelis:
> There is one final thing to fix tho... when importing `SnapshotTesting` we get [...]

Maybe something like this?

```
module WKSnapshotConfigurationShim [system] {
  header "$(SDK_PATH)/System/Library/Frameworks/WebKit.framework/Headers/WKSnapshotConfiguration.h"
  export *
}
```

mbrandonw:
> Maybe something like this?

didnt work :/ it's looking for `WKSnapshotConfigurationShim`. how do we point it to that?

PR #44, opened 2017-12-04
Possible solution to private episode transcripts

mbrandonw:
This is more of a proof-of-concept/work-in-progress. Lemme know if you can think of a better way or see a way to simplify things.

There are 3 other repos that go with this PR (woof!). If we're cool with this direction I will PR those later. The repos are:

* `pointfreeco/episode-models`: A repo that holds only the data types for modeling our episodes. Here's the branch I'm working off now: https://github.com/pointfreeco/episode-models/compare/initial-models
* `pointfreeco/episodes-oss`: A repo that depends on `episode-models` and exposes a single public value, `allEpisodes`, which is an array of `Episode` values. This is the repo that will be open-sourced and anyone can see. We can think of it as just mock values that will be useful for development and even tests. You can see my branch here: https://github.com/pointfreeco/episodes-oss/compare/eps
* `mbrandonw/episodes`: A repo that also depends on `episode-models` and also exposes a single public value, `allEpisodes`. This will be the private repo that only we have access to, hence it's under my personal account instead of `pointfreeco` cause we dont yet pay for private repos. You can see my working branch here: https://github.com/mbrandonw/episodes/compare/mock-episodes

So, notice that the `episodes-oss` and `episodes` repos basically have the same structure, hence those modules are interchangeable. I then use an environment variable in the `Package.swift` of this repo to figure out which package we should include. I default to the OSS repo, and in fact think that'll be the one we work with the most. I think the only time we'll use the private repo is when deploying most likely.

And that's about it. I also made a few small tweaks to the model while doing this to better normalize it.


Package.swift
```diff
@@ -1,7 +1,14 @@
 // swift-tools-version:4.0
-
+import Foundation
 import PackageDescription
 
+let episodesDependency: Package.Dependency
+if ProcessInfo.processInfo.environment["ENV"] == "PF" {
```
mbrandonw: so doing `ENV=PF swift package update` will pull in the private episode values.

mbrandonw: open to other names fo the environment var and value

Sources/PointFree/Episode/Episode.swift
```diff
@@ -101,11 +103,16 @@ private let topLevelTagsView = View<[Tag]> { tags in
     ])
 }
 
-private let episodeTocView = View<[Episode.TranscriptBlock]> { blocks in
+private let episodeTocView = View<[TranscriptBlock]> { blocks in
   blocks
-    .filter { $0.type == .title && $0.timestamp != nil }
-    .flatMap { block in
-      tocEntryView.view((block.content, block.timestamp ?? 0))
+    .flatMap { block -> [Node] in
```
mbrandonw: this `-> [Node]` seems to be required, otherwise Swift tries to infer this as an `-> B?` flat map :/

Sources/PointFree/Episode/Episode.swift
```diff
@@ -101,11 +103,16 @@ private let topLevelTagsView = View<[Tag]> { tags in
...
+      switch block.type {
+      case let .title(timestamp):
```
mbrandonw: I moved the `timestamp` value into the `BlockType` instead of the `TranscriptionBlock`. the idea being not every block as a timestamp that makes sense (like images and code snippets), only certain blocks.

Sources/PointFree/Episode/EpisodeModelExtensions.swift
```diff
@@ -0,0 +1,23 @@
+import EpisodeModels
+
+extension Episode {
+  var slug: String {
+    return "ep\(self.sequence)-\(PointFree.slug(for: self.title))"
```
mbrandonw: keeping the slug computation in this repo since it's domain specific and determines how our permalinks look. seems like it should stay local to this code.

Sources/PointFree/Episode/EpisodeModelExtensions.swift
```diff
@@ -0,0 +1,23 @@
...
+
+  init?(slug: String) {
+    guard let tag = array(Tag.all).first(where: { PointFree.slug(for: slug) == $0.slug })
+      else { return nil }
+    self = tag
+  }
+}
+
+private func slug(for string: String) -> String {
+  return string.lowercased().replacingOccurrences(of: " ", with: "-")
```
mbrandonw: should prob do more work to slugify strings, but maybe this will do for now

Sources/PointFree/Routes.swift
```diff
@@ -61,16 +62,6 @@ private let routers: [Router<Route>] = [
 
 ]
 
-// TODO: Move to swift-web
-extension PartialIso where A == String, B: RawRepresentable, B.RawValue == String {
```
mbrandonw: should moved to the `TODO.swift` file

mbrandonw:
> Oof. This definitely has a lot of moving parts but no alternatives really come to mind immediately. Let’s sleep on it, though this isn’t so bad if it’s the way we end up going!

👍 sounds good. while doing this I did make a few edits to transcripts to get a feel of the flow, and it's not _terrible_, but would love to find a simpler way too.

mbrandonw:
Closing this for #90

PR #45, opened 2017-12-04
Derive more from just the atomic classes

mbrandonw:
Just wanted to have a bit of fun. I replaced some CSS with just composition of selectors. So, instead of having CSS like `.pf-title1 { /* styles */ }`, we now have a single `CssSelector` that is the union of all the atomic classes that make up a "title 1", i.e. bold, black text color, h2, line-height 125% etc. That means if you do `h2([`class`([Class.pf.type.title1])], "Hello world!"])` it will expand to html like:

```html 
<h2 class="fg-black bold h2 lh-1">Hello world!</h2>
```

Allows us to describe high level semantic styles in code while having it expand to atomic units at runtime.


Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -37,12 +38,43 @@ extension Class {
...
-      public static let title1 = CssSelector.class("pf-title-1")
-      public static let title2 = CssSelector.class("pf-title-2")
-      public static let title3 = CssSelector.class("pf-title-3")
-      public static let headline = CssSelector.class("pf-headline")
-      public static let subhead = CssSelector.class("pf-subhead")
+      public static let largeTitle =
+        Class.pf.colors.fg.black
+          | Class.type.bold
+          | Class.h1
+          | Class.type.lineHeight(0)
```
mbrandonw: all of these selectors used to have their styles defined down below, but really we can just define them as the union of styles that we already have from `DesignSystems`.

Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -181,39 +214,13 @@ private let aStyles =
...
-      | Class.pf.type.subhead) % fontWeight(.w700)
-
-    <> Class.pf.type.largeTitle % (fontSize(.rem(3.998)) <> lineHeight(1.15))
-    <> Class.pf.type.title1 % (fontSize(.rem(2.827)) <> lineHeight(1.25))
-    <> Class.pf.type.title2 % (fontSize(.rem(1.999)) <> lineHeight(1.25))
-    <> Class.pf.type.title3 % (fontSize(.rem(1.414)) <> lineHeight(1.45))
+private let _headline = CssSelector.class("pf-headline")
 
+private let typeStyles =
+  _headline % fontWeight(.w600)
```
mbrandonw: we dont have a name for weight 600 in design systems. we should prob have `w-100`, `w-200`, etc... classes

PR #46, opened 2017-12-04
CC license for content, MIT for code

mbrandonw:
We can use a creative commons license for the content of the site (images, copy, etc...) and use an MIT license for the actual code.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Point Free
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Point Free
+Copyright (c) 2017 Point-Free, Inc.
```
mbrandonw: no we're a c-corp! gonna update the other repos too

PR #47, opened 2017-12-04
Send a confirmation email when signing up.

mbrandonw:
### What 

Sends an email to the user when they sign up for our launch.

I wanted to use this as a testbed for getting some basic email template infrastructure (with `<table>` grids, etc), but I think we need to do a bit more work on the HTML library first.


Sources/PointFree/LaunchSignup/LaunchSignup.swift
```diff
@@ -36,14 +37,24 @@ private func airtableStuff<I>(_ conn: Conn<I, String>) -> IO<Conn<I, Either<Prel
 func notifyUsOfNewSignup<I>(_ conn: Conn<I, String>) -> IO<Conn<I, String>> {
   return IO {
 
-    // Fire-and-forget to notify us that someone signed up
-    parallel(
-      sendEmail(
+    // Fire-and-forget to send some emails
+    zip(
```
stephencelis: Are you sure this is parallelizing? I think `run` itself is synchronous, and that we need to do something like:

``` swift
let both = tuple
  <^> parallel <| sendEmail(…)
  <*> parallel <| sendEmail(…)
both.run { _ in }
```

stephencelis: (This'll hopefully improve when we rethink some of our IO story.)

mbrandonw: Well there’s two runs! One for EitherIO which just gets the underlying IO and then I run it as parallel. So I do think this is parallel

stephencelis: Oh nm missed which `zip` was being used!

PR #50, opened 2017-12-05
Proper "fetch or register" with basic registration email

mbrandonw:
Fixes the bug where fetch-or-create didnt actually work, and now when we detect a new user we send a confirmation email.


Sources/PointFree/Auth.swift
```diff
@@ -1,15 +1,19 @@
...
 import HttpPipelineHtmlSupport
 import Optics
 import Prelude
+import Styleguide
 import UrlFormEncoding
 
 let secretHomeResponse: (Conn<StatusLineOpen, Prelude.Unit>) -> IO<Conn<ResponseEnded, Data>> =
   writeStatus(.ok)
     >-> readGitHubSessionCookieMiddleware
+    >-> (ignoreErrors >>> pure)
```
mbrandonw: just some middleware to convert `Either<E, A>` to `A?`. there are a few other ways to write this, so down to experiment.

Sources/PointFree/Auth.swift
```diff
@@ -128,6 +136,40 @@ extension EitherIO {
   }
 }
 
+private func fetchOrRegisterUser(env: GitHub.UserEnvelope) -> EitherIO<Prelude.Unit, Database.User> {
+
+  return AppEnvironment.current.database.fetchUserByGitHub(env.accessToken)
+    .flatMap { user in
+      EitherIO(run: IO { user.map(Either.right) ?? .left(unit) })
+        .catch(const(registerUser(env: env)))
```
mbrandonw: catching the error in here means we didnt find the user (as opposed to having a sql error)

Sources/PointFree/Auth.swift
```diff
@@ -180,8 +218,31 @@ private func gitHubAuthorizationUrl(withRedirect redirect: String?) -> String {
...
+        style(styleguide),
+        ]),
+
+      body([
+        gridRow([
+          gridColumn(sizes: [:], [
+            div([`class`([Class.padding.all(2)])], [
+              h3([`class`([Class.h3])], ["Thanks for signing up!"]),
+              p([`class`([Class.padding.topBottom(2)])], [
+                "You’re one step closer to our weekly video series!",
```
mbrandonw: clearly we need more here soon...

PR #51, opened 2017-12-05
Third party routers

mbrandonw:
Who says we can only use the router for point-free routes?!

I added a router for GitHub and Twitter so that we can get rid of stringy urls!


.sourcery-templates/RoutePartialIsos.stencil
```diff
@@ -7,13 +7,13 @@ import Prelude
       {% for case in enum.cases %}
         {% if case.hasAssociatedValue %}
           {{ enum.accessLevel }} static let {{ case.name }} = parenthesize <| PartialIso(
-            apply: Route.{{ case.name }},
+            apply: {{ enum.name }}.{{ case.name }},
```
mbrandonw: this was a mistake in my original template

Makefile
```diff
@@ -37,14 +37,15 @@ linux-main:
...
+		--args testimports='$(imports)'
+	mv ./Tests/LinuxMain.generated.swift ./Tests/LinuxMain.swift
 
 route-partial-iso:
+	mkdir -p ./Sources/PointFree/Generated
 	sourcery \
 		--sources ./Sources/PointFree/ \
 		--templates ./.sourcery-templates/RoutePartialIsos.stencil \
-		--output ./Sources/PointFree/
+		--output ./Sources/PointFree/Generated/DerivedPartialIsos.swift
```
mbrandonw: moving code gen files to a `Generated` directory. was hoping I could use `.generated`, but SPM seems to ignore hidden directories.

Sources/PointFree/Auth.swift
```diff
@@ -164,20 +164,13 @@ private func authTokenMiddleware(
...
-  ]
-
-  let queryString = params
-    .map { key, value in
-      key + "=" + (value.addingPercentEncoding(withAllowedCharacters: .urlQueryParamAllowed) ?? "")
-    }
-    .joined(separator: "&")
-
-  return "https://github.com/login/oauth/authorize?\(queryString)"
+  return gitHubUrl(
```
mbrandonw: since to get rid of all this query param manipulation and just use the router!

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -9,11 +9,16 @@ extension Environment {
...
     envVars: EnvVars(),
     gitHub: .mock,
+    logger: .mock,
     sendEmail: const(pure(.init(id: "deadbeef", message: "success!"))),
     stripe: .mock
   )
 }
 
+extension Logger {
+  public static let mock = Logger(level: .debug, logger: { _ in })
```
mbrandonw: this ok to do for tests? so we dont get flooded with logs

stephencelis: Yep looks good!

Makefile
stephencelis: Could also do `__Generated__/`!

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
stephencelis: We could even make a buffer and write tests/snapshots against our logging!

Makefile
mbrandonw: ah duh! will do!

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
mbrandonw: yeah that'd be super cool

PR #53, opened 2017-12-10
Most of homepage

mbrandonw:
I don't think I can open a PR with less than 2K LOC changed anymore...

### What

This has most of the homepage designs @amotion worked on. The biggest thing missing is a lil mobile finesse.

However, in doing this I added quite a bit to our design systems and (partially) refactored how we do breakpoints so that we can support responsive paddings/margins. Will add comments!

testhomepage 2 desktop copy


Sources/PointFree/About.swift
```diff
@@ -22,7 +22,7 @@ private let aboutView = View<RequestContext<Prelude.Unit>> { ctx in
       body(navView.view(ctx) + [
         gridRow([
           gridColumn(sizes: [.xs: 12], [
-            div([`class`([Class.padding.all(4)])], [
+            div([`class`([Class.padding([.mobile: [.all: 4]])])], [
```
mbrandonw: this is now the interface to adding padding/margin. You specify a dictionary that maps breakpoints to sides+value. If you are only applying a single side it's a lil more verbose, but if you are doing multiple it's shorter:

```
[Class.padding([.mobile: [.leftRight: 2, .topBottom: 3]])]

// vs

[Class.padding(leftRight: 2), Class.padding(topBottom: 3)]
```

and even shorter when you start adding breakpoint specific values.

mbrandonw: oh also, we can cook up components to automate some stuff. for example, the default padding in a grid column should prob be:

```swift 
let defaultColumnPadding = Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]])
```

And then we can just use `defaultColumnPadding` everywhere.

Sources/PointFree/Auth.swift
```diff
@@ -10,12 +10,6 @@ import Prelude
 import Styleguide
 import UrlFormEncoding
 
-let secretHomeResponse: (Conn<StatusLineOpen, Prelude.Unit>) -> IO<Conn<ResponseEnded, Data>> =
```
mbrandonw: moved homepage stuff to `Home.swift`

Sources/PointFree/Episode/EpisodeModel.swift
```diff
@@ -1,5 +1,6 @@
 public struct Episode {
   var blurb: String
+  var codeSampleDirectory: String
```
mbrandonw: a directory inside https://github.com/pointfreeco/episode-code-samples corresponding to that episode's code sample

Sources/PointFreeTestSupport/TestCase.swift
```diff
@@ -1,16 +1,15 @@
 @testable import PointFree
-import PointFreeTestSupport
 import SnapshotTesting
 import XCTest
 
 // sourcery: disableTests
-class TestCase: XCTestCase {
-  override func setUp() {
+open class TestCase: XCTestCase {
```
mbrandonw: moved to `PointFreeTestSupport` so that we could use in style guide tests too

Sources/Styleguide/DesignSystems/Breakpoint.swift
```diff
@@ -1,4 +1,44 @@
 import Css
+import Prelude
+
+public enum _Breakpoint: String {
```
mbrandonw: this is still underscored cause I havent replaced all uses of `Breakpoint` yet. it's mostly the grid which I will take care of soon.

Sources/Styleguide/DesignSystems/Spacing.swift
```diff
@@ -1,116 +1,139 @@
...
 
-    public static func bottom(_ n: Int) -> CssSelector {
-      return .class("mb\(n)")
-    }
+  static let all: [Whitespace] = [.margin, .padding]
+}
 
-    public static func left(_ n: Int) -> CssSelector {
-      return .class("ml\(n)")
+private let responsivePaddingStyles =
```
mbrandonw: this wild 3-deep nested loop generates all the utility classes for padding on all sides, all breakpoints, and all values.

amazingly, the generated css is _smaller_ than the previous way of doing this (which didnt even include breakpoints).

Sources/Styleguide/DesignSystems/TypeScale.swift
```diff
@@ -11,9 +11,9 @@ extension Class {
 }
 
 public let typescale =
-       Class.h1 % fontSize(.rem(3.998))
-    <> Class.h2 % fontSize(.rem(2.827))
-    <> Class.h3 % fontSize(.rem(1.999))
-    <> Class.h4 % fontSize(.rem(1.414))
-    <> Class.h5 % fontSize(.rem(1.0))
-    <> Class.h6 % fontSize(.rem(0.707))
+  Class.h1 % fontSize(.rem(4))         // 64  56
```
mbrandonw: just simpler. golden ratio is a lie anyway

Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -20,87 +28,147 @@ extension Class {
...
-        _navBar
+    public enum buttons {
+      private static let base =
+        baseButtonClass
+          | Class.type.medium
+          | Class.h5
+          | Class.padding([.mobile: [.leftRight: 2]])
+          | Class.padding([.mobile: [.topBottom: 1]])
+          | Class.pf.colors.link.white
+          | Class.border.rounded.all
```
mbrandonw: components can just be unions of utility classes! no extra css at all

Tests/PointFreeTests/__Snapshots__/EpisodeTests/testEpisodePage.1.Conn.txt
```diff
@@ -9,11 +9,11 @@
 
 ▿ Response
   Status 200 OK
-  Content-Length: 25436
+  Content-Length: 27455
```
mbrandonw: can see that the episode page increased by a whopping 2k! most of that is from an svg in the page now, and then a bit from the styleguide and design systems additions.

Sources/Styleguide/DesignSystems/Spacing.swift
stephencelis: Cool!

Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -3,10 +3,18 @@ import Prelude
...
 public enum Colors {
   public static let black = Color.other("#121212")
-  public static let purple = Color.other("#974DFF")
   public static let blue = Color.other("#4CCCFF")
+  public static let gray300 = Color.other("#555555")
+  public static let gray400 = Color.other("#666666")
+  public static let gray650 = Color.other("#a8a8a8")
+  public static let gray850 = Color.other("#d8d8d8")
+  public static let gray900 = Color.other("#f6f6f6")
   public static let green = Color.other("#79F2B0")
```
stephencelis: Should we get rid of `green`?

Tests/LinuxMain.swift
```diff
@@ -4,80 +4,8 @@
 import XCTest
 
 @testable import PointFreeTests; @testable import StyleguideTests;
-extension AuthTests {
```
stephencelis: Linux turned off now?

Sources/PointFree/Episode/Episode.swift
```diff
@@ -83,75 +82,129 @@ private let videoView = View<Prelude.Unit> { _ in
...
+      [
+        h6(
+          [`class`([Class.pf.type.title6, Class.pf.colors.fg.gray850, Class.padding([.mobile: [.bottom: 1]])])],
+          ["Credits"]
+        ),
+        p(
+          [`class`([Class.pf.colors.fg.gray850])],
+          [
+            "Hosted by ",
+            a([`class`([Class.pf.colors.link.white]), href("mailto:brandon@pointfree.co")], ["Brandon Williams"]),
```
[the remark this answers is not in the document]
mbrandonw: do you think this should just take a string? otherwise you have to use it like:

```swift 
mailto(EmailAddress(unwrap: "brandon@pointfree.co"))
```

not sure what benefits that's giving...

Sources/Styleguide/PointFreeBaseStyles.swift
mbrandonw: we use it on the episode page for the TOC

Tests/LinuxMain.swift
mbrandonw: um whoops. dont know what happened here, looking into it

mbrandonw: ah, it was cause `TestCase` moved modules. fixed the stencil template to just look for `...TestCase` subclasses.

Package.swift
```diff
@@ -11,6 +11,7 @@ let package = Package(
     ],
   dependencies: [
     .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("b70de46")),
+    .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .revision("c510e7d")),
```
mbrandonw: I had to add this to fix this error on circle:

```
❌  /Users/distiller/project/Sources/PointFreeTestSupport/TestCase.swift:2:8: missing required module 'WKSnapshotConfigurationShim'

import SnapshotTesting
       ^
```

I dont really understand. it didnt happen locally.

PR #57, opened 2017-12-14
WIP Pricing module

mbrandonw:
Still lots to do (hooking up, mobile, etc...) but here's a decent chunk!


Makefile
```diff
@@ -4,7 +4,7 @@ imports = \
 
 bootstrap: xcodeproj postgres-mm sourcery
 
-xcodeproj: sourcery
```
mbrandonw: removing this for now cause sourcery is having troubles

PR #60, opened 2017-12-17
[wip] Team invites

mbrandonw:
### What

Beginnings of team invites! It's not actually hooked up to any real data models, but once we have that data in the subscriptions it will be quite easy to do. I have the following (mostly) supported:

* Create an invite
* Send email for invite
* See invites you've sent
* Resend an invite
* Revoke an invite
* "Accept" an invite, which just sends a confirmation email and deletes the invite. Doesn't actually add the person to the subscription.

invite


Sources/PointFree/Account/Account.swift
```diff
@@ -0,0 +1,24 @@
...
+import Foundation
+import Html
+import HtmlCssSupport
+import HttpPipeline
+import HttpPipelineHtmlSupport
+import Optics
+import Prelude
+import Styleguide
+import Tuple
+import UrlFormEncoding
```
stephencelis: Lol, wow, are more lines of import required than implementation?

Sources/PointFree/Account/Invite.swift
```diff
@@ -0,0 +1,137 @@
...
+private let showInviteView = View<Tuple2<Database.User?, Database.TeamInvite>> { data in
+
+  get1(data)
+    .map { showInviteLoggedInView.view($0 .*. get2(data)) }
+    ?? showInviteLoggedOutView.view(get2(data))
+}
+
+private let showInviteLoggedOutView = View<Database.TeamInvite> { invite in
+  [
+    p(["You must be logged in to accept this invitation. Would you like to log in with GitHub?"]),
```
stephencelis: Should we just redirect immediately?

Sources/PointFree/Auth.swift
```diff
@@ -62,14 +62,14 @@ extension URLRequest {
 
 public func readSessionCookieMiddleware<I, A>(
   _ conn: Conn<I, A>)
-  -> IO<Conn<I, Tuple2<A, Database.User?>>> {
+  -> IO<Conn<I, Tuple2<Database.User?, A>>> {
```
stephencelis: The swap feels right but curious what drove it!

Sources/PointFree/Auth.swift
```diff
@@ -62,14 +62,14 @@ extension URLRequest {
...
 
     let user = conn.request.cookies[pointFreeUserSession]
       .flatMap {
         ResponseHeader
           .verifiedString(signedCookieValue: $0, secret: AppEnvironment.current.envVars.appSecret)
       }
-      .flatMap(UUID.init >-> Database.User.Id.init)
+      .flatMap(UUID.init(uuidString:) >-> Database.User.Id.init)
```
stephencelis: Did this become ambiguous?

Sources/PointFree/Database.swift
```diff
@@ -114,20 +141,83 @@ private func fetchUser(byUserId id: Database.User.Id) -> EitherIO<Error, Databas
...
+private func fetchUser(byGitHubUserId userId: GitHub.User.Id) -> EitherIO<Error, Database.User?> {
   return firstRow(
     """
     SELECT "email", "github_user_id", "github_access_token", "id", "name"
     FROM "users"
-    WHERE "github_access_token" = $1
+    WHERE "github_user_id" = $1
     LIMIT 1
     """,
     // TODO: make this fetch by the github id, not access token, since it can change. also maybe build in
```
stephencelis: Can delete this TODO now, no? :D

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -34,6 +49,25 @@ private let routers: [Router<Route>] = [
   Route.iso.home
     <¢> get %> queryParam("success", opt(.bool)) <% end,
 
+  Route.iso.invite <<< Route.Invite.iso.accept
```
stephencelis: This is pretty amazing.

Sources/PointFree/Account/Account.swift
mbrandonw: prob lots of copy pasta. will clean up imports

Sources/PointFree/Account/Invite.swift
mbrandonw: I thought it would be nice to let them know why we are asking for GH auth first

Sources/PointFree/Auth.swift
mbrandonw: yeah cause I added `RawRepresentable` conformance to `UUID`

Sources/PointFree/Database.swift
mbrandonw: oh yup

Sources/PointFree/Auth.swift
mbrandonw: nah nothing in particular I dont think

Sources/PointFree/Account/Invite.swift
```diff
@@ -0,0 +1,158 @@
...
+                .run
+                .map(const(unit))
+              }
+              ?? pure(unit)
+      }
+
+      let deleteInvite = parallel
+        <| AppEnvironment.current.database.deleteTeamInvite(get2(conn.data).id).run
+
+      zip(sendInviterEmailOfAcceptance, deleteInvite).run({ _ in })
```
mbrandonw: just made an update to send an email to the inviter to let them know the invite was accepted, and then to delete the invite.

still need to add the current user to the inviter's subscription tho.

Sources/PointFree/Account/Invite.swift
```diff
@@ -14,7 +14,7 @@ let showInviteMiddleware =
   requireTeamInvite
     <| currentUserMiddleware
     >-> writeStatus(.ok)
-    >-> respond(showInviteView)
+    >-> respond(showInviteView.contramap(lower))
```
mbrandonw: realized we dont need to use `Tuple`'s in views cause we can just `contramap(lower)` to get a proper tuple that can be destructured in the view function

Sources/PointFree/Database.swift
```diff
@@ -99,7 +123,10 @@ private func upsertUser(withGitHubEnvelope envelope: GitHub.UserEnvelope) -> Eit
       envelope.gitHubUser.name
     ]
     )
-    .flatMap { _ in fetchUser(gitHubAccessToken: envelope.accessToken) }
+    .flatMap { _ in
+      AppEnvironment.current.database
+        .fetchUserByGitHub(envelope.gitHubUser.id)
```
mbrandonw: oh this was a bit of a gotcha. we gotta make sure to always use the app environment's functions and never directly use the `private` ones in this file.

Sources/PointFree/RequestContext.swift
```diff
@@ -42,47 +42,36 @@ public func requireUser<A>(
   -> Middleware<StatusLineOpen, ResponseEnded, A, Data> {
 
     return { conn in
-
-      let currentUser = extractedGitHubUserEnvelope(from: conn.request)
```
mbrandonw: I rewrote these helpers cause they weren't really working, and `readSessionCookieMiddleware` does all the main heavy lifting.

Sources/PointFree/RequestContext.swift
```diff
@@ -42,47 +42,36 @@ public func requireUser<A>(
...
-        ?? pure(nil)
-
-      return currentUser.flatMap { user in
-        user.map { conn.map(const($0 .*. conn.data)) |> middleware }
-          ?? (conn |> redirect(to: path(to: .login(redirect: conn.request.url?.absoluteString))))
+      (conn |> readSessionCookieMiddleware)
+        .flatMap { c in
+          c.data.first.map { user in
+            c.map(const(user .*. c.data.second))
+              |> middleware
```
mbrandonw: but also the way I rewrote it is wild. can prob find a way to improve this

Sources/PointFree/RequestContext.swift
```diff
@@ -42,47 +42,36 @@ public func requireUser<A>(
...
-//func currentRequestMiddleware<A, I>(
-//  _ conn: Conn<I, A>
-//  ) -> IO<Conn<I, T2<URLRequest, A>>> {
-//
-//  return pure <|
-//    conn.map(
-//      const(conn.request .*. conn.data)
-//  )
-//}
+  return conn |> readSessionCookieMiddleware
```
mbrandonw: we can just get rid of `currentUserMiddleware` cause it's the same as `readSessionCookieMiddleware`. might want a better name.

Sources/PointFree/TODO.swift
```diff
@@ -220,3 +220,37 @@ public func onclick<T>(javascript: StaticString) -> Attribute<T> {
...
+  }
+
+  public init?(rawValue: String) {
+    guard let uuid = UUID(uuidString: rawValue) else { return nil }
+    self = uuid
+  }
+}
+
+extension PartialIso where A == B.RawValue, B: RawRepresentable {
+  public static var _rawRepresentable: PartialIso {
```
mbrandonw: I think this should replace our `rawRepresentable`

Sources/PointFree/TODO.swift
```diff
@@ -220,3 +220,37 @@ public func onclick<T>(javascript: StaticString) -> Attribute<T> {
...
+    return .init(
+      apply: B.init(rawValue:),
+      unapply: ^\.rawValue
+    )
+  }
+}
+
+public func mapExcept<E, F, A, B>(_ f: @escaping (Either<E, A>) -> Either<F, B>) -> (EitherIO<E, A>) -> EitherIO<F, B> {
```
mbrandonw: at one point I made use of this, but doesnt seem like I have it anymore. tho im sure it could be useful

Sources/PointFree/__Generated__/DerivedPartialIsos.swift
```diff
@@ -109,6 +127,40 @@ import Prelude
           })
     }
   }
+  extension Route.Invite {
+    enum iso {
+          public static let accept = parenthesize <| PartialIso(
```
mbrandonw: glad we dont have to write these

Sources/PointFree/Account/Invite.swift
```diff
@@ -0,0 +1,158 @@
...
+    >-> writeStatus(.ok)
+    >-> respond(showInviteView.contramap(lower))
+
+let revokeInviteMiddleware =
+  requireTeamInvite
+    <<< requireUser
+    <| { conn in
+      AppEnvironment.current.database.deleteTeamInvite(get2(conn.data).id)
+        .run
+        .flatMap(const(conn |> redirect(to: path(to: .team))))
```
mbrandonw: sometimes these middlewares can compose nicely. here we stack two transformers for requiring the team invite (from the `id` passed from the route), then requiring the user, and then we can just inline a middleware for doing the delete logic.

mbrandonw:
just pushed this commit which now does a join on subscriptions and users to find all of a user's teammates! https://github.com/pointfreeco/pointfreeco/pull/60/commits/3d4432ae5f8a57d07a2d777f7b81a3d9ebe49714

Sources/PointFree/Account/Invite.swift
```diff
@@ -0,0 +1,208 @@
...
+        .flatMap { AppEnvironment.current.database.fetchSubscriptionById($0) }
+        .mapExcept(requireSome)
+        .flatMap { subscription in
+          AppEnvironment.current.database.addUserIdToSubscriptionId(currentUser.id, subscription.id)
+      }
+
+      // VERIFY: only do this if the invite was successfully taken
+      let deleteInvite = parallel(
+        subscription
+          // TODO: should `const` be @autoclosure so that we can do:
```
stephencelis: I think we probably need a way of doing both strict and lazy evaluation. Scalaz 8 IO does this by having both `IO.now` (strict) and `IO.point` (lazy). The applicative `pure` is actually `now` (for performance, I think), so I bet `const` would want to strict, too, with how often we use it.

PR #61, opened 2017-12-19
Database tests

stephencelis:
Still not sure how we'll manage this infrastructure. The tests run and we can write assertions, though they're a bit of a pain since we don't have helper for executing `Either{IO}`.


Sources/PointFree/Database.swift
```diff
@@ -99,7 +99,7 @@ private func upsertUser(withGitHubEnvelope envelope: GitHub.UserEnvelope) -> Eit
       envelope.gitHubUser.name
     ]
     )
-    .flatMap { _ in fetchUser(gitHubAccessToken: envelope.accessToken) }
+    .flatMap { _ in AppEnvironment.current.database.fetchUserByGitHub(envelope.accessToken) }
```
mbrandonw: I pulled this in from my other PR.

not even sure what the future of our DB mock is, but either way seems like we should just always use the environment's database endpoints

Sources/PointFreeTestSupport/TestCase.swift
```diff
@@ -1,13 +1,38 @@
...
+
+    _ = try! execute("DROP SCHEMA public CASCADE;")
+      .flatMap(const(execute("CREATE SCHEMA public;")))
+      .flatMap(const(execute("GRANT ALL ON SCHEMA public TO pointfreeco;")))
+      .flatMap(const(execute("GRANT ALL ON SCHEMA public TO public;")))
+      .flatMap(const(AppEnvironment.current.database.migrate()))
+      .flatMap(const(execute("CREATE SEQUENCE test_uuids")))
+      .flatMap(const(execute(
+        """
+        CREATE OR REPLACE FUNCTION uuid_generate_v1mc() RETURNS uuid AS $$
```
mbrandonw: beautiful!

PR #63, opened 2017-12-20
Some of Cornett's newest designs

mbrandonw:
Implementing the new header and footer, and did some clean up.


Sources/PointFree/About.swift
```diff
@@ -1,25 +1,27 @@
...
 import Html
 import HtmlCssSupport
 import HttpPipeline
 import HttpPipelineHtmlSupport
 import Prelude
 import Styleguide
+import Tuple
 
 let aboutResponse: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
-  requestContextMiddleware
```
mbrandonw: was able to entirely get rid of the request context stuff now that we load all of that stuff upfront in the site middleware.

Sources/PointFree/Auth.swift
```diff
@@ -123,22 +123,28 @@ private func writeSessionCookieMiddleware(
...
-//        .map(fetchDatabaseUser)
-//        ?? pure(nil)
-//
-//      return currentUser.flatMap { user in
-//        user.map { conn.map(const($0 .*. conn.data .*. unit)) |> middleware }
-//          ?? (conn |> redirect(to: path(to: .login(redirect: conn.request.url?.absoluteString))))
-//      }
-//    }
-//}
+public func requireUser<A>(
```
mbrandonw: moved this stuff from `RequestContext.swfit` since I deleted that file

PR #65, opened 2017-12-21
Fetch subscription info from Stripe

stephencelis:
This fetches subscription/account info to render on the account page!

Still need to render currency and dates according to locale and currency code!


stephencelis:
Linux ish 😞

Just looked at the time so gonna crash! Will look with fresh eyes tomorrow.

Sources/PointFree/Account/Account.swift
```diff
@@ -249,29 +264,47 @@ private let subscriptionInviteMoreRowView = View<Prelude.Unit> { _ in
...
         ])
       ])
-    ])
+  ]
+}
+
+private let dueDateFormatter = DateFormatter()
+  |> \.dateStyle .~ .short
+  |> \.timeStyle .~ .none
+  |> \.timeZone .~ TimeZone(secondsFromGMT: 0)
```
stephencelis: I guess Linux runs in a diff time zone! We should prob figure out what to do for this but it can wait.

PR #68, opened 2017-12-21
Hook up invites to subscription model

mbrandonw:
Now that we have live subscription data I can derive invites stuff from that info.

This also makes it possible to remove someone from a team.

While talking about this today I realized we also need to show this info to teammate and allow them to leave the team. I may add that to this PR or do it separately.


Sources/PointFree/Database.swift
```diff
@@ -120,9 +122,9 @@ private func add(userId: Database.User.Id, toSubscriptionId subscriptionId: Data
   return execute(
     """
     UPDATE "users"
-    SET "subscription_id" = $1,
+    SET "subscription_id" = $2,
         "updated_at" = NOW()
-    WHERE "users"."id" = $2
+    WHERE "users"."id" = $1
```
mbrandonw: whoops!

Sources/PointFree/Account/Account.swift
```diff
@@ -233,16 +233,23 @@ private let inviteRowView = View<Database.TeamInvite> { invite in
...
-      ]),
-    gridColumn(sizes: [.mobile: 10], [
-      div([`class`([Class.padding([.mobile: [.leftRight: 1]])])], [
-        p(["You have 10 open spots on your team. Invite a team member below:"]),
+private let subscriptionInviteMoreRowView = View<(Stripe.Subscription?, [Database.TeamInvite], [Database.User])> { subscription, invites, teammates -> [Node] in
+
+  guard let subscription = subscription else { return [] }
+  guard subscription.quantity > 1 else { return [] }
+  let invitesRemaining = subscription.quantity - invites.count - teammates.count
+  guard invitesRemaining > 0 else { return [] }
```
stephencelis: Nice way of short-circuiting this view!

Sources/PointFree/Database.swift
stephencelis: Lol! Maybe flip the args instead?

mbrandonw:
will merge once I get the green!

PR #69, opened 2017-12-22
Update payment info

stephencelis:
Still WIP. Sneaking some account page stuff in too.


stephencelis:
So, this PR meandered a bit but should prob be merged before the title feature finishes. It:

- Hooks up user profile update
- Stubs out cancellation
- Stubs out cancellation confirmation
- Starts implementing the payment info update view
- Some design tweaks for the account page

Sources/PointFree/Account/Account.swift
```diff
@@ -91,12 +91,12 @@ private let profileRowView = View<Database.User> { currentUser in
       div([
         h2([`class`([Class.pf.type.title4])], ["Profile"]),
 
-        form([action("#"), method(.post)], [
+        form([action(path(to: .updateProfile(nil))), method(.post)], [
           label([`class`([labelClass])], ["Name"]),
-          input([`class`([blockInputClass]), type(.text), value(currentUser.name)]),
+          input([`class`([blockInputClass]), name("name"), type(.text), value(currentUser.name)]),
```
mbrandonw: whoa, could we actually use `Database.User.CodingKeys.name.rawValue` here?

Sources/PointFree/Account/Account.swift
```diff
@@ -340,17 +411,17 @@ private let logoutView = View<Prelude.Unit> { _ in
...
 
 private let subscriptionInfoRowClass =
-  Class.border.bottom
+  Class.border.top
     | Class.pf.colors.border.gray800
-    | Class.padding([.mobile: [.topBottom: 3]])
+    | Class.padding([.mobile: [.top: 2, .bottom: 3]])
 
-private let labelClass =
+let labelClass =
```
mbrandonw: we could now moves these into the base styles too now that we are using it more.

I have a simple idea to slightly organize that better if you wanna chat about it today

Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -0,0 +1,22 @@
...
+}
+
+let updateProfileMiddleware =
+  requireUser
+    <| { (conn: Conn<StatusLineOpen, Tuple2<Database.User, ProfileData?>>) -> IO<Conn<ResponseEnded, Data>> in
+      let (user, data) = lower(conn.data)
+
+      return pure(data)
+        .mapExcept(requireSome)
+        .flatMap { AppEnvironment.current.database.updateUser(user.id, $0.name, $0.email) }
```
mbrandonw: we should prob have them confirm their email change, but I think there's a very simple stateless way of doing it. we can send an email with a confirmation link that has a signed and encrypted value inside it, and we can use that to confirm the change.

no need to do it now, but I'll add a Trello card

Sources/PointFree/Database.swift
```diff
@@ -4,18 +4,19 @@ import Prelude
 import PostgreSQL
 
 public struct Database {
-  var addUserIdToSubscriptionId: (Database.User.Id, Database.Subscription.Id) -> EitherIO<Error, Prelude.Unit>
```
mbrandonw: ah nice, get rid of some clutter

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -10,18 +10,22 @@ public protocol DerivePartialIsos {}
 public enum Route: DerivePartialIsos {
   case about
   case account
+  case cancel
+  case confirmCancel
```
mbrandonw: we could start nesting some of these routes inside an `Account` sub route...

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -34,6 +34,12 @@ private func render(conn: Conn<StatusLineOpen, Tuple2<Database.User?, Route>>)
       return conn.map(const(unit))
         |> accountResponse
 
+    case .cancel:
+      fatalError()
+
+    case .confirmCancel:
+      fatalError()
```
mbrandonw: that's one way to keep people form canceling!

Sources/PointFree/StripeHtml.swift
```diff
@@ -0,0 +1,88 @@
...
+          """
+        )
+      ]
+    }
+  }
+}
+
+private let stripeInputClass =
+  regularInputClass
+    | Class.flex.column
```
mbrandonw: nice nice nice!

Sources/PointFree/Account/Account.swift
```diff
@@ -149,32 +149,125 @@ private func planName(for subscription: Stripe.Subscription) -> String {
...
+              ])
+            ]),
+          gridColumn(sizes: [.mobile: 12, .desktop: 6], [
+            div([`class`([Class.padding([.mobile: [.leftRight: 1]]), Class.grid.end(.desktop)])], [
+              p([
+                a([
+                  `class`([Class.pf.components.button(color: .purple, size: .small)]),
+                  href("#")
+                  ],
+                  ["Cancel"])
```
mbrandonw: I need to update our style guide for text only buttons. feel this button should prob not be accepted like all the other positive-oriented actions, and should prob be red.

Tests/PointFreeTests/PaymentInfoTests.swift
```diff
@@ -0,0 +1,44 @@
...
+
+  func testRender() {
+    let conn = connection(from: request(to: url(to: .paymentInfo)))
+    let result = conn |> siteMiddleware
+
+    assertSnapshot(matching: result.perform())
+
+    #if !os(Linux)
+      if #available(OSX 10.13, *) {
+        let webView = WKWebView(frame: .init(x: 0, y: 0, width: 1080, height: 2000))
```
mbrandonw: these screenshots seem to be much smaller than the others, so could cut the size of em down

Sources/PointFree/Routers/PointFreeRoutes.swift
stephencelis: Oh definitely. I started going down that path but it was too late for me to understand how that works for the first time. Can refactor today!

Sources/PointFree/Account/Account.swift
stephencelis: Yep! Just Gott make them public. Nested keys we need another solution for, tho.

Tests/PointFreeTests/PaymentInfoTests.swift
stephencelis: Yeah, this view isn't ready yet. Will definitely fix the height when I finish up tho.

stephencelis:
Gonna merge and address feedback in a future PR!

PR #70, opened 2017-12-22
Css style inliner

mbrandonw:
### What

A simple way of inlining a stylesheet's styles into an html document! This is not a full solution as we do not attempt to match _any_ kind of selector to a node, however it works beautifully with the stylesheets that we care most about.

I have a test that shows a good use case, but here it is for posterity:

```swift 
let stylesheet =
  body % fontSize(.px(16))
    <> p % color(.black)
    <> "#hero" % maxWidth(.pct(100))
    <> "#some-id" % backgroundColor(.red)
    <> ".p1" % padding(all: .px(16))
    <> ".bold" % fontWeight(.bold)
    <> ".leading" % fontSize(.px(18))
    <> (body | html) % height(.pct(100))
    <> ("#footer" | footer | ".footer") % display(.block)

let doc = document([
  html([
    body([
      div([id("hero")], [
        h1(["Point-Free"])
        ]),

      p([Html.`class`("p1")], [
        "This p tag gets styles from the tag element and classes."
        ]),

      p([id("some-id"), Html.`class`("bold leading"), style(lineHeight(1.25))], [
        "This p tag gets styles from the id, classes, inline styles, and element tag!"
        ]),

      div([id("footer")], [
        "I'm a div footer with an id"
        ]),
      div([Html.`class`("footer")], [
        "I'm a div footer with a class"
        ]),
      footer([
        "I'm a footer element"
        ])
      ])
    ])
  ])
```

Applying the `applyInlineStyles` function to this document and stylesheet gives us:

```html
<!DOCTYPE html>
<html style="height:100%;">
  <body style="font-size:16px; height:100%;">
    <div id="hero" style="max-width:100%;">
      <h1>
        Point-Free
      </h1>
    </div>
    <p class="p1"
       style="padding-top:16px; padding-right:16px; padding-bottom:16px; padding-left:16px; color:#000000;">
      This p tag gets styles from the tag element and classes.
    </p>
    <p id="some-id"
       class="bold leading"
       style="line-height:1.25; font-weight:bold; font-size:18px; color:#000000; background-color:#ff0000;">
      This p tag gets styles from the id, classes, inline styles, and element tag!
    </p>
    <div id="footer" style="display:block;">
      I&#39;m a div footer with an id
    </div>
    <div class="footer" style="display:block;">
      I&#39;m a div footer with a class
    </div>
    <footer style="display:block;">
      I&#39;m a footer element
    </footer>
  </body>
</html>
```

### DIFF
--- a/Sources/PointFree/HtmlCssInliner.swift
+++ b/Sources/PointFree/HtmlCssInliner.swift
@@ -1,0 +1,157 @@
+@testable import Css
+import Html
+import Optics
+import Prelude
+
+/// Transforms `nodes` passed in by applying a reasonable number of styles from `stylesheet` to each
+/// element.
+public func applyInlineStyles(nodes: [Node], stylesheet: Stylesheet) -> [Node] {
+  return nodes.map { applyInlineStyles(node: $0, stylesheet: stylesheet) }
+}
+
+/// Transforms `node` by applying a reasonable number of the styles from `stylesheet` to each element.
+public func applyInlineStyles(node: Node, stylesheet: Stylesheet) -> Node {
+  switch node {
+  case .comment:
+    return node
+
+  case let .document(nodes):
+    return .document(applyInlineStyles(nodes: nodes, stylesheet: stylesheet))
+
+  case let .element(element):
+    return .element(applyInlineStyles(element: element, stylesheet: stylesheet))
+
+  case .text:
+    return node
+  }
+}
+
+private func applyInlineStyles(element: Element, stylesheet: Stylesheet) -> Element {
+  let currentStyles = element.attribs
+    .first(where: { $0.key == "style" })?
+    .value
+    .renderedValue()?
+    .string
+    ?? ""
+
+  // Computes all inline styles based on the classes on this element
+  let classStyles = (
+    element.attribs
+      .first(where: { $0.key == "class" })?
+      .value
+      .renderedValue()?
+      .string
+      ?? ""
+    )
+    .components(separatedBy: " ")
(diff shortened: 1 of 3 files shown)


Sources/PointFree/HtmlCssInliner.swift
```diff
@@ -0,0 +1,157 @@
+@testable import Css
```
mbrandonw: need to make some stuff public in `Css`

Sources/PointFree/HtmlCssInliner.swift
```diff
@@ -0,0 +1,157 @@
...
+    if doesApp(app, contain: selector) {
+      return rules
+        .map { inlineStyles(for: selector, rule: $0) }
+        .joined(separator: ";")
+    }
+    return ""
+
+  case .query, .face, .keyframe, .import:
+    // TODO: what to do about `query` here? Media queries can't be used in inline styles, but we might still
+    //       want to render its styles?
```
stephencelis: Huh! This comment makes me wonder where inlining styles in attributes is better than just embedding `style` tags.

mbrandonw: some email clients support style tags, but sometimes only a subset of things inside that tag. and even with the style tag things can get overridden by the client if the style isnt actually inlined. it's a wild world! 

thinking we are in a good sitch by just rendering inline and providing the style tag. espesh since we use 0 cascading in our stylesheets

mbrandonw:
travis was having some troubles with complex expressions, but now I think we have the green. gonna merge!

PR #73, opened 2017-12-27
Cancellation/Reactivation

stephencelis:
This is mostly functionality (for now). Still needs more tests and design.


Sources/PointFree/Account/Account.swift
```diff
@@ -10,7 +10,7 @@ import Styleguide
 import Tuple
 
 let accountResponse =
-  requireUser
+  _requireUser
```
stephencelis: Renamed the "bad" `requireUser` to `_requireUser` and made `requireUser` handle `T{N}` instead of `Tuple{N}`. I think we'll wanna change to use the `T{N}` system everywhere eventually.

Sources/PointFree/Account/Account.swift
```diff
@@ -93,10 +93,20 @@ private let profileRowView = View<Database.User> { currentUser in
 
         form([action(path(to: .updateProfile(nil))), method(.post)], [
           label([`class`([labelClass])], ["Name"]),
-          input([`class`([blockInputClass]), name("name"), type(.text), value(currentUser.name)]),
+          input([
+            `class`([blockInputClass]),
+            name(Database.User.CodingKeys.name.stringValue),
```
stephencelis: Using coding keys here now!

Sources/PointFree/Account/Account.swift
```diff
@@ -205,7 +215,7 @@ private let subscriptionPlanRows = View<Stripe.Subscription> { subscription in
                   `class`([Class.pf.components.button(color: .purple, size: .small)]),
                   href("#")
                   ],
-                  ["Upgrade"])
+                  ["Upgrade"]) // TODO: disable when subscription.status == .canceled
```
stephencelis: We prob want a disabled style. And/or a `button(to:)` helper that knows when to be a button, when to be a link, when to be disabled, etc.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+      let (subscription, data) = conn.data
+
+      return AppEnvironment.current.stripe.reactivateSubscription(subscription)
+        .run
+        .map(^\.right)
+        .map(const(conn.map(const(unit))))
+    }
+    >-> redirect(to: .account)
+
+func requireUser<A>(
```
stephencelis: The new `requireUser` lives in **Cancel.swift** for now.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+        ]),
+      body(
+        darkNavView.view((currentUser, nil))
+          <> [
+            gridRow([
+              gridColumn(sizes: [.mobile: 12, .desktop: 8], [style(margin(leftRight: .auto))], [
+                div(
+                  [`class`([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]])])],
+                  titleRowView.view(unit)
+                    <> formRowView.view(unit)
```
stephencelis: This is a pretty basic screen right now (just a title and a big ol button). Gonna add subscription details here, copy around what unsubscribing means, and maybe a set of radio buttons for feedback.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+              ])
+          ]
+          <> footerView.view(unit)
+      )
+      ])
+    ])
+  ])
+}
+
+private let titleRowView = View<Prelude.Unit> { _ in
```
stephencelis: Probably want to extract some of these row styles to `Styleguide`.

Sources/PointFree/Auth.swift
```diff
@@ -46,21 +46,7 @@ let logoutResponse: (Conn<StatusLineOpen, Prelude.Unit>) -> IO<Conn<ResponseEnde
...
-        $0.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
-          .map(String.init)
-      }
-      .flatMap { tuple <¢> $0.first <*> $0.last }
-    return .init(uniqueKeysWithValues: pairs)
-  }
-}
-
-public func readSessionCookieMiddleware<I, A>(
+public func _readSessionCookieMiddleware<I, A>(
```
stephencelis: This still uses `Tuple2` but should use `T2`.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -21,6 +21,7 @@ public enum Route: DerivePartialIsos {
   case logout
   case paymentInfo
   case pricing(String?, Int?)
+  case reactivate
```
stephencelis: Added a `reactivate` endpoint for subscriptions that are canceling this period but haven't canceled yet.

Sources/PointFree/Stripe.swift
```diff
@@ -213,18 +228,40 @@ private func createSubscription(customer: Stripe.Customer.Id, plan: Stripe.Plan.
...
+
+    guard
+      subscription.cancelAtPeriodEnd,
+      let item = subscription.items.data.first
+      else { return throwE(unit) }
+
+    return updateSubscription(subscription, item.plan.id, item.quantity)
+}
+
+private func updateSubscription(_ subscription: Stripe.Subscription, _ plan: Stripe.Plan.Id, _ quantity: Int)
```
stephencelis: Thinking we may wanna move back to passing values instead of ids. We never manually build ids (outside of tests), so we have the full values at our disposal. In the case of subscription updates, we need quite a bit of data from the subscription structure:

- Subscription id
- Subscription item id
- Subscription item plan (for reactivation)
- Subscription item quantity (for reactivation)

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+          .map(^\.right)
+          .flatMap { sub in
+
+            guard let sub = sub, suchThat(sub) else {
+              return conn
+                |> writeStatus(.notFound)
+                >-> respond(text: "Stripe subscription not found :(")
+            }
+
+            return conn.map(const(.init(first: sub, second: .init(first: user, second: rest))))
```
mbrandonw: Im hoping this can just be `sub .*. user .*. rest` once we merge prelude stuff

Sources/PointFree/Stripe.swift
```diff
@@ -189,7 +204,7 @@ extension Tagged where Tag == Stripe.Plan, A == String {
 //  .map(^\.data >>> filter(StripeSubscriptionPlan.Id.all.contains <<< ^\.id))
 
 private func cancelSubscription(id: Stripe.Subscription.Id) -> EitherIO<Prelude.Unit, Stripe.Subscription> {
-  return stripeDataTask("subscriptions/\(id.rawValue)", .delete)
+  return stripeDataTask("subscriptions/" + id.rawValue, .delete(["at_period_end": "true"]))
```
mbrandonw: are you just leaning away from interpolation or is there another reason for this?

Sources/PointFree/Stripe.swift
mbrandonw: gotcha gotcha. I forget why I started doing that, but we can try it and see what happens

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+      ])
+    ])
+}
+
+private let formRowView = View<Prelude.Unit> { _ in
+  gridRow([`class`([Class.padding([.mobile: [.bottom: 4]])])], [
+    gridColumn(sizes: [.mobile: 12], [
+      form([action(path(to: .cancel)), method(.post)], [
+        button(
+          [`class`([Class.pf.components.button(color: .red), Class.margin([.mobile: [.top: 3]])])],
```
mbrandonw: was thinking we could just use an underline style for this button instead of a full on button, to not make it stand out and seem too enticing.

we could add a `style` param to buttons so that  we could do `button(color: .purple, style: .filled)` vs `button(color: .red, style: .underlined)` vs `button(color: .red, style: .outlined)`.

mbrandonw: sorry this button is good, I meant for the first cancel button.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+        .map(^\.right)
+        .map(const(conn.map(const(unit))))
+    }
+    >-> redirect(to: .account)
+
+let reactivateMiddleware =
+  requireUser
+    <<< requireSubscription
+    <<< requireSubscriptionOwner
+    <<< requireStripeSubscription(^\.cancelAtPeriodEnd)
```
mbrandonw: very cooooool!

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+    <| map(lower)
+    >>> { conn -> IO<Conn<StatusLineOpen, Prelude.Unit>> in
+      let (subscription, data) = conn.data
+
+      return AppEnvironment.current.stripe.reactivateSubscription(subscription)
```
mbrandonw: wanna add some `// TODO`s or create some Trello cards for all the emails we gotta send with these?

Sources/PointFree/Account/Cancel.swift
```diff
@@ -0,0 +1,192 @@
...
+    <<< requireStripeSubscription(^\.status != .canceled)
+    <| writeStatus(.ok)
+    >-> respond(confirmCancelView.contramap(lower))
+
+let cancelMiddleware =
+  requireUser
+    <<< requireSubscription
+    <<< requireSubscriptionOwner
+    <<< requireStripeSubscription(^\.status != .canceled)
+    <| map(lower)
```
mbrandonw: ah nice nice

Sources/PointFree/Stripe.swift
stephencelis: Yeah, leaning away to avoid bugs.

PR #76, opened 2017-12-28
Update dependencies and use new tuple stuff

mbrandonw:
### What

Updates all the dependencies, and updates middleware to properly understand tuples.

I have a bit of clean up I need to do, but need to step away for a bit. It's mostly all here tho.

### DIFF
--- a/Sources/PointFree/About.swift
+++ b/Sources/PointFree/About.swift
@@ -8,12 +8,16 @@ import Prelude
 import Styleguide
 import Tuple
 
+public func lower<A>(_ tuple: Tuple1<A>) -> A {
+  return get1(tuple)
+}
+
 let aboutResponse: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
   currentUserMiddleware
     >-> writeStatus(.ok)
     >-> respond(aboutView.contramap(lower))
 
-private let aboutView = View<(Database.User?, Prelude.Unit)> { currentUser, _ in
+private let aboutView = View<Database.User?> { currentUser in
   document([
     html([
       head([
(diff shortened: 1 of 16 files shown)


Sources/PointFree/About.swift
```diff
@@ -13,7 +13,7 @@ let aboutResponse: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data>
     >-> writeStatus(.ok)
     >-> respond(aboutView.contramap(lower))
 
-private let aboutView = View<(Database.User?, Prelude.Unit)> { currentUser, _ in
```
mbrandonw: we're able to get rid of lots of needless trailing `Unit`s!

Sources/PointFree/Account/Account.swift
```diff
@@ -10,16 +10,16 @@ import Styleguide
 import Tuple
 
 let accountResponse =
-  _requireUser
+  filterMap(require1)
```
mbrandonw: went ahead and renamed to `filterMap`

Sources/PointFree/Account/Cancel.swift
```diff
@@ -59,16 +59,14 @@ func requireUser<A>(
...
-          let (optionalUser, rest) = ($0.data.first, $0.data.second.first)
-
-          guard let user = optionalUser else {
+          guard let user = get1($0.data) else {
             return $0
               |> redirect(to: .login(redirect: $0.request.url?.absoluteString))
           }
 
-          return $0.map(const(.init(first: user, second: rest)))
+          return $0.map(over1(const(user)))
```
stephencelis: Neato!

PR #78, opened 2017-12-30
Start shaving off expression times

stephencelis:
Got a lot of things that exponentially affect compile time :(

### DIFF
--- a/Development.xcconfig
+++ b/Development.xcconfig
@@ -1,1 +1,1 @@
-OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=500
+OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=400
--- a/Development.xcconfig
+++ b/Development.xcconfig
@@ -1,0 +1,1 @@
+OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=500
--- a/Sources/PointFree/Account/Cancel.swift
+++ b/Sources/PointFree/Account/Cancel.swift
@@ -9,18 +9,25 @@ import Prelude
 import Styleguide
 @testable import Tuple
 
+func requireSubscriptionAndOwner<A>(
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, T3<Database.Subscription, Database.User, A>, Data>
+  )
+  -> Middleware<StatusLineOpen, ResponseEnded, T2<Database.User?, A>, Data> {
+
+    return filterMap(require1 >>> pure, or: loginAndRedirect)
+      <<< requireSubscription
+      <<< requireSubscriptionOwner
+      <| middleware
+}
+
 let confirmCancelResponse =
-  filterMap(require1 >>> pure, or: loginAndRedirect)
-    <<< requireSubscription
-    <<< requireSubscriptionOwner
+  requireSubscriptionAndOwner
     <<< requireStripeSubscription(^\.status != .canceled)
     <| writeStatus(.ok)
     >-> respond(confirmCancelView.contramap(lower))
 
 let cancelMiddleware =
-  filterMap(require1 >>> pure, or: loginAndRedirect)
-    <<< requireSubscription
-    <<< requireSubscriptionOwner
+  requireSubscriptionAndOwner
     <<< requireStripeSubscription(^\.status != .canceled)
     <| map(lower)
     >>> { conn -> IO<Conn<StatusLineOpen, Prelude.Unit>> in
@@ -36,9 +43,7 @@ let cancelMiddleware =
     >-> redirect(to: .account)
 
 let reactivateMiddleware =
-  filterMap(require1 >>> pure, or: loginAndRedirect)
-    <<< requireSubscription
-    <<< requireSubscriptionOwner
+  requireSubscriptionAndOwner
     <<< requireStripeSubscription(^\.cancelAtPeriodEnd)
     <| map(lower)
     >>> { conn -> IO<Conn<StatusLineOpen, Prelude.Unit>> in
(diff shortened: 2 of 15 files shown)


Development.xcconfig
```diff
@@ -0,0 +1 @@
+OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=500
```
stephencelis: Starting with 500...

Sources/PointFree/Account/Cancel.swift
```diff
@@ -9,18 +9,25 @@ import Prelude
 import Styleguide
 @testable import Tuple
 
+func requireSubscriptionAndOwner<A>(
```
stephencelis: All of the middleware below took several seconds to compile before extracting these shared filters. Now it's under 500 ms.

Sources/PointFree/Admin/Index.swift
```diff
@@ -62,18 +64,15 @@ private let newEpisodeEmailRowView = View<Episode> { ep in
     ])
 }
 
-let sendNewEpisodeEmailMiddleware =
-  filterMap(require1 >>> pure, or: loginAndRedirect)
-    <<< requireAdmin
-    <<< filterMap(fetchEpisode >>> pure, or: redirect(to: .admin(.newEpisodeEmail(.show))))
-    <| { conn in pure(conn.map(get2)) }
-    >-> sendNewEpisodeEmails
+let sendNewEpisodeEmailMiddleware:
```
stephencelis: This was taking 20 seconds to compile for me :(

I cleaned up some of the filtering logic and extracted some other stuff and now it's under 500 ms.

Sources/PointFree/Emails/Mailgun.swift
```diff
@@ -47,20 +47,17 @@ public struct SendEmailResponse: Decodable {
...
-    "subject": email.subject,
-    "text": email.text,
-    "html": email.html,
-    "tracking": email.tracking?.rawValue,
-    "tracking-clicks": email.trackingClicks?.rawValue,
-    "tracking-opens": email.trackingOpens?.rawValue
-    ]
-    |> compact
-
+  var params: [String: String] = [:]
```
stephencelis: A lil mutation never hurt anyone. 9 seconds to sub-500 ms.

Sources/PointFree/LaunchSignup/Views.swift
```diff
@@ -153,18 +153,15 @@ let responsiveSuccessStyle =
     <> bottom(0)
 
 let responsiveStyle = queryOnly(screen, [minWidth(.px(800))]) {
-  ".hero, .signup, .success" % (
```
stephencelis: Another 9ish second bit of code down to under 500 ms.

Sources/PointFree/Admin/Index.swift
```diff
@@ -15,19 +15,22 @@ private let adminEmails = [
...
-  ) -> Middleware<StatusLineOpen, ResponseEnded, T2<Database.User, A>, Data> {
+  )
+  -> Middleware<StatusLineOpen, ResponseEnded, T2<Database.User?, A>, Data> {
 
-  return { conn in
-    conn
-      |> (adminEmails.contains(get1(conn.data).email.unwrap) ? middleware : redirect(to: .secretHome))
-  }
+    return filterMap(require1 >>> pure, or: loginAndRedirect)
+      <<< filter(get1 >>> isAdmin, or: redirect(to: .secretHome))
```
stephencelis: A new `filter` :D

Sources/PointFree/TODO.swift
```diff
@@ -125,6 +125,16 @@ public func filterMap<A, B>(
     }
 }
 
+public func filter<A>(
```
stephencelis: Was using this on another branch and wanted to use it to clean up above!

stephencelis:
Build times have quite a range but this first commit shaves off about 2 min of build time!

image

image

Sources/PointFree/Footer.swift
```diff
@@ -8,15 +8,23 @@ let footerView =
   curry(footer)([`class`([footerClass])]) >>> pure
     <¢> footerInfoColumnsView
 
+private func column(sizes: [Breakpoint: Int]) -> ([Node]) -> [Node] {
+  return gridColumn(sizes: sizes) >>> pure
```
stephencelis: Been finding that `pure` inference is having a tough time. Should talk about making them `static`s in Prelude.

Sources/PointFree/Account/Cancel.swift
mbrandonw: dang. well at least its nice to see how easy it is to define new transformers out of existing!

Sources/PointFree/Admin/Index.swift
mbrandonw: nice! I like!

Sources/PointFree/Footer.swift
mbrandonw: yeah makes sense to me. could start with code gen'ing them both. think we should figure out a code gen sitch for us anyway.

Sources/PointFree/LaunchSignup/Views.swift
mbrandonw: yeeeesh

Development.xcconfig
```diff
@@ -0,0 +1 @@
+OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=400
```
mbrandonw: does `400` give us mostly no warnings?

stephencelis: Still about 10 or so :(

mbrandonw: Ah I was gonna say we should raise it till the warnings go away and then we can lower if we wanna do another pass later

stephencelis: Ah, it's not too noisy at 400, though, and it only shows up for `make xcodeproj`. Maybe it'll motivate us to optimize a few along the way :)

PR #79, opened 2017-12-30
Layouts and flash messages

stephencelis:
image

This PR solidifies shared layouts and flash messaging.


Sources/PointFree/Account/Cancel.swift
```diff
@@ -9,63 +9,100 @@ import Prelude
 import Styleguide
 @testable import Tuple
 
-func requireSubscriptionAndOwner<A>(
```
stephencelis: General cleanup in this file to use `filterMap` and `filter` instead of more ad-hoc, logic-full transformers.

Sources/PointFree/Account/Account.swift
```diff
@@ -9,69 +9,11 @@ import Prelude
...
-      title: "Account"
-    )
-}
-
 let accountResponse =
   filterMap(require1 >>> pure, or: loginAndRedirect)
     <| fetchAccountData
     >-> writeStatus(.ok)
-    >-> respond(wrappedAccountView.contramap(lower))
+    >-> respond(accountView.contramap(lower), layout: simplePageLayout(title: "Account", currentUser: get4))
```
stephencelis: This is what rendering with layout looks like. May change to better fit our needs in the future!

Sources/PointFree/Account/Account.swift
```diff
@@ -9,69 +9,11 @@ import Prelude
 import Styleguide
 import Tuple
 
-public struct Flash: Codable {
```
stephencelis: (Moved to **Session.swift**.)

Sources/PointFree/Account/Cancel.swift
```diff
@@ -9,63 +9,100 @@ import Prelude
...
-}
+// MARK: Middleware
 
 let confirmCancelResponse =
-  requireSubscriptionAndOwner
-    <<< requireStripeSubscription(^\.status != .canceled)
+  requireStripeSubscription
+    <<< filter(
+      get1 >>> isActive,
+      or: redirect(to: .account, headersMiddleware: flash(.error, "Your subscription is already canceled!"))
```
stephencelis: Flashing a message is just headers middleware on a redirect!

Sources/PointFree/PageLayout.swift
```diff
@@ -0,0 +1,81 @@
...
+import Html
+import HtmlCssSupport
+import HttpPipeline
+import HttpPipelineHtmlSupport
+import Optics
+import Prelude
+import Styleguide
+import Tuple
+
+func respond<A>(_ view: View<A>, layout: @escaping (Flash?, View<A>) -> View<A>)
```
stephencelis: This interface is maybe a lil bit clunky. Suggestions for polish are welcome!

Sources/PointFree/Session.swift
```diff
@@ -1,25 +1,21 @@
...
-      Response.Header
+      let value = update(conn.request.session)
+      guard value != conn.request.session else { return pure(conn) }
+      return Response.Header
         .setSignedCookie(
           key: pointFreeUserSession,
-          value: update(conn.request.session),
-          options: [],
+          value: value,
+          options: [.path("/")],
```
stephencelis: We're writing to the session more now (it used to be just once, on login), and I found that browsers were setting different paths for different cookies. I've hard-coded the root path here to avoid multiple sessions from proliferating.

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -203,17 +204,52 @@ extension Stripe.Subscription.Item {
...
-  9146d686873dfde259849eb38151c9f448d718ff57b50f22d67cc9c37c9acca2\
-  b8da4c3e919a230d2b2d4222dedc4e9c354d84975c9e5f532503e6be65314e30\
-  0f74e793d154bf8f3c85054ae8df44717e856dc441f24a070cad129b4072a87a
-  """
+extension Session {
+  public static let mock = empty
+    |> \.userId .~ Database.User.mock.id
+}
+
+public func authedRequest(to route: Route, session: Session = .mock) -> URLRequest {
```
stephencelis: This can prob be further refactored, but found it really helpful to be able to test with the session in a specific state! How do you think we should expose the `private` helpers below, though?

Sources/PointFree/Account/Cancel.swift
```diff
@@ -74,60 +111,28 @@ func requireSubscription<A>(
...
 
     return { conn in
-      guard conn.data.first.userId == conn.data.second.first.id else {
-        return conn
-          |> writeStatus(.notFound)
-          >-> respond(text: "Not the subscription owner :(")
-      }
-
-      return conn |> middleware
+      AppEnvironment.current.stripe.fetchSubscription(conn.data.first.stripeSubscriptionId)
```
stephencelis: I made a bunch of these fetch transformers that never failover. Is that the right way to go about preparing optional data?

Sources/PointFree/Footer.swift
```diff
@@ -38,7 +38,7 @@ private let legalView = View<Prelude.Unit> { _ in
     ])
 }
 
-private let pointFreeView = View<Prelude.Unit> { _ in
+private let pointFreeView = View<Prelude.Unit> { _ -> Node in
```
mbrandonw: was this for compilation times?

I'm starting to think we should get rid of the `View.init` overload that returns a node instead of an array. it isn't the worst to always wrap in an array

Sources/PointFree/PageLayout.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+let flashView = View<Flash> { flash in
+  gridRow([`class`([flashClass(for: flash.priority)])], [
+    gridColumn(sizes: [.mobile: 12], [text(flash.message)])
+    ])
+}
+
+private func flashClass(for priority: Flash.Priority) -> CssSelector {
```
mbrandonw: ok this is cool!

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -203,17 +204,52 @@ extension Stripe.Subscription.Item {
...
 
-  return URLRequest(url: URL(string: urlString)!)
+  return request
     |> \.allHTTPHeaderFields .~ [
       "Cookie": "pf_session=\(sessionCookie)",
       "Authorization": "Basic " + Data("hello:world".utf8).base64EncodedString()
   ]
 }
+
+// TODO: expose methods from http-pipeline
```
mbrandonw: ha I just added these to my branch too. we should go ahead and make em public

Sources/PointFree/Account/Account.swift
stephencelis: The latter was the thing I tried first but it still felt heavier than this layout-specific solution. Wanna sketch something out?

Sources/PointFree/Footer.swift
stephencelis: Had the same thought!

Sources/PointFree/Account/Account.swift
stephencelis: (Layout data via contra map still needs to pluck out current user and pass along the view data intact, so I made this helper contain that logic and forward the one shared but instead: current user.)

stephencelis:
Merging for now. Let's explore more composable layout calling in the future!

PR #83, opened 2018-01-02
Redirect /pricing to /account with flash when a subscriber.

mbrandonw:
Redirects the `/pricing` page to `/account` page with a warning flash when the current user is already a subscriber.


Sources/PointFree/Pricing.swift
```diff
@@ -109,10 +110,11 @@ public enum Pricing: Codable, DerivePartialIsos {
...
 
 let pricingResponse =
-  writeStatus(.ok)
-    >-> respond(pricingView)
+  redirectCurrentSubscribers
+    <| writeStatus(.ok)
+    >-> respond(pricingView.contramap(lower))
 
-private let pricingView = View<(Pricing, Database.User?, Route)> { pricing, user, currentRoute in
+private let pricingView = View<(Database.User?, Pricing, Route)> { user, pricing, currentRoute in
```
mbrandonw: re-arranged order of tuple so that the middleware transformer could operator on `T2<User?, A>`, but could have just as easily made the transformer operate on `T3<A, User?, B>`...

Tests/PointFreeTests/PricingTests.swift
```diff
@@ -51,22 +52,31 @@ class PricingTests: TestCase {
     #endif
   }
 
-  func testPricingLoggedIn() {
-    let conn = connection(from: authedRequest(to: .pricing(nil, nil)))
-    let result = conn |> siteMiddleware
+  func testPricingLoggedIn_NonSubscriber() {
+    AppEnvironment.with(\.stripe.fetchSubscription .~ const(throwE(unit))) {
```
mbrandonw: changed this test to explicitly state that a user is logged in and they are not a subscriber.

Tests/PointFreeTests/__Snapshots__/PricingTests/testPricingLoggedIn_Subscriber.1.Conn.txt
```diff
@@ -0,0 +1,14 @@
...
+▿ Request
+  GET http://localhost:8080/pricing
+  Authorization: Basic aGVsbG86d29ybGQ=
+  Cookie: pf_session=be847b5f12f9571a96252587105c91f788704838959a4818df9555a7dbb636e09146d686873dfde259849eb38151c9f4d23819086d02dc0e7bf948441b5e73380dfda4c6e77aaa52c7d9e5d23eef28b3c4520f773d736dddff2d2f6f02048e8629f7faf219707bc7478183b3b400e3ec1373f51f8ce623ceb17e8056b22cbd49
+
+  (Data, 0 bytes)
+
+▿ Response
+  Status 302 FOUND
+  Location: /account
```
mbrandonw: snapshot to show that `/pricing` redirects to `/account`!

mbrandonw: can't wait till we get the session stuff unencrypted in tests

Sources/PointFree/Pricing.swift
stephencelis: Yeah, I've found keeping `User` first helps make user-based middleware more reusable. I wonder though if using `over{n}` means it doesn't matter.

mbrandonw: that's true, we should be able to write transformers on `T2<A, B>`, and just use some kind of `over` or some new combinator to lift existing middleware into that world.

Tests/PointFreeTests/__Snapshots__/PricingTests/testPricingLoggedIn_Subscriber.1.Conn.txt
stephencelis: Cool!

PR #88, opened 2018-01-05
Misc

stephencelis:
Gonna accumulate some small things here.


Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -27,17 +27,23 @@ extension ProfileData: Decodable {
...
+}
+
 let updateProfileMiddleware =
-  filterMap(require1 >>> pure, or: redirect(to: .account(.index)))
-    <<< filterMap(require2 >>> pure, or: loginAndRedirect)
+  filterMap(require2 >>> pure, or: loginAndRedirect)
+    <<< filterMap(require1 >>> pure, or: redirect(to: .account(.index)))
+    <<< filter(
+      get1 >>> ^\.email >>> isValidEmail,
+      or: redirect(to: .account(.index), headersMiddleware: flash(.error, "Please enter a valid email."))
```
mbrandonw: nice nice

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -217,27 +217,15 @@ extension Stripe.Subscription.Item {
...
-    else { return request }
-
-  request.allHTTPHeaderFields = (request.allHTTPHeaderFields ?? [:])
-    .merging(["Cookie": "pf_session=\(sessionCookie)"], uniquingKeysWith: { $1 })
-
-  return request
-}
-
-public func unauthedRequest(to route: Route) -> URLRequest {
+public func request(to route: Route, session: Session = .loggedOut) -> URLRequest {
```
mbrandonw: much better

Tests/LinuxMain.swift
```diff
@@ -68,6 +68,21 @@ extension HtmlCssInlinerTests {
     ("testHtmlCssInliner", testHtmlCssInliner)
   ]
 }
+extension InviteTests {
```
mbrandonw: dang, wonder if we should have a test phase that runs `make sourcery` and sees if `git status` outputs anything

stephencelis: I wonder if it's possible to hook into build phases while running `swift generate-xcodeproj`.

PR #89, opened 2018-01-05
New layouts

mbrandonw:
Implementing layouts using a dedicated struct and `contramap`ing on views. Only two screenshots broke and it was only cause we weren't properly threading the current user to the footer and now we are.


Sources/PointFree/PageLayout.swift
```diff
@@ -9,49 +9,88 @@ import Prelude
 import Styleguide
 import Tuple
 
-func respond<A>(_ view: View<A>, layout: @escaping (Flash?, View<A>) -> View<A>)
+struct SimplePageLayoutData<A> {
```
mbrandonw: this is the layout struct that holds everything the layout needs to do its job. it's already pretty big and I could imagine it getting bigger!

Sources/PointFree/PageLayout.swift
```diff
@@ -9,49 +9,88 @@ import Prelude
...
+    self.title = title
+    self.useHighlightJs = useHighlightJs
+  }
+}
+
+func respond<A, B>(
+  view: View<B>,
+  layoutData: @escaping (A) -> SimplePageLayoutData<B>
+  )
   -> Middleware<HeadersOpen, ResponseEnded, A, Data> {
```
mbrandonw: this is the layout helper. the two generic parameters is what allows this middleware to take all the data needed for the layout to do its thing, while still allowing the view to only take the data it needs to do its job.

Sources/PointFree/About.swift
```diff
@@ -10,32 +10,26 @@ import Tuple
 
 let aboutResponse =
   writeStatus(.ok)
-    >-> respond(aboutView.contramap(lower))
+    >-> map(lower)
+    >>> respond(
+      view: aboutView,
+      layoutData: { currentUser in
+        SimplePageLayoutData(currentUser: currentUser, data: unit, title: "About Us")
+    }
```
mbrandonw: here's how you use it. note that the layout needs the current user, but the view only needs `Unit`

Sources/PointFree/PageLayout.swift
```diff
@@ -9,49 +9,88 @@ import Prelude
...
+  private(set) var currentUser: Database.User?
+  private(set) var extraStyles: Stylesheet
+  private(set) var data: A
+  private(set) var flash: Flash?
+  private(set) var showTopNav: Bool
+  private(set) var title: String
+  private(set) var useHighlightJs: Bool
+
+  init(
+    currentRoute: Route? = nil,
```
mbrandonw: I provide default values for the layout fields that don't always need to be specified. however, this one in particular _should_ be required, as it's makes the nav properly redirect after a login. but we weren't doing a very good job of threading that info to the nav, so rather than breaking a bunch of stuff and fixing I made it optional

Sources/PointFree/Account/Account.swift
```diff
@@ -51,10 +61,16 @@ private func fetchAccountData<I, A>(
 
 let accountView = View<(Stripe.Subscription?, [Database.TeamInvite], [Database.User], [Database.EmailSetting], Database.User)> { subscription, teamInvites, teammates, emailSettings, currentUser in
 
-  titleRowView.view(unit)
-    <> profileRowView.view((currentUser, emailSettings))
-    <> subscriptionRowView.view((subscription, teamInvites, teammates))
-    <> logoutView.view(unit)
+  gridRow([
+    gridColumn(sizes: [.mobile: 12, .desktop: 8], [style(margin(leftRight: .auto))], [
+      div([`class`([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]])])],
```
mbrandonw: I copied and pasted this type of row (mobile 12, desktop 8, auto margins, lots of padding) a few times just to get tests passing, but it could also make sense to either create a new type of layout for columnar pages (e.g. account, terms, etc..), or create an option in `SimplePageLayoutData`

Sources/PointFree/Account/Account.swift
```diff
@@ -13,7 +13,17 @@ let accountResponse =
   filterMap(require1 >>> pure, or: loginAndRedirect)
     <| fetchAccountData
     >-> writeStatus(.ok)
-    >-> respond(accountView.contramap(lower), layout: simplePageLayout(title: "Account", currentUser: get5))
+    >-> map(lower)
```
mbrandonw: now instead of `myView.contramap(lower)` you'll probably insert this transformation just before the layout `respond` middleware.

Sources/PointFree/Account/Account.swift
```diff
@@ -13,7 +13,17 @@ let accountResponse =
...
+    >>> respond(
+      view: accountView,
+      layoutData: { subscription, teamInvites, teammates, emailSettings, currentUser in
```
stephencelis: There's a lot of noise here, but maybe that's OK. Wonder about trying something shorter like:

``` swift
layoutData: { .init(currentUser: $0.4, data: $0, title: "Account" }
```

Sources/PointFree/Home.swift
```diff
@@ -11,25 +11,23 @@ import UrlFormEncoding
...
-    >-> respond(secretHomeView.map(addGoogleAnalytics))
+    >-> respond(
+      view: secretHomeView,
+      layoutData: { currentUser in
+        SimplePageLayoutData(
+          currentUser: currentUser,
+          data: currentUser,
+          extraStyles: pricingExtraStyles,
+          showTopNav: false,
+          title: "Point-Free: A weekly video series on functional programming and the Swift programming language."
```
stephencelis: Realizing there's a bunch of `meta` data we probably want to be able to customize here!

Sources/PointFree/PageLayout.swift
stephencelis: Sounds good! We can revisit fixing soon tho.

Sources/PointFree/Home.swift
mbrandonw: yup yup, we have a metadata layout we can compose with this and `contramap` on, or we could just add all those fields to this struct

PR #90, opened 2018-01-05
External episode repos

mbrandonw:
This supersedes #44. It flattens the level of extra repos by one. There are two external repos that hold episode transcripts. Both of them contain the full `Episode` struct and related models, so it's on us to keep them in sync.


Package.swift
```diff
@@ -14,6 +15,10 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .revision("c510e7d")),
     .package(url: "https://github.com/pointfreeco/swift-web.git", .revision("54412eb")),
     .package(url: "https://github.com/vapor/postgresql.git", from: "2.0.0"),
+
+    ProcessInfo.processInfo.environment["OSS"] == "1"
```
mbrandonw: we'll need to finesse the OSS experience a bit, but for now by default we'll use the private transcript repos.

Sources/PointFree/Admin/Index.swift
```diff
@@ -49,7 +50,7 @@ let showNewEpisodeEmailMiddleware =
 
 private let showNewEpisodeView = View<Database.User> { currentUser in
   ul(
-    episodes
+    AppEnvironment.current.episodes()
```
mbrandonw: I hide the episodes inside the environment so that we can control it in tests

Sources/PointFreeTestSupport/MockEpisode.swift
```diff
@@ -0,0 +1,99 @@
+@testable import EpisodeTranscripts
+
+extension Episode {
+  public static let mock = proofInFunctions
+}
+
+private let proofInFunctions = Episode(
```
mbrandonw: this is the mock episode we'll use for tests. I wanted to keep this PR small, but soon I want to update this mock episode to be a better example of all the different things that can show up in an episode.

Sources/PointFree/Admin/Index.swift
stephencelis: Any reason why it's a function and not a property?

mbrandonw: just anticipating this function taking arguments some day for filtering by tags / sorting by fields etc...

mbrandonw:
deleted `Package.resolved` and added to `.gitignore` to prevent CI from loading the wrong transcripts repo... got the green now.

gonna merge the episode PR's and then update and merge this

PR #91, opened 2018-01-06
Subscription changes

stephencelis:
A WIP, but wanted to push what I had so far before running some errands.

Adds the ability to upgrade/downgrade individual subscriptions and will allow team subscriptions to add/remove seats!


Sources/PointFree/Account/Account.swift
```diff
@@ -310,6 +304,32 @@ private func mainAction(for subscription: Stripe.Subscription) -> Node {
   }
 }
 
+private func changeAction(for subscription: Stripe.Subscription) -> [Node] {
+  guard subscription.status == .active else { return [] }
+
+  let (action, route): (String, Route)
+  switch subscription.plan.id.unwrap {
+  case Stripe.Plan.Id.individualMonthly.unwrap:
```
stephencelis: This'll get much better with conditional conformance!

Sources/PointFree/Account/Cancel.swift
```diff
@@ -59,17 +60,31 @@ private func cancel(_ conn: Conn<StatusLineOpen, Tuple2<Stripe.Subscription, Dat
...
 
 // MARK: - Transformers
 
+func requireSubscriptionItem<A>(
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, T3<Stripe.Subscription.Item, Stripe.Subscription, A>, Data>
+  )
+  -> Middleware<StatusLineOpen, ResponseEnded, T2<Stripe.Subscription, A>, Data> {
+
+    return filterMap(
+      { data in pure(data.first.items.data.first.map { $0 .*. data }) },
```
stephencelis: A tuple `prepend` filter map ends up looking like this.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -53,10 +53,32 @@ private func render(conn: Conn<StatusLineOpen, T2<Database.User?, Route>>)
       return conn.map(const(user .*. unit))
         |> cancelMiddleware
 
+    case .account(.subscription(.changeSeats(.show))):
+      fatalError()
```
stephencelis: Adding/removing seats is still TODO.

Sources/PointFree/Stripe.swift
```diff
@@ -244,17 +246,6 @@ private func fetchSubscription(id: Stripe.Subscription.Id) -> EitherIO<Prelude.U
   return stripeDataTask("subscriptions/" + id.unwrap + "?expand[]=customer")
 }
 
-private func reactivateSubscription(_ subscription: Stripe.Subscription)
-  -> EitherIO<Prelude.Unit, Stripe.Subscription> {
-
-    guard
-      subscription.cancelAtPeriodEnd,
-      let item = subscription.items.data.first
-      else { return throwE(unit) }
```
stephencelis: This `guard` is already fulfilled with a filterMap, so just using `updateSubscription` from the reactivation middleware.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -12,9 +12,10 @@ import Tuple
 // MARK: Middleware
 
 let confirmCancelResponse =
-  requireStripeSubscription
+  filterMap(require1 >>> pure, or: loginAndRedirect)
```
stephencelis: Requiring user is now explicit.

Sources/PointFree/Account/ChangeSeats.swift
```diff
@@ -0,0 +1,190 @@
...
+
+private func fetchSeatsTaken<A>(
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, T3<Database.User, Int, A>, Data>
+  )
+  -> Middleware<StatusLineOpen, ResponseEnded, T2<Database.User, A>, Data> {
+
+    return { conn -> IO<Conn<ResponseEnded, Data>> in
+      let user = conn.data.first
+
+      let invitesAndTeammates = sequence([
```
stephencelis: More of this stuff could be parallelized and cleaned up in the future. If we add a proper query builder we can just `SELECT count(*)` here instead.

Sources/PointFree/Account/ChangeSeats.swift
```diff
@@ -0,0 +1,191 @@
...
+
+private func changeSeats(_ conn: Conn<StatusLineOpen, Tuple4<Stripe.Subscription, Database.User, Int, Int>>)
+  -> IO<Conn<StatusLineOpen, Prelude.Unit>> {
+
+    let (subscription, _, seatsTaken, quantity) = lower(conn.data)
+
+    // TODO: send emails
+    return  AppEnvironment.current.stripe.updateSubscription(subscription, .teamYearly, quantity)
+      .flatMap { sub -> EitherIO<Prelude.Unit, Stripe.Subscription> in
+        if sub.quantity > subscription.quantity {
```
stephencelis: We immediately invoice when seats are added.

PR #92, opened 2018-01-06
Store subscription status in DB

mbrandonw:
### What

Adds a `stripe_subscription_status` column to the `subscriptions` table so that we can more easily show/hide stuff based on it. Currently we only update it on login, but we will update it more often once we get stripe hooks in place.


Sources/PointFree/SiteMiddleware.swift
```diff
@@ -18,12 +19,15 @@ public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Uni
...
     <| currentUserMiddleware
+    >-> currentSubscriptionMiddleware
     >-> render(conn:)
 
-private func render(conn: Conn<StatusLineOpen, T2<Database.User?, Route>>)
+private func render(conn: Conn<StatusLineOpen, T3<Database.Subscription?, Database.User?, Route>>)
   -> IO<Conn<ResponseEnded, Data>> {
 
-    let (user, route) = (conn.data.first, conn.data.second)
+    let (subscription, user, route) = (conn.data.first, conn.data.second.first, conn.data.second.second)
```
mbrandonw: really starting to feel we should bring `RequestContext` back! These 3 values should prob go into every page's middleware...

stephencelis: It's possible! Thought here they can be optional and we `require` them a lot of places in middleware, so not sure how we resolve both worlds yet.

mbrandonw:
Oh right. Maybe then just a typealias of a tuple so we can still manipulate
it.

Starting to think we need a tuple union function that combines tuples but
keeps em flat

PR #93, opened 2018-01-06
Better partial iso generator

mbrandonw:
Some of our nested routes were getting pretty gnarly, so by extension `PartialIso` directly we can leverage type inference!


.sourcery-templates/DerivePartialIsos.stencil
```diff
@@ -0,0 +1,41 @@
...
+
+{% for enum in types.implementing.DerivePartialIsos|enum %}
+  {% for case in enum.cases %}
+
+    {% if case.hasAssociatedValue %}
+
+      extension PartialIso where A == (
+        {% for value in case.associatedValues %}
+          {% if value.type %}
+            {{ value.type.globalName }}{% if value.typeName.isOptional %}?{% endif %}
```
mbrandonw: there's a lot of hacky stuff happening in this stencil cause apparently sourcery and/or stencil doesnt have great nested type support. see more here https://github.com/krzysztofzablocki/Sourcery/issues/398

Sources/PointFree/Routers/GitHubRoutes.swift
```diff
@@ -19,25 +19,25 @@ public enum GitHubRoute: DerivePartialIsos {
 
 let gitHubRouter = [
 
-  GitHubRoute.iso.authorize
+  PartialIso.authorize
```
mbrandonw: for some strange reason I could not omit this `PartialIso` ... it's the only one that needed it!

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -92,135 +92,127 @@ public enum Route: DerivePartialIsos {
...
+  .account <<< .subscription <<< .downgrade <<< .update
     <¢> post %> lit("account") %> lit("subscription") %> lit("downgrade") <% end,
 
-  Route.iso.account <<< Route.Account.iso.subscription <<< Route.Account.Subscription.iso.reactivate
+  .account <<< .subscription <<< .reactivate
     <¢> post %> lit("account") %> lit("subscription") %> lit("reactivate") <% end,
 
-  Route.iso.account <<< Route.Account.iso.subscription <<< Route.Account.Subscription.iso.upgrade
-    <<< Route.Account.Subscription.Upgrade.iso.show
+  .account <<< .subscription <<< .upgrade <<< .show
```
mbrandonw: much better

Tests/LinuxMain.swift
```diff
@@ -29,14 +29,37 @@ extension AuthTests {
 extension CancelTests {
   static var allTests: [(String, (CancelTests) -> () throws -> Void)] = [
     ("testConfirmCancel", testConfirmCancel),
-    ("testCancel", testCancel)
+    ("testConfirmCancelLoggedOut", testConfirmCancelLoggedOut),
```
mbrandonw: all your tests weren't running on linux!

Sources/PointFree/Routers/GitHubRoutes.swift
stephencelis: You need one to hint the compiler! I think you could hint by giving `gitHubRouter` an explicit type, too.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -92,135 +92,127 @@ public enum Route: DerivePartialIsos {
 
 private let routers: [Router<Route>] = [
 
-  Route.iso.about
+  .about
     <¢> get %> lit("about") <% end,
 
-  Route.iso.account <<< Route.Account.iso.confirmEmailChange
+  .account <<< .confirmEmailChange
```
stephencelis: Beaut!

Tests/LinuxMain.swift
stephencelis: 🙈

Sources/PointFree/Routers/GitHubRoutes.swift
mbrandonw: but somehow it works for `TwitterRoute`!

PR #95, opened 2018-01-06
Don't allow /login when already logged in

mbrandonw:
eom


Sources/PointFree/Auth.swift
```diff
@@ -49,6 +51,20 @@ public func currentUserMiddleware<A>(_ conn: Conn<StatusLineOpen, A>)
     return user.map { conn.map(const($0 .*. conn.data)) }
 }
 
+public func requireLoggedOutUser<A>(
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, A, Data>
+  ) -> Middleware<StatusLineOpen, ResponseEnded, T2<Database.User?, A>, Data> {
```
mbrandonw: I like how the type signature encodes how we are dropping `Database.User?` from the tuple to force a logged out user

Tests/PointFreeTests/__Snapshots__/AuthTests/testLogin_AlreadyLoggedIn.1.Conn.txt
```diff
@@ -0,0 +1,14 @@
...
+  GET http://localhost:8080/login
+  Authorization: Basic aGVsbG86d29ybGQ=
+  Cookie: pf_session={"userId":"00000000-0000-0000-0000-000000000000"}
+
+  (Data, 0 bytes)
+
+▿ Response
+  Status 302 FOUND
+  Location: /account
+  Set-Cookie: pf_session={"flash":{"message":"You’re already logged in.","priority":"warning"},"userId":"00000000-0000-0000-0000-000000000000"}; Path=/
```
mbrandonw: love snapshot tests on flash messages

PR #97, opened 2018-01-07
Newest homepage nav

mbrandonw:
Implemented @amotion's newest nav:

testhomepage 2 desktop copy


Sources/PointFree/Home.swift
```diff
@@ -21,25 +21,24 @@ let secretHomeMiddleware: (Conn<StatusLineOpen, Tuple3<Database.User?, Stripe.Su
           currentSubscriptionStatus: currentSubscriptionStatus,
           currentUser: currentUser,
           data: (currentUser, currentSubscriptionStatus),
-          extraStyles: pricingExtraStyles,
+          extraStyles: pricingExtraStyles <> blueGradientStyles <> reflectStyles,
```
mbrandonw: these extra styles could prob be moved to our design systems lib, but just leaving in this file for now

Sources/PointFree/Home.swift
```diff
@@ -53,16 +52,17 @@ let headerView = View<Prelude.Unit> { _ in
...
+          headerLinks.view((currentUser, currentSubscriptionStatus, currentRoute))
+        )
         ])
       ]),
 
     gridRow([`class`([Class.grid.top(.mobile), Class.grid.between(.mobile), Class.padding([.mobile: [.top: 3], .desktop: [.top: 0]])])], [
 
       gridColumn(sizes: [.mobile: 5], [`class`([Class.padding([.mobile: [.top: 4], .desktop: [.top: 0]])]), style(lineHeight(0))], [
-        img(base64: heroLeftMountainSvgBase64, mediaType: .image(.svg), alt: "", [width(.pct(100))])
+        img(base64: heroMountainSvgBase64, mediaType: .image(.svg), alt: "", [width(.pct(100))])
```
mbrandonw: instead of a left and right mountain svg, we just have a single mountain and the right one is reflected across the x-axis

Sources/PointFree/Home.swift
```diff
@@ -124,3 +143,25 @@ private let episodeInfoColumnView = View<Episode> { ep in
     ]
   )
 }
+
+private let blueGradientClass = CssSelector.class("blue-gradient")
+private let blueGradientStyles =
+  blueGradientClass % (
+    key("background", "rgba(128,219,255,1)")
+      <> key("background", "-moz-linear-gradient(top, rgba(128,219,255,1) 0%, rgba(128,219,255,0) 100%)")
```
mbrandonw: we dont have first class support for gradients or transforms in our css lib, so just doing raw css for now

Sources/Styleguide/DesignSystems/Typography.swift
```diff
@@ -30,6 +30,9 @@ extension Class {
     public static let fontSizeInherit = CssSelector.class("font-size-inherit")
     public static let textDecorationNone = CssSelector.class("text-decoration-none")
 
+    public static let breakWord = CssSelector.class("break-word")
+    public static let nowrap = CssSelector.class("nowrap")
```
mbrandonw: some new atomic styles. realized we should apply `nowrap` to our buttons so that they dont break when in constrained areas

mbrandonw:
mobile screenshots dont look great, but we just need a lot of work on mobile in general

PR #99, opened 2018-01-07
Privacy Policy

stephencelis:
image


Sources/PointFree/Privacy.swift
```diff
@@ -0,0 +1,174 @@
...
+    <> howWeUseCollectedInformation
+    <> howWeProtectYourInformation
+    <> sharingYourPersonalInformation
+    <> complianceWithChildrensOnlinePrivacyProtectionAct
+    <> changesToThisPrivacyPolicy
+    <> contactingUs
+
+private let personalIdentificationInformation = [
+  h2(
+    [`class`([Class.h3, Class.padding([.mobile: [.top: 2]])])],
```
mbrandonw: we should use `Class.pf.type.title3` instead of `Class.h3` cause it sets color and line-heights in addition to font size

Sources/PointFree/Privacy.swift
```diff
@@ -0,0 +1,174 @@
...
+      view: privacyView,
+      layoutData: { currentUser in
+        SimplePageLayoutData(currentUser: currentUser, data: unit, title: title)
+    }
+)
+
+private let privacyView = View<Prelude.Unit> { _ in
+  gridRow([
+    gridColumn(sizes: [.mobile: 12], [
+      div(
```
mbrandonw: we should treat this page like the account page by giving it `auto` left/right margins to make it a more columnar style page. really we should make a layout to support this or put support into the existing layout. 

this is the grid that does it:

```swift
  gridRow([
    gridColumn(sizes: [.mobile: 12, .desktop: 8], [style(margin(leftRight: .auto))], [
```

PR #102, opened 2018-01-10
Use meta layout with simple layout

stephencelis:
Exposes some extra attributes on our simple layout and feeds them to our meta layout helper!


Sources/PointFree/PageLayout.swift
```diff
@@ -51,36 +63,39 @@ func respond<A, B>(
...
-    return document([
+    document([
       html([
         head([
           title(layoutData.title),
           style(renderedNormalizeCss),
           style(styleguide),
           style(layoutData.extraStyles),
           meta(viewport: .width(.deviceWidth), .initialScale(1)),
-          script(
```
stephencelis: Get this for free with `addGoogleAnalytics` :)

mbrandonw: oh ha, I was actually going to deprecate `addGoogleAnalytics` in favor of just putting the script in directly.

stephencelis: lol. either way! are you into reusing the meta stuff or should that be direct?

stephencelis: Tbh I thought we could add the site `id` as a configuration option to `addGoogleAnalytics` and it's a nice example of an HTML `Node` transformer.

mbrandonw: I think that could go direct too since we're putting the fields into the layout data.

someday we should sit down and figure out how/if layouts compose, but maybe for now it's just easily to be very direct.

stephencelis: I actually started going with the direct route (see remnants from the second commit in this PR), but ended up reusing the meta stuff just because it illustrates a nice boundary between what could live in swift-web and what can't. Some of that layer is a direct mapping, but others we stub out.

mbrandonw: oh cool cool, in that case let's leave as is!

PR #103, opened 2018-01-10
Atom feed

stephencelis:
Preliminary work for Atom/RSS feeds.


stephencelis:
(Still a WIP, but getting there.)

Sources/PointFree/AtomFeed.swift
```diff
@@ -0,0 +1,40 @@
...
+let pointFreeFeed = View<[Episode]> { episodes in
+  atomLayout.view(
+    AtomFeed(
+      author: AtomAuthor(
+        email: "support@pointfree.co",
+        name: "Point-Free"
+      ),
+      entries: episodes.map(atomEntry(for:)),
+      atomUrl: url(to: .feed(.atom)),
+      siteUrl: "/", // FIXME url(to: .root),
```
mbrandonw: can't we make this `.secretHome` or `.home(signedUpSuccessfully: nil)` right now?

Sources/PointFree/AtomFeed.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+import Html
+import HttpPipeline
+import MediaType
+import Prelude
+
+let atomFeedResponse =
+  writeStatus(.ok)
+    >-> respond(pointFreeFeed, contentType: .html)
```
mbrandonw: is the content type supposed to be xml?

Sources/PointFree/AtomFeed.swift
```diff
@@ -0,0 +1,40 @@
...
+    conn |> respond(body: view.rendered(with: conn.data), contentType: contentType)
+  }
+}
+
+private func atomEntry(for episode: Episode) -> AtomEntry {
+  return AtomEntry(
+    title: episode.title,
+    siteUrl: url(to: .episode(.left(episode.slug))),
+    updated: episode.publishedAt,
+    content: []
```
mbrandonw: should we use the `episode.blurb` here? or we might even be able to reuse the exact view from the "new episode" email. are atom entries allowed to have inline css styles?

Sources/PointFree/Atom.swift
```diff
@@ -0,0 +1,114 @@
...
+
+public func content(_ attribs: [Attribute<Element.Content>], _ content: [Node]) -> Node {
+  return node("content", attribs, [.text(unsafeUnencodedString("<![CDATA[" + render(content).string + "]]>"))])
+}
+
+public func type(_ type: String) -> Attribute<Element.Content> {
+  return attribute("type", type)
+}
+
+private let atomDateFormatter = DateFormatter() // FIXME
```
mbrandonw: what needs to be fixed here?

Sources/PointFree/AtomFeed.swift
stephencelis: It is! Or `application/atom+xml`/`application(.atom(.xml))` Needs to be fixed.

Sources/PointFree/AtomFeed.swift
stephencelis: Yep yep.

Sources/PointFree/Atom.swift
stephencelis: Needs to be configured for Atom's formatting.

Sources/PointFree/AtomFeed.swift
stephencelis: I don't think they'd render. Atom entries should just be basic DOM HTML.

Sources/PointFree/AtomFeed.swift
mbrandonw: `MediaType` should prob be switched to a a struct huh? or at least have an `other` case?

PR #106, opened 2018-01-10
Send email confirmation for cancel/reactivate

stephencelis:
Knocking out a coupla these.


Sources/PointFree/Account/Account.swift
```diff
@@ -182,8 +182,9 @@ private func planName(for subscription: Stripe.Subscription) -> String {
 public func status(for subscription: Stripe.Subscription) -> String {
   switch subscription.status {
   case .active:
-    let currentPeriodEndString = subscription.currentPeriodEnd
-      .map { subscription.cancelAtPeriodEnd ? " through " + dateFormatter.string(from: $0) : "" } ?? ""
+    let currentPeriodEndString = subscription.cancelAtPeriodEnd
```
stephencelis: Looks like `current_period_end` always has a value from Stripe.

Sources/PointFree/Account/Account.swift
```diff
@@ -480,10 +479,11 @@ private let subscriptionPaymentInfoView = View<Stripe.Subscription> { subscripti
...
 }
 
-private func totalAmount(for subscription: Stripe.Subscription) -> String? {
+private func totalAmount(for subscription: Stripe.Subscription) -> String {
   let totalCents = subscription.plan.amount.rawValue * subscription.quantity
-  let totalDollars = Double(totalCents) / 100
-  return currencyFormatter.string(from: NSNumber(value: totalDollars))
+  let totalDollars = NSNumber(value: Double(totalCents) / 100)
+  return currencyFormatter.string(from: totalDollars)
+    ?? NumberFormatter.localizedString(from: totalDollars, number: .currency)
```
stephencelis: `NumberFormatter.string(from:)` returns an optional, but `localizedString(from:)` does not 😕

Sources/PointFree/Account/Cancel.swift
```diff
@@ -73,23 +75,23 @@ private func cancel(_ conn: Conn<StatusLineOpen, Tuple2<Stripe.Subscription, Dat
...
             )
-          ),
-          const(
-            conn |> redirect(
-              to: .account(.index),
-              headersMiddleware: flash(.notice, "We’ve canceled your subscription.")
-            )
           )
-        )
+        ) { _ in
```
mbrandonw: whoa my brain is having a hard time parsing this. is this a trailing closure for a curried `either` or something?

stephencelis: Ah! It's `either(L -> A, B -> A)`. The left path is a const so I pass it through. I'm using a trailing closure for the "right" path.

PR #108, opened 2018-01-12
Bootstrap function for starting up the site.

mbrandonw:
This started from me investigating what we'd need to do to support a `List-Unsubscribe` header in our emails. One component of it is for us to main a "mailgun route" for each of our newsletter types so that we can forward a reply to "unsubscribe-newsletter@pointfree.co" to our server, and then we handle the response.

Then that led me to wanting to synchronize those routes on mailgun via their API with the cases in our enum.

Then that led me to wanting to have a bootstrap function that runs when the server first boots up.

But we already have something like that in `pointfreeco-server` for starting up postgres.

So, I just made a general bootstrap function that can be called from the server. This is the place I would add that route synchronization stuff once I get to it.

Whatdya think?

### DIFF
--- a/Sources/PointFree/TODO.swift
+++ b/Sources/PointFree/TODO.swift
@@ -392,13 +392,16 @@ extension EitherIO {
   }
 
   func retry(maxRetries: Int, backoff: @escaping (Int) -> DispatchTimeInterval) -> EitherIO {
-    return self.retry(maxRetries: maxRetries, attempts: 0, backoff: backoff)
+    return self.retry(maxRetries: maxRetries, attempts: 1, backoff: backoff)
   }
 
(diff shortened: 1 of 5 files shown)


Sources/PointFree/TODO.swift
```diff
@@ -392,13 +392,16 @@ extension EitherIO {
...
+    return self.retry(maxRetries: maxRetries, attempts: 1, backoff: backoff)
   }
 
   private func retry(maxRetries: Int, attempts: Int, backoff: @escaping (Int) -> DispatchTimeInterval) -> EitherIO {
 
+    guard attempts < maxRetries else { return self }
+
     return self <|> .init(run:
       self
+        .retry(maxRetries: maxRetries, attempts: attempts + 1, backoff: backoff)
```
mbrandonw: wow I totally did not finish implementing this `retry` helper when I started it! it now has a proper implementation and some tests

PR #110, opened 2018-01-13
Add preliminary seat change email

stephencelis:
Needs polish but the logic's here!


Sources/Styleguide/DesignSystems/Layout.swift
```diff
@@ -44,4 +44,4 @@ private let floatStyles =
     <> Class.layout.right % float(.right)
 
 private let widthStyles =
-  Class.layout.fit % maxWidth(.pct(100))
+  Class.layout.fit % width(.pct(100))
```
stephencelis: This pushes the video tag to be full-width before loading. Less glitchy.

PR #112, opened 2018-01-14
Update postgres package to point to community package.

mbrandonw:
It seems that https://github.com/vapor/postgresql no longer exists, and instead we should point to https://github.com/vapor-community/postgresql ¯\_(ツ)_/¯


Package.swift
```diff
@@ -15,7 +15,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .revision("c510e7d")),
     .package(url: "https://github.com/pointfreeco/swift-web.git", .revision("35421ef")),
     .package(url: "https://github.com/pointfreeco/Ccmark.git", .branch("master")),
-    .package(url: "https://github.com/vapor/postgresql.git", from: "2.0.0"),
+    .package(url: "https://github.com/vapor-community/postgresql.git", .exact(.init(2, 1, 1))),
```
mbrandonw: oh also upgrading us to 2.1.1. everything still works and it seems they fixed a memory leak in newer releases.

PR #114, opened 2018-01-15
Use gradient instead of image online

mbrandonw:
A quick (kinda hacky) fix...


Sources/PointFree/Home.swift
```diff
@@ -44,10 +44,10 @@ private let episodeRowView = View<Episode> { ep in
       gridColumn(sizes: [.mobile: 12, .desktop: 7], episodeInfoColumnView.view(ep)),
 
       gridColumn(sizes: [.mobile: 12, .desktop: 5], [`class`([Class.grid.first(.mobile), Class.grid.last(.desktop)])], [
-        div([`class`([Class.size.height100pct]), style(lineHeight(0))], [
+        div([`class`([Class.size.height100pct]), style(lineHeight(0) <> gradient <> minHeight(.px(300)))], [
           a([href(path(to: .episode(.left(ep.slug))))], [
             img(
-              src: "https://waltpaper.com/wp-content/uploads/2017/06/6c7b04501f6e98ab94982a2743ede747-1024x768.jpeg",
```
mbrandonw: don't use this online image

Sources/PointFree/Home.swift
```diff
@@ -44,10 +44,10 @@ private let episodeRowView = View<Episode> { ep in
       gridColumn(sizes: [.mobile: 12, .desktop: 7], episodeInfoColumnView.view(ep)),
 
       gridColumn(sizes: [.mobile: 12, .desktop: 5], [`class`([Class.grid.first(.mobile), Class.grid.last(.desktop)])], [
-        div([`class`([Class.size.height100pct]), style(lineHeight(0))], [
+        div([`class`([Class.size.height100pct]), style(lineHeight(0) <> gradient <> minHeight(.px(300)))], [
```
mbrandonw: have to do the `min-height` cause without an image the div collapses in mobile with no image.

stephencelis: Wonder if we can set an aspect ratio to prevent glitches in general.

mbrandonw: I tried a few approaches towards that, none of em worked great

PR #115, opened 2018-01-19
Fix javascript number formatting.

mbrandonw:


mbrandonw:
just pushed some code to validate subscription quantity, and that led me down a lil refactor path. will make some notes...

Makefile
```diff
@@ -16,6 +16,8 @@ init-db:
 deinit-db:
 	psql template1 < database/deinit.sql
 
+reset-db: deinit-db init-db
```
mbrandonw: this can be useful

Sources/PointFree/Pricing.swift
```diff
@@ -523,27 +524,32 @@ private func tabStyles(
         <> selectedStyles
 }
 
-private func redirectCurrentSubscribers<A>(
```
mbrandonw: made this reusable, new version down below...

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -22,7 +22,7 @@ public enum Route: DerivePartialIsos {
   case launchSignup(EmailAddress)
   case login(redirect: String?)
   case logout
-  case pricing(String?, Int?)
+  case pricing(Pricing?)
```
mbrandonw: we can now use a first class `Pricing` value

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -210,7 +210,10 @@ private let routers: [Router<Route>] = [
     <¢> get %> lit("logout") <% end,
 
   .pricing
-    <¢> get %> lit("pricing") %> queryParam("plan", opt(.string)) <%> queryParam("quantity", opt(.int)) <% end,
+    <¢> get %> lit("pricing")
+    %> (queryParam("plan", opt(.string)) <%> queryParam("quantity", opt(.int)))
+      .map(PartialIso.pricing >>> Optional.iso.some)
```
mbrandonw: here `PartialIso.pricing` is an iso `(String?, Int?) -> Pricing`, and by mapping with `PartialIso.pricing >>> Optional.iso.some` we guarantee that this succeeds in parsing even if parsing the pricing value fails.

`Optional.iso.some` is a really bad name, it should be moved to an extension on `PartialIso where B == A?`, but we can clean up later.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -277,3 +282,24 @@ private func verify(payload: MailgunForwardPayload) -> Bool {
...
+  fileprivate static var pricing: PartialIso {
+    return PartialIso(
+      apply: { plan, quantity in
+        let pricing: Pricing
+        if let quantity = quantity {
+          pricing = .team(quantity)
+        } else if let plan = plan, let billing = Pricing.Billing(rawValue: plan) {
+          pricing = .individual(billing)
+        } else {
+          pricing = .default
```
mbrandonw: I moved this logic from the site middleware to here.

now, I would have expected that this `.default` would not be necessary but it is :/ I need to dig into this more, but going to leave for now

Tests/PointFreeTests/SubscribeTests.swift
```diff
@@ -0,0 +1,158 @@
...
+import HtmlPrettyPrint
+import HttpPipeline
+@testable import PointFree
+import PointFreeTestSupport
+import Prelude
+import Optics
+import SnapshotTesting
+import XCTest
+
+final class SubscribeTests: TestCase {
```
mbrandonw: I dont think we are testing the subscribe endpoint at all right now, so I added a bunch of tests

Sources/PointFree/Subscribe.swift
```diff
@@ -11,42 +11,70 @@ public struct SubscribeData: Codable {
 }
 
 let subscribeMiddleware =
-  filterMap(require2 >>> pure, or: loginAndRedirect)
+  filterMap(
+    require1 >>> pure,
+    or: redirect(to: .pricing(nil), headersMiddleware: flash(.error, "Error creating subscription!"))
```
mbrandonw: this filter map makes sure there is `SubscribeData` present...

Sources/PointFree/Subscribe.swift
```diff
@@ -11,42 +11,70 @@ public struct SubscribeData: Codable {
...
+    )
+    <<< filter(
+      get1 >>> ^\.pricing >>> validateQuantity,
+      or: redirect(to: .pricing(nil), headersMiddleware: flash(.error, "An invalid subscription quantity was used."))
```
mbrandonw: then this filter map makes sure the quantity is correct

Sources/PointFree/Subscribe.swift
```diff
@@ -11,42 +11,70 @@ public struct SubscribeData: Codable {
...
+    )
+    <<< redirectActiveSubscribers(user: get2)
```
mbrandonw: this we redirect active subscribers away

Sources/PointFree/Subscribe.swift
```diff
@@ -11,42 +11,70 @@ public struct SubscribeData: Codable {
...
+    <<< filterMap(require2 >>> pure, or: loginAndRedirectToPricing)
```
mbrandonw: and finally we require the user to be present

Tests/PointFreeTests/SubscribeTests.swift
```diff
@@ -0,0 +1,178 @@
...
+    super.tearDown()
+    AppEnvironment.pop()
+  }
+
+  func testNotLoggedIn_IndividualMonthly() {
+    let conn = connection(from: request(to: .subscribe(.some(.individualMonthly))))
+      |> siteMiddleware
+      |> Prelude.perform
+
+    #if !os(Linux)
```
mbrandonw: had to disable linux tests cause snapshots of dictionaries getting reordered :/

```diff
−  pricing[individual]=monthly&token=stripe-deadbeef
+  token=stripe-deadbeef&pricing[individual]=monthly

```

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -157,17 +157,8 @@ private func render(conn: Conn<StatusLineOpen, T3<Database.Subscription?, Databa
...
-        pricing = .team(quantity)
-      } else if let plan = plan, let billing = Pricing.Billing(rawValue: plan) {
-        pricing = .individual(billing)
-      } else {
-        pricing = .default
-      }
-
-      return conn.map(const(user .*. pricing .*. route .*. unit))
+    case let .pricing(pricing):
+      return conn.map(const(user .*. (pricing ?? .default) .*. route .*. unit))
```
stephencelis: Nice cleanup!

PR #116, opened 2018-01-20
Retry stripe POSTs

mbrandonw:
Well, this was way easier than expected!

Stripe supports an "idempotent key" (https://stripe.com/docs/api?lang=curl#idempotent_requests) on `POST`s so that if the network is flaky (or stripe/our servers had an intermittent hiccup) we can retry the request without fear of accidentally perform the action twice (like charging someone!).

I've implemented this by attaching a `UUID` to every `POST` stripe call, and then making every `POST` task retry 10 times. 

We could even take this a step further by putting the user into an "authorizing" state (with appropriate global banner on all pages) and kicking off all the work into the background and allow more time between retries. Maybe someday...


Tests/PointFreeTests/EitherIOTests.swift
```diff
@@ -30,4 +30,15 @@ class EitherIOTests: TestCase {
 
     XCTAssertTrue(result.isRight)
   }
+
+  func testRetry_MaxRetriesZero() {
+    let thing = EitherIO<Prelude.Unit, Prelude.Unit>(run: IO {
+      return .right(unit)
+    })
+      .retry(maxRetries: 0)
```
mbrandonw: I just wanted to make sure the degenerate case of `maxRetries: 0` works as expected.

Sources/PointFree/Stripe.swift
```diff
@@ -319,10 +325,14 @@ private func stripeDataTask<A>(_ path: String, _ method: Method = .get)
...
 
-    return pure(stripeUrlRequest(path, method))
+    let task: EitherIO<Prelude.Unit, A> = pure(stripeUrlRequest(path, method))
       .flatMap { jsonDataTask(with: auth <| $0, decoder: stripeJsonDecoder) }
       .map(tap(AppEnvironment.current.logger.debug))
       .withExcept(tap(AppEnvironment.current.logger.error) >>> const(unit))
+
+    return method.isPost
+      ? task.retry(maxRetries: 10)
+      : task
```
mbrandonw: I could also do `.retry(maxRetries: method.isPost ? 10 : 0)` but I also like this

stephencelis: Do we wanna do the same for `DELETE`s? Maybe the logic should be:

``` swift
return method.isGet
  ? task
  : task.retry(maxRetries: 10)
 ```

mbrandonw: `DELETE`s should be idempotent to begin with, so those should be ok. stripe's docs specifically call out POST's as the main culprit. 

however, perhaps better is to just switch here so that we have to consider this again when adding a new verb.

PR #117, opened 2018-01-21
Global not found page

mbrandonw:
Adding dedicated 404 page for when a route is not recognized.


Tests/PointFreeTests/NotFoundMiddlewareTests.swift
```diff
@@ -0,0 +1,44 @@
...
+    super.tearDown()
+    AppEnvironment.pop()
+  }
+
+  func testNotFound() {
+    let result = connection(from: URLRequest(url: URL(string: "http://localhost:8080/404")!))
+      |> siteMiddleware
+      |> Prelude.perform
+
+    assertSnapshot(matching: result)
```
stephencelis: Just wondering if we want to test logged-in/subscribed states or if that's overkill.

mbrandonw: yeah might as well!

PR #121, opened 2018-01-27
Coupla episode updates

stephencelis:
- Adds a image/meta tag data
- Adds in-page linking to transcript section


Sources/PointFree/Episode/Show.swift
```diff
@@ -306,15 +314,13 @@ private let timestampLinkView = View<Int?> { timestamp -> [Node] in
...
-    a(
-      timestampLinkAttributes(timestamp) + [
-        `class`([
-          Class.display.block,
-          Class.pf.components.videoTimeLink,
-          ])
-      ],
-      [.text(encode(timestampLabel(for: timestamp)))]
-    )
+    div([id("t\(timestamp)"), `class`([Class.display.block])], [
```
stephencelis: Wrapping in a div caused a sliiightly different render.

PR #132, opened 2018-01-29
Hide empty playgrounds

stephencelis:
Small lil fix for our intro page. Can make it optional in the future.


stephencelis:
Snuck in a nullability fix for user name. We can address this differently in the future if we need to.

PR #139, opened 2018-01-30
Make tabs more tabby

mbrandonw:
Just a quick design update to make the individual/team tabs look more tabby:

screenshot 2018-01-30 09 37 17


haaakon:
Looks like a good update, might save you for future support 👍

mergesort:
Just a thought, but I was actually a bit confused in figuring out which is the currently selected tab now.

Because it's now disconnected I expected the purple color to be the selected one (because usually selected items are darker). How about maybe considering a colored border around the selected tab, or some other way to denote that the tab is selected?

mergesort:
@mbrandonw @stephencelis Just a thought, but I was actually a bit confused in figuring out which is the currently selected tab now.

Because it's now disconnected I expected the purple color to be the selected one (because usually selected items are darker). How about maybe considering a colored border around the selected tab, a shadow underneath it, or some other way to denote that the tab is selected?

stephencelis:
Good points! I think there are a few things we can do to clean it up! We'll try to tackle soon.

PR #140, opened 2018-02-01
More JavaScript

stephencelis:
Alright! So the enable/disable form logic is p nuanced and embedding it into a static string requires some thought. I've thought, so here it is. I think we can improve this in the future, but it should prevent the odd double-submit!


Sources/PointFree/Bootstrap.swift
```diff
@@ -20,6 +20,7 @@ private let stepDivider = print(message: "  -----------------------------")
 private let connectToPostgres =
   print(message: "  ⚠️ Connecting to PostgreSQL...")
     .flatMap { _ in AppEnvironment.current.database.migrate() }
+    .catch { print(message: "  ❌ Error! \($0)").flatMap(const(throwE($0))) }
```
stephencelis: I was in a weird state locally. This is nice to have tho!

stephencelis:
I don't get the screenshot diffs. Seems immaterial in GitHub. Hopefully passes!

Sources/PointFree/Bootstrap.swift
mbrandonw: Oh nice!

PR #141, opened 2018-02-01
Better chapter anchors

mbrandonw:
Improves the UX of the chapter anchors by _not_ jumping to the text content when on mobile, only for desktop.


Sources/PointFree/Episode/Show.swift
```diff
@@ -108,15 +108,23 @@ private let episodeTocView = View<(blocks: [Episode.TranscriptBlock], isEpisodeV
   )
 }
 
-private func timestampLinkAttributes(_ timestamp: Int) -> [Attribute<Element.A>] {
-  return [
-    href("#t\(timestamp)"),
+private func timestampLinkAttributes(timestamp: Int, useAnchors: Bool) -> [Attribute<Element.A>] {
```
mbrandonw: now we can customize timestamp links based on whether or not we also want to anchor them.

Sources/PointFree/Episode/Show.swift
```diff
@@ -137,19 +145,34 @@ private let tocChapterView = View<(title: String, timestamp: Int, isEpisodeViewa
...
+private let tocChapterLinkView = View<(title: String, timestamp: Int, active: Bool)> { title, timestamp, active -> [Node] in
   if active {
-    return a(
-      timestampLinkAttributes(timestamp) +
-        [`class`([Class.pf.colors.link.green, Class.type.textDecorationNone, Class.pf.type.body.regular])],
-      [text(title)]
-    )
+    return
+      [
+        div([`class`([Class.hide(.mobile)])], [
```
mbrandonw: we render two chapter links: one for mobile without anchors, one for desktop with anchors.

Sources/PointFree/TODO.swift
```diff
@@ -309,6 +309,10 @@ public func onclick<T>(javascript: StaticString) -> Attribute<T> {
   return .init("onclick", "javascript:\(javascript)")
 }
 
+public func onclick<T>(unsafeJavascript: String) -> Attribute<T> {
+  return .init("onclick", "javascript:\(unsafeJavascript)")
```
mbrandonw: new helper for being explicit when we are doing potentially unsafe things. without this I needed to create two separate scripts for the timestamp links, one that prevents default and one that doesnt.

mbrandonw:
I also got weird small changes in the numbers on the privacy page :/

Sources/PointFree/Episode/Show.swift
stephencelis: Easy peasy.

stephencelis:
I got the same changes :/ I think it's when my OS got upgraded.

PR #142, opened 2018-02-01
No rule to make target `commoncrypto-mm'

Eddpt:
When running `make` I get the following error:
```
make: *** No rule to make target `commoncrypto-mm', needed by `bootstrap'.  Stop.
```

Turns out there was a typo in the `Makefile`.


mbrandonw:
Nice catch, thanks for fixing!

PR #144, opened 2018-02-01
Misc fixes

stephencelis:
- Retains Stripe errors so we get emails (at least during auth, for now)
- Fixes the email address we mention during a change


Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -46,7 +46,7 @@ let updateProfileMiddleware =
 
       let updateFlash: Middleware<HeadersOpen, HeadersOpen, Prelude.Unit, Prelude.Unit>
       if data.email.unwrap.lowercased() != user.email.unwrap.lowercased() {
-        updateFlash = flash(.warning, "We’ve sent an email to \(data.email.unwrap) to confirm this change.")
+        updateFlash = flash(.warning, "We’ve sent an email to \(user.email.unwrap) to confirm this change.")
```
stephencelis: I think this is what you mentioned earlier, no?

mbrandonw: so with this the flash now matches the email actually being sent, but I think we actually want to send the email to `data.email` instead of `user.email`. right? like if im changing my email I need to confirm at the NEW email, not the existing one...

stephencelis: Yeah...but should we also let the old email know that the change was made in case someone hijacked the account?

stephencelis: I'm gonna merge but let's revisit this.

mbrandonw: on confirmation I believe I send to both emails letting them know it happens. tho maybe not...

PR #147, opened 2018-02-03
Small pricing update and beginning work on Apple Pay

mbrandonw:
* Prevents hitting return on teammate number spinner, which just 404's
* adds a route for apple domain association, first step for Apple Pay.


Sources/PointFree/Pricing.swift
```diff
@@ -160,11 +160,17 @@ let pricingOptionsView = View<(Database.User?, Pricing)> { currentUser, pricing
...
-                 pricingTabsView.view(pricing)
-                  + individualPricingRowView.view(pricing)
-                  + teamPricingRowView.view(pricing)
-                  + pricingFooterView.view(currentUser)
+            form(
+              [
+                action(path(to: .subscribe(nil))),
+                id(Stripe.html.formId),
+                method(.post),
+                onsubmit(javascript: "return false")
```
stephencelis: Is this the standard way of doing this? Not preventDefault?

mbrandonw: oh that works too. I just came across more suggestions of `return false` but `preventDefault` seems more standard. gonna switch

PR #148, opened 2018-02-06
\User.isAdmin

stephencelis:
This adds a proper field to our users table rather than hard-code to our emails. Should hopefully make things more flexible and secure.


Sources/PointFree/Database.swift
```diff
@@ -351,7 +354,7 @@ private func upsertUser(
 private func fetchUser(byEmail email: EmailAddress) -> EitherIO<Error, Database.User?> {
   return firstRow(
     """
-    SELECT "github_user_id", "github_access_token", "id", "name", "subscription_id"
+    SELECT "email", "github_user_id", "github_access_token", "id", "is_admin", "name", "subscription_id"
```
stephencelis: We definitely need to make this world more type-safe. I think `SELECT *` would be a quick fix so we don't forget to fetch fields, but it'd be nice to use `Decodable` to generate this logic.

mbrandonw:
happy now @paulyoung ?!

paulyoung:
I'll take a look tonight and let you know.

paulyoung:
Looks good :+1:

PR #150, opened 2018-02-06
Don't ignore errors

stephencelis:
Well...ignore them less explicitly. I've removed our use of mapping all errors to `Unit` a few more place to make error tracking easier. I haven't added additional error tracking because:

1. If Sendgrid is having issues, not sure if we can send ourselves emails _via_ Sendgrid to let us know.
2. We're no longer actively using Airtable, so didn't wanna spend too much time there.


stephencelis:
All these PRs have the same screen shot noise for our privacy page. Sorry!

PR #151, opened 2018-02-06
Add subscribe callout to free episodes

stephencelis:
Just a nice lil cross-link for interested folks.


Sources/PointFreeTestSupport/TestCase.swift
```diff
@@ -16,7 +16,7 @@ open class TestCase: XCTestCase {
       )
     )
 
-    _ = try! execute("DROP SCHEMA public CASCADE")
+    _ = try! execute("DROP SCHEMA IF EXISTS public CASCADE")
```
stephencelis: I canceled running the full suite in just the wrong moment, where `public` had been dropped. Easy fix!

mbrandonw: oh nice I have also run into this

Sources/PointFree/Episode/Show.swift
```diff
@@ -261,10 +270,14 @@ private let subscribeView = View<(Database.User?, Episode)> { user, episode -> N
...
-        This episode is for subscribers only. To access it, and all past and future episodes, become a
-        subscriber today!
-        """
+        episode.subscriberOnly
+          ? """
+            This episode is for subscribers only. To access it, and all past and future episodes, become a
+            subscriber today!
+            """
+          : """
+            Subscribe today and unlock all episodes!
```
mbrandonw: How about: "This episode is free to all users. To get access to all past and future episodes, become a subscriber today!"

PR #152, opened 2018-02-06
Past-due banner

stephencelis:
Adds a banner for the past-due state.


Sources/PointFree/PageLayout.swift
```diff
@@ -118,7 +118,8 @@ func simplePageLayout<A>(_ contentView: View<A>) -> View<SimplePageLayoutData<A>
           <> favicons
         ),
         body(
-          (layoutData.flash.map(flashView.view) ?? [])
+          pastDueBanner(layoutData.currentSubscriptionStatus)
+            <> (layoutData.flash.map(flashView.view) ?? [])
```
stephencelis: Just reusing flash for now.

Sources/PointFree/PageLayout.swift
```diff
@@ -128,6 +129,21 @@ func simplePageLayout<A>(_ contentView: View<A>) -> View<SimplePageLayoutData<A>
...
 
+func pastDueBanner(_ subscriptionStatus: Stripe.Subscription.Status?) -> [Node] {
+  guard subscriptionStatus == .some(.pastDue) else { return [] }
+
+  return flashView.view(
+    .init(
+      priority: .warning,
+      message: """
+      Your subscription is past-due! Please
+      [update your payment info](\(path(to: .account(.paymentInfo(.show))))) to ensure access to
```
mbrandonw: one bummer is that the link isnt underlined. this is cause we globally turn of underlined links, and then opt into on  a case-by-case basis with the `underline` utility class. we cant do that here because of the markdown. 

could we make a version of flash that is allowed to take `Node`s so that we can underline? 

in the future it'd be cool if:

1) `Node` was serializable so that we could use actual nodes in the flash
2) we turn back on global underline, and instead opt out of it on a case by case basis instead of the opposite.

stephencelis: Yeah, what's strange is I swear when I first added markdown support to banners, I added a bunch of email links and they highlighted just fine

stephencelis: (Codable `Node` was my first idea but a bit complicated. Can revisit soon!)

PR #156, opened 2018-02-11
Add exercises to transcript

stephencelis:
Very basic support for now.


Sources/PointFree/Episode/Show.swift
```diff
@@ -339,10 +339,25 @@ let dividerView = View<Prelude.Unit> { _ in
 
 private let transcriptView = View<[Episode.TranscriptBlock]> { blocks in
   div([`class`([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]]), Class.pf.colors.bg.white])],
-      blocks.flatMap(transcriptBlockView.view)
+      blocks.filter((!) <<< ^\.type.isExercise).flatMap(transcriptBlockView.view)
```
stephencelis: I'm filtering them from the inline view for now since I'm not sure how to style them nicely.

Sources/PointFree/Episode/Show.swift
```diff
@@ -339,10 +339,25 @@ let dividerView = View<Prelude.Unit> { _ in
...
+        + exercisesView.view(blocks.filter(^\.type.isExercise))
```
stephencelis: At the bottom we accumulate them all.

Sources/PointFree/Episode/Show.swift
```diff
@@ -353,6 +368,12 @@ private let transcriptBlockView = View<Episode.TranscriptBlock> { block -> Node
       )
       ])
 
+  case .exercise:
+    return div(
+      timestampLinkView.view(block.timestamp)
```
stephencelis: The one strange thing here is that we probably want them rendering one way in the body, another way at the end. Got any ideas?

PR #159, opened 2018-02-20
Add billing name to card form

stephencelis:
This should hopefully allow folks to subscribe when more verification is required.


Sources/PointFree/Account/Cancel.swift
```diff
@@ -75,7 +75,7 @@ private func cancel(_ conn: Conn<StatusLineOpen, (Stripe.Subscription, Database.
         either(
           const(
             conn |> redirect(
-              to: .account(.subscription(.changeSeats(.show))),
+              to: .account(.index),
```
stephencelis: When I was canceling locally I got caught in this flow. Need to investigate more but the redirect was definitely wrong.

Sources/PointFree/Account/PaymentInfo.swift
```diff
@@ -65,7 +65,7 @@ let paymentInfoView = View<(Stripe.Subscription, Database.User)> { subscription,
       div([`class`([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]])])],
           titleRowView.view(unit)
             <> (subscription.customer.sources.data.first.map(currentPaymentInfoRowView.view) ?? [])
-            <> updatePaymentInfoRowView.view(unit)
+            <> updatePaymentInfoRowView.view(currentUser.name)
```
stephencelis: Can default the card form with the GitHub name.

PR #160, opened 2018-02-21
Expose extra billing fields

stephencelis:
This allows folks to tack on `?expand=true` and enter their address and/or VAT number if required. It expands the "update payment info" form by default, though showing VAT (even with the "Optional" parenthetical) is admittedly confusing. I'm cool addressing later, but if you have a quick idea to fix, lemme know!


Sources/PointFree/Account/Account.swift
```diff
@@ -293,7 +293,7 @@ private func mainAction(for subscription: Stripe.Subscription) -> Node {
     return a(
       [
         `class`([Class.pf.components.button(color: .purple, size: .small)]),
-        href(path(to: .pricing(nil)))
+        href(path(to: .pricing(nil, expand: nil)))
```
stephencelis: FWIW I was gonna just do `.pricing(Pricing, address: Bool?, vat: Bool?)` but we need to update our router to support more matches.

mbrandonw: whatdya mean more matches? it didn't work to do `queryParam("vat", opt(.bool))`?

stephencelis: Yeah I think our router can only match 3 times right now.

mbrandonw: it should be able to do more no prob, as long as there is a `parenthesize` defined for it:

https://github.com/pointfreeco/swift-web/blob/34b4143b4153f4b624364203e55e61e2fa9edc38/Sources/ApplicativeRouter/PartialIso.swift#L71

may have been something else

PR #164, opened 2018-02-25
Move server code into this repo!

mbrandonw:
I think we previously couldn't do this because we ran tests on the iOS simulator, and Kitura + some deps dont build for iOS. But we no longer do that, so think we can remove a layer of dependencies!


Sources/PointFree/__Generated__/DerivedPartialIsos.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.10.1 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.9.0 — https://github.com/krzysztofzablocki/Sourcery
```
stephencelis: Lol. Get with the times.

PR #166, opened 2018-02-26
Fix Docker command

mbrandonw:
I just deployed and momentarily brought down the site... oops!

I found this in the logs:

```
2018-02-26T01:42:45.558032+00:00 app[web.2]: /bin/sh: 1: ./.build/release/pointfreeco: not found
```

and realized there was another server reference I didn't change in #164

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get install -y postgresql libpq-dev
 WORKDIR /app
 
 COPY Makefile ./
-RUN make install-cmark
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,12 +10,12 @@ RUN apt-get install -y postgresql libpq-dev
 WORKDIR /app
 
 COPY Makefile ./
-RUN make install-cmark
+RUN make linux-install-cmark
```
mbrandonw: also this...

PR #170, opened 2018-03-03
Remove name from subscription form

stephencelis:
Also exposes "expand" as hidden option from payment info. Preliminary work for changing around the subscription form in general.


Sources/PointFree/Pricing.swift
```diff
@@ -470,17 +470,17 @@ private let pricingFooterView = View<(Database.User?, Bool)> { currentUser, expa
       div(
         [`class`([Class.padding([.mobile: [.top: 2, .bottom: 3]])])],
         currentUser
-          .map { stripeForm.view(($0.name, expand)) }
-          ?? [gitHubLink(text: "Sign in with GitHub", type: .black, redirectRoute: .pricing(nil, expand: nil))]
+          .map(const(stripeForm.view(expand)))
```
stephencelis: Thinking we should make `const` lazy using `@autoclosure`...

PR #171, opened 2018-03-04
Introduce a Team Monthly plan

stephencelis:
mar-04-2018 12-25-52

This allows folks to sign up for a team plan on a monthly billing basis! Still need some work on the account page to allow for upgrades/downgrades/crossgrades, but that'll come soon!


Sources/PointFree/Pricing.swift
```diff
@@ -10,11 +10,11 @@ import Prelude
 import Styleguide
 import Tuple
 
-public enum Pricing: Codable, DerivePartialIsos {
-  case individual(Billing)
-  case team(Int)
+public struct Pricing {
```
stephencelis: Don't mind me just using algebra with my types.

Sources/PointFree/Pricing.swift
```diff
@@ -25,90 +25,64 @@ public enum Pricing: Codable, DerivePartialIsos {
...
-    switch self {
-    case .team:
-      return true
-    case .individual:
-      return false
+    return self.lane == .team
+  }
+}
+
+extension Pricing: Codable {
```
stephencelis: Being a struct means we could auto-derive here, but we do some fancy things on the subscription page with shown/not-shown fields, so this is a bit safer.

Sources/PointFree/Pricing.swift
```diff
@@ -165,12 +139,13 @@ let pricingOptionsView = View<(Database.User?, Pricing, Bool)> { currentUser, pr
...
                 id(Stripe.html.formId),
                 method(.post),
-                onsubmit(javascript: "event.preventDefault();")
+                onsubmit(javascript: "event.preventDefault()")
               ],
               pricingTabsView.view(pricing)
-                + individualPricingRowView.view(pricing)
-                + teamPricingRowView.view(pricing)
-                + pricingFooterView.view((currentUser, expand))
+                <> [div([`class`([Class.margin([.mobile: [.bottom: 3]])])], [])]
```
stephencelis: Just letting a semigroup be a semigroup.

Sources/PointFree/Pricing.swift
```diff
@@ -274,7 +249,7 @@ private let whoAreYou = View<Prelude.Unit> { _ in
         "We’re ",
         a([href("http://www.fewbutripe.com"), style(faqLinkStyles)], ["Brandon Williams"]),
         " and ",
-        a([href("http://www.stephencelis.com"), style(faqLinkStyles)], ["Stephen Celis"]),
+        a([href("https://www.stephencelis.com"), style(faqLinkStyles)], ["Stephen Celis"]),
```
stephencelis: FYI.

Sources/PointFree/Pricing.swift
```diff
@@ -358,91 +349,86 @@ private func isChecked(_ billing: Pricing.Billing, _ pricing: Pricing) -> Bool {
...
+          id(radioId(for: billing)),
+          name("pricing[billing]"),
           type(.radio),
-          value($0.billing.rawValue),
+          value(billing.rawValue),
           ]),
         gridColumn(sizes: [.mobile: 12], [], [
-          h2([`class`([Class.pf.type.title2, Class.type.light, Class.pf.colors.fg.gray650])], [.text(encode(pricingText(for: $0.billing)))]),
+          h2([`class`([Class.pf.type.responsiveTitle2, Class.type.light, Class.pf.colors.fg.gray650])], [
+            span([`class`([selectors.content.0])], [
```
stephencelis: Scoping these using CSS still to avoid writing more JavaScript...

Sources/PointFree/Pricing.swift
```diff
@@ -536,28 +540,28 @@ private let selectors = (
     CssSelector.id("tab1")
   ),
   content: (
-    CssSelector.id("content0"),
-    CssSelector.id("content1")
+    CssSelector.class("content0"),
```
stephencelis: Changed these to use `class` selectors so we can hide/show things more granularly.

Sources/PointFree/Pricing.swift
```diff
@@ -536,28 +540,28 @@ private let selectors = (
...
+    CssSelector.class("content1")
   )
 )
 
 private func tabStyles(
-  idSelectors: [(input: CssSelector, content: CssSelector)],
-  showTabStyles: Stylesheet = display(.flex)
+  selectors: [(input: CssSelector, content: CssSelector)],
+  showTabStyles: Stylesheet = display(.inherit)
```
stephencelis: This is kind of a hack, but works! Just need to make sure the parent node uses the proper display.

Sources/PointFree/Pricing.swift
mbrandonw: lol

Sources/PointFree/Pricing.swift
```diff
@@ -10,11 +10,11 @@ import Prelude
...
-  case individual(Billing)
-  case team(Int)
+public struct Pricing {
+  let billing: Billing
+  let quantity: Int
 
-  public static let `default` = individual(.monthly)
+  public static let `default` = Pricing(billing: .monthly, quantity: 1)
 
   public static let teamYearlyBase = 160
```
mbrandonw: this caught my eye, doesn't actually seem to be used anymore...

Sources/PointFree/Pricing.swift
mbrandonw: ah nice, that works! cool that we have the screenshot tests that execute JS to switch tabs!

Tests/PointFreeTests/PricingTests.swift
```diff
@@ -35,9 +35,9 @@ class PricingTests: TestCase {
         
         webView.evaluateJavaScript(
           """
-          document.getElementById("tab0").checked = false;
-          document.getElementById("tab1").checked = true;
-          var quantity = document.getElementsByName("pricing[team]")[0];
+          document.getElementById('tab0').checked = false;
+          document.getElementById('tab1').checked = true;
+          var quantity = document.getElementsByName('pricing[quantity]')[0];
```
mbrandonw: lol I heard that single quotes are more performant than double quotes......

jkjk

stephencelis: Lolol we use single quotes elsewhere so...consistency dragging me. Makes it at least nice to do non-multiline JS where we'd have to escape all `"`s.

PR #172, opened 2018-03-05
Subscription changes

stephencelis:
Adds a consolidated subscription change page and deletes a bunch of older one-off pages.


Sources/PointFree/Account/Change.swift
```diff
@@ -0,0 +1,392 @@
...
+import HttpPipelineHtmlSupport
+import Optics
+import Prelude
+import Styleguide
+import Tuple
+
+// MARK: Middleware
+
+let subscriptionChangeShowResponse =
+  filterMap(require1 >>> pure, or: loginAndRedirect)
```
mbrandonw: we do this a lot so maybe we just finally make a `requireCurrentUser` middleware transformers to act on `T2<User?, Z>`

Sources/PointFree/Account/Change.swift
```diff
@@ -0,0 +1,392 @@
...
+      let invitesAndTeammates = sequence([
+        parallel(AppEnvironment.current.database.fetchTeamInvites(user.id).run)
+          .map { $0.right?.count ?? 0 },
+        parallel(AppEnvironment.current.database.fetchSubscriptionTeammatesByOwnerId(user.id).run)
+          .map { $0.right?.count ?? 0 }
+        ])
+
+      return invitesAndTeammates
+        .sequential
+        .flatMap { middleware(conn.map(const(user .*. $0.reduce(1, +) .*. conn.data.second))) }
```
mbrandonw: we could cook up an `insert1`, `insert2`, ... family of operators for inserting values into tuples:

```swift
func insert1<A, B, Z>(_ a: A) -> (T2<B, Z>) -> T3<A, B, Z> {
  return { t in 
    return a .*. t.first .*. t.second
  }
}
```

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -41,30 +41,13 @@ public enum Route: DerivePartialIsos {
...
-
-      public enum Downgrade: DerivePartialIsos {
-        case show
-        case update
-      }
-
-      public enum Upgrade: DerivePartialIsos {
-        case show
-        case update
+        case update(Pricing?)
```
mbrandonw: so much simpler!

Sources/PointFree/Account/Change.swift
```diff
@@ -0,0 +1,392 @@
...
+            headersMiddleware: flash(.notice, "We’ve modified your subscription.")
+          )
+        }
+    )
+}
+
+func requireActiveSubscription<A>(
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, T3<Stripe.Subscription, Database.User, A>, Data>
+  )
+  -> Middleware<StatusLineOpen, ResponseEnded, T3<Stripe.Subscription, Database.User, A>, Data> {
```
mbrandonw: it doesn't look like `User` is used in this middleware so could it just be `T2<Subscription, Z>`?

Sources/PointFree/Account/Change.swift
```diff
@@ -0,0 +1,392 @@
...
+                .error,
+                """
+                We couldn’t modify your subscription at this time. Please try again or contact
+                <support@pointfree.co>.
+                """
+              )
+            )
+          )
+        ) { _ in
+          // TODO: Send email?
```
mbrandonw: if you trello a list of emails that you think should be sent I could take care of that!

Sources/PointFree/Account/Change.swift
```diff
@@ -0,0 +1,392 @@
...
+      h3([`class`([Class.pf.type.responsiveTitle4])], [text(subtitle)]),
+      p(description + [" ",
+        """
+        Additional costs will be billed immediately, prorated against the remaining time of
+        the current billing period.
+        """
+        ]),
+      input([
+        type(.number),
+        min(max(1, seatsTaken)),
```
mbrandonw: if they choose `1` does it change them to individual? or should we use `2` as the minimum?

mbrandonw: oh I see, the individual vs team is only a distinction we make based on quantity. cool!

Sources/PointFree/Account/Change.swift
```diff
@@ -0,0 +1,392 @@
...
+        p([
+          "You are currently enrolled in the ", strong([text(subscription.plan.name)]), " plan. ",
+          "Your subscription will ",
+          subscription.isRenewing ? "renew" : "end",
+          " on ",
+          strong([text(dateFormatter.string(from: subscription.currentPeriodEnd))]),
+          ".",
+          subscription.isRenewing
+            ? ""
+            : " Reactivate your subscription by submitting the form below."
```
mbrandonw: 2 things:

1) When I tested locally to cancel my membership, I got a `We couldn’t cancel your subscription at this time.` error flash, but it did seem to actually cancel in stripe.

2) after canceling my account page becomes:

screenshot 2018-03-07 08 41 24

And that `Reactivate` button does the reactivate action at once and sends me back to account. So, is this state in the change screen possible?

stephencelis: 1. Good catch. I'll try to repro figure out what's going on.
2. Yeah, I made the change to show the reactivation info in "change" just in case we wanted to change this. Hitting the change page is totally possible when you've canceled but it's still pending, but I didn't yet link to it. Thoughts?

PR #176, opened 2018-03-11
Add spacebar keyboard shortcut for play/pause

### LINKED ISSUE #173: [Feature Request] Support keyboard shortcuts for play/pause when video streaming in full-screen
I'm using Chrome, and I think it would be very helpful if pressing `␣` can toggle play/pause when video streaming in full-screen.

stephencelis:
Fixes #173.


Sources/PointFree/Episode/Show.swift
```diff
@@ -62,7 +62,27 @@ let episodeView = View<(Database.User?, Stripe.Subscription.Status?, Episode)> {
           )
         ]
       )
-      ])
+      ]),
+
+    script(
+      """
+      var hasPlayed = false;
```
stephencelis: I don't enable the shortcut till someone's started to play the video. In Chrome with auto-play, the shortcut is enabled immediately. For Safari, it's lazy, so folks don't get spacebar highjacked till they start watching.

PR #177, opened 2018-03-13
Episode credits

mbrandonw:
### What

This adds basic support for "episode credits". If a user has a "credit" they can get any episode for free. Includes:

* Admin tool for giving any user a particular episode for free
* Account page shows how many credits the user has and what episodes they have gotten with credits
* Ability to use a credit from the episode page.

Some things we will want to do soon:

* Automatically give users a credit when they sign up
* Sign up callouts to tell users they will get a free episode
* Give them another credit after some amount of time has passed 
* Admin tool to give credits to users
* and more!


Sources/PointFree/Account/Account.swift
```diff
@@ -46,37 +46,118 @@ private func fetchAccountData<I, A>(
     )
     ?? pure(nil)
 
-  return
-    zip(
-      parallel(subscription),
+  return zip5(
```
mbrandonw: was getting worried with overloads so decided to just give it a unique name

Sources/PointFree/Admin/EpisodeCredit.swift
```diff
@@ -0,0 +1,77 @@
...
+
+let showEpisodeCreditsMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple1<Database.User?>, Data> =
+  requireAdmin
+    <| writeStatus(.ok)
+    >-> respond(showEpisodeCreditsView.contramap(const(unit)))
+
+let addEpisodeCreditMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple3<Database.User?, String?, Int?>, Data> =
+  requireAdmin
+    <<< filterMap(
+      over2(fetchUser(uuidString:)) >>> sequence2 >>> map(require2),
```
mbrandonw: intense, but it works. transforms a 

```swift
Tuple3<Database.User, String?, Int?>
```

into a 

```swift
Tuple3<Database.User, Database.User, Int?>
```

Sources/PointFree/Episode/Show.swift
```diff
@@ -38,7 +38,125 @@ let episodeResponse: Middleware<StatusLineOpen, ResponseEnded, Tuple4<Either<Str
     }
 )
 
-let episodeView = View<(Database.User?, Stripe.Subscription.Status?, Episode)> { user, subscriptionStatus, episode in
+let useCreditResponse =
+  filterMap(
+    over1(episode(forParam:)) >>> require1 >>> pure,
+    or: writeStatus(.notFound) >-> respond(episodeNotFoundView.contramap(lower))
+    )
+    <<< { userEpisodePermission >-> $0 } // Is there a name for this and/or do we want to introduce someting like "lift" to do it?
```
mbrandonw: this is kinda interesting. `userEpisodePermission` is just a plain old middleware, but we can easily lift it into a transformer by doing `{ userEpisodePermission >-> $0 }`. wonder if we should give that a name.

Sources/PointFree/Episode/Show.swift
```diff
@@ -88,30 +208,25 @@ let episodeView = View<(Database.User?, Stripe.Subscription.Status?, Episode)> {
...
 
-private func isEpisodeViewable(_ episode: Episode, _ subscriptionStatus: Stripe.Subscription.Status?) -> Bool {
-  return !episode.subscriberOnly || subscriptionStatus == .some(.active)
-}
-
-private let downloadsAndCredits =
+private let downloadsAndHosts =
   downloadsView
-    <> creditsView.contramap(const(unit))
+    <> hostsView.contramap(const(unit))
```
mbrandonw: renamed so that credits refers to episode credits instead of the lil host blurb

Sources/PointFree/Episode/Show.swift
```diff
@@ -499,5 +709,90 @@ func unsafeMark(from markdown: String) -> String {
...
+private func isSubscribeBannerVisible(for permission: EpisodePermission) -> Bool {
+  switch permission {
+  case .loggedIn(_, .isSubscriber):
+    return false
+  case .loggedIn(_, _), .loggedOut(_):
+    return true
+  }
+}
+
+private enum EpisodePermission {
```
mbrandonw: this is how I model what kind of permission a user has for an episode. we wanna be able to message depending on if you are logged in/out, subscriber or not, and have used a credit or not.

Sources/PointFree/Account/Account.swift
```diff
@@ -46,37 +46,118 @@ private func fetchAccountData<I, A>(
...
+  a(
+    [
+      href(path(to: .episode(.left(episode.slug)))),
+      `class`(
+        [
+          Class.pf.colors.link.purple
+        ]
+      )
+    ],
+    [.text(encode("#\(episode.sequence): \(episode.title)"))]
```
stephencelis: Can use the `text` function.

Sources/PointFree/Admin/EpisodeCredit.swift
stephencelis: Ah nice, makes sense! Wish we had row types to annotate.

stephencelis: (One thought is to `map` to `UUID.init >>> User.Id.init` so it gets captured in the sig, but it's like a minor readability improvement.)

Sources/PointFree/Database.swift
```diff
@@ -69,8 +73,19 @@ public struct Database {
     }
   }
 
+  public struct EpisodeCredit: Decodable {
```
stephencelis: Very cool.

Sources/PointFree/Database.swift
```diff
@@ -354,8 +373,14 @@ private func upsertUser(
 private func fetchUser(byEmail email: EmailAddress) -> EitherIO<Error, Database.User?> {
   return firstRow(
     """
-    SELECT "email", "github_user_id", "github_access_token", "id", "is_admin", "name", "subscription_id"
-    FROM "users"
```
stephencelis: How's this still working without `FROM "users"`?

Sources/PointFree/Database.swift
```diff
@@ -69,8 +73,19 @@ public struct Database {
...
+
+    public enum CodingKeys: String, CodingKey {
+      case episodeSequence = "episode_sequence"
+      case userId = "user_id"
+    }
+  }
+
   public struct User: Decodable {
     public internal(set) var email: EmailAddress
+    public internal(set) var episodeCreditCount: Int
```
stephencelis: One thought is that credits count could be derived from "unfulfilled" credit rows in the credits table...not sure if it's worth the refactor, though.

Sources/PointFree/Episode/Show.swift
stephencelis: That's neat! Will have to think on it. Maybe we wait for more use cases?

Sources/PointFree/Episode/Show.swift
```diff
@@ -38,7 +38,125 @@ let episodeResponse: Middleware<StatusLineOpen, ResponseEnded, Tuple4<Either<Str
...
+
+  guard user.episodeCreditCount > 0 else {
+    return conn
+      |> redirect(
+        to: .episode(.left(episode.slug)),
+        headersMiddleware: flash(.error, "You do not have any credits to use.")
+    )
+  }
+
+  return AppEnvironment.current.database.addEpisodeCredit(episode.sequence, user.id)
```
stephencelis: Lol sorry to get bikeshed-y, but "add credit" feels like adding to credit for later redeeming. Maybe this should be `redeemCredit`?

Sources/PointFree/Episode/Show.swift
```diff
@@ -38,7 +38,125 @@ let episodeResponse: Middleware<StatusLineOpen, ResponseEnded, Tuple4<Either<Str
...
+            |> redirect(
+              to: .episode(.left(episode.slug)),
+              headersMiddleware: flash(.notice, "You now have access to this episode!")
+          )
+        )
+      )
+  )
+}
+
+private func validateCreditRequest<Z>(
```
stephencelis: Nice!

Sources/PointFree/Episode/Show.swift
```diff
@@ -38,7 +38,125 @@ let episodeResponse: Middleware<StatusLineOpen, ResponseEnded, Tuple4<Either<Str
...
+    )
+  }
+}
+
+private func userEpisodePermission<I, Z>(
+  _ conn: Conn<I, T4<Episode, Database.User?, Stripe.Subscription.Status?, Z>>
+  )
+  -> IO<Conn<I, T5<EpisodePermission, Episode, Database.User?, Stripe.Subscription.Status?, Z>>> {
+
+    let (episode, currentUser, subscriptionStatus) = (get1(conn.data), get2(conn.data), get3(conn.data))
```
stephencelis: Don't we have a restructure function for this? At the very least with `_` for ignored args?

stephencelis: Don't we have a destructure function for this? At the very least with `_` for ignored args?

Sources/PointFree/Episode/Show.swift
```diff
@@ -88,30 +208,25 @@ let episodeView = View<(Database.User?, Stripe.Subscription.Status?, Episode)> {
...
   video(
     [
-      autoplay(true),
       `class`([Class.size.width100pct]),
       controls(true),
       playsinline(true),
-      poster(episode.image),
-      tabindex(0)
+      tabindex(0),
+      poster(episode.image)
```
stephencelis: RIP alphabetize.

Sources/PointFree/Episode/Show.swift
```diff
@@ -264,65 +379,153 @@ private func timestampLabel(for timestamp: Int) -> String {
...
           ? transcriptView.view(episode.transcriptBlocks)
           : []
     )
   )
 }
 
-private let subscribeView = View<(Database.User?, Episode)> { user, episode -> Node in
-
-  div([`class`([Class.type.align.center, Class.margin([.mobile: [.all: 4], .desktop: [.all: 4]]), Class.padding([.mobile: [.top: 1, .leftRight: 1, .bottom: 3], .desktop: [.top: 2, .leftRight: 2]]), Class.pf.colors.bg.gray900])], [
+private func subscribeBlurb(for permission: EpisodePermission) -> StaticString {
```
stephencelis: Nice.

Sources/PointFree/Episode/Show.swift
```diff
@@ -264,65 +379,153 @@ private func timestampLabel(for timestamp: Int) -> String {
...
+        `class`(
+          [
+            Class.pf.type.body.regular,
+            Class.padding([.mobile: [.top: 4, .bottom: 2]])
+          ]
+        )
+      ],
+      [
+        """
+        You currently have 1 episode credit available. Do you want to use it to view this episode for free
```
stephencelis: Hard-coded?

Sources/PointFree/Episode/Show.swift
```diff
@@ -356,14 +559,21 @@ let topLevelEpisodeInfoView = View<Episode> { ep in
...
+  div(
+    [
+      `class`(
+        [
+          Class.padding([.mobile: [.all: 3], .desktop: [.leftRight: 4, .bottom: 4, .top: 2]]),
+          Class.pf.colors.bg.white
+        ]
+      )
+    ],
+    blocks.filter((!) <<< ^\.type.isExercise).flatMap(transcriptBlockView.view)
```
stephencelis: Noice!

Sources/PointFree/Episode/Show.swift
stephencelis: V cool.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -202,10 +218,13 @@ private let routers: [Router<Route>] = [
 
   .team <<< .remove
     <¢> post %> lit("account") %> lit("team") %> lit("members")
-    %> pathParam(.rawRepresentable >>> .rawRepresentable)
+    %> pathParam(.uuid >>> .tagged)
```
stephencelis: 😄

Sources/PointFree/TODO.swift
```diff
@@ -113,6 +113,30 @@ public func zip<A, B, C, D>(
   return tuple4 <¢> a <*> b <*> c <*> d
 }
 
+public func zip5<A, B, C, D, E>(
```
stephencelis: Rename the others in another PR?

Sources/PointFree/TODO.swift
```diff
@@ -186,6 +210,12 @@ extension PartialIso where A == String, B == String {
   }
 }
 
+public func sequence2<A, B, Z>(_ t: T3<A, IO<B>, Z>) -> IO<T3<A, B, Z>> {
```
stephencelis: Trippy.

Sources/PointFree/TODO.swift
```diff
@@ -209,3 +239,11 @@ public func payload<A, B>(
         return "\(first)\(separator)\(second)"
     })
 }
+
+public func zurry<A>(_ f: () -> A) -> A {
```
stephencelis: TODO?

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -44,11 +44,13 @@ extension Mailgun {
 
 extension Database {
   public static let mock = Database(
+    addEpisodeCredit: { _, _ in hole() },
```
stephencelis: Hehe. Untested?

Sources/PointFree/Admin/EpisodeCredit.swift
mbrandonw: done! b7696ed

Sources/PointFree/Account/Account.swift
mbrandonw: always forget about that...

Sources/PointFree/Database.swift
mbrandonw: jeez, good catch. it definitely wouldn't be. but also doesn't seem we actually use this endpoint, so gonna remove.

Sources/PointFree/Database.swift
mbrandonw: ah so that's not how the episode credits are structured. the "source of truth" of how many credits one has is this field. If this field is >0 you will get the "use credit" button on a subscriber only episode.

The `EpisodeCredit` table holds the credits that were used. So if a row is present for a (episode, user), they get access to that episode.

Does that make sense?

Sources/PointFree/Episode/Show.swift
mbrandonw: yeah def, can do

Sources/PointFree/Episode/Show.swift
mbrandonw: the `lower` function only works on `TupleN` types, not `TN` types because it's not clear what should be done with the `Z` continuation piece. it's pretty annoying, but haven't figured out the best way to handle yet.

Sources/PointFree/Episode/Show.swift
mbrandonw: ha oops

Sources/PointFree/Episode/Show.swift
mbrandonw: oops, fixed

Sources/PointFree/TODO.swift
mbrandonw: sup?

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
mbrandonw: yikes. ya lots of tests I can write here. will do...

mbrandonw:
just pushed a bunch of clean up and tests. if tests pass I might merge!

Sources/PointFree/Database.swift
stephencelis: @mbrandonw It totally makes sense. I was just suggesting an alternate structure so the single source of truth would be rows in the credits table.

PR #180, opened 2018-03-13
Free episode email

mbrandonw:
Allows us to send an email to non-subscribers when we release an episode for free.


Sources/PointFree/Admin/FreeEpisodes.swift
```diff
@@ -0,0 +1,111 @@
...
+private func sendEmail(forFreeEpisode episode: Episode, toUsers users: [Database.User]) -> EitherIO<Prelude.Unit, Prelude.Unit> {
+
+  // A personalized email to send to each user.
+  let freeEpisodeEmails = users.map { user in
+    lift(IO { inj2(freeEpisodeEmail.view((episode, user))) })
+      .flatMap { nodes in
+        sendEmail(
+          to: [user.email],
+          subject: "Free Point-Free Episode: \(episode.title)",
+          unsubscribeData: (user.id, .newEpisode),
```
mbrandonw: I'm sending this email under the "new episode" email, but it could maybe also be an announcement email. or a new newsletter. @stephencelis what do you think?

Sources/PointFree/Admin/FreeEpisodes.swift
```diff
@@ -0,0 +1,111 @@
...
+import Html
+import HtmlCssSupport
+import HttpPipeline
+import HttpPipelineHtmlSupport
+import Optics
+import Prelude
+import Styleguide
+import Tuple
+
+let indexFreeEpisodeEmailMiddleware =
```
mbrandonw: all the stuff in this file is very similar to what we do for new episode emails, so I could prob combine a lot of this code later.

Sources/PointFree/Emails/FreeEpisodeEmail.swift
```diff
@@ -0,0 +1,115 @@
...
+      user: user,
+      newsletter: .newEpisode,
+      title: "Free Point-Free Episode: \(ep.title)",
+      preheader: freeEpisodeBlurb,
+      data: ep
+    )
+}
+
+let freeEpisodeBlurb = """
+Every once in awhile we release a past episode for free to all of our viewers, and today is that day!
```
mbrandonw: extracted this out so that we could use as the `preheader` in the email.

Sources/PointFree/Admin/FreeEpisodes.swift
stephencelis: I think this is fine for now.

PR #185, opened 2018-03-19
Adds a better sign up callout on episode page

mbrandonw:


Sources/PointFree/Episode/Show.swift
```diff
@@ -453,9 +453,9 @@ private let creditBlurb = View<(EpisodePermission, Episode)> { permission, episo
       ],
       [
         text("""
-        You currently have \(user.episodeCreditCount) episode credit available. Do you want to use it to
-        view this episode for free right now?
-        """)
+          You currently have \(user.episodeCreditCount) episode credit available. Do you want to use it to
```
stephencelis: Pluralize `credit`?

mbrandonw: doneso

PR #190, opened 2018-03-25
Move public episodes to this repo

### LINKED ISSUE #181: Issue running playgrounds locally
First of all, I really love the work you guys are doing to help everyone learn FP concepts in Swift. It's been immensely helpful.

In order to better understand what, why and how you are writing software. I'm able to run the server locally using `make bootstrap-oss` and `make run-oss` however however I've run into one issue when trying to build the server in xCode so I can see how you are using playground-driven development for the view layer.

I get the following error when trying to build the server in xCode (On line 40 of `main.swf`).

```
AppEnvironment.push(\.episodes .~ allEpisodes) // Use of unresolved identifier 'allEpisodes'
```

How can I resolve this so I can play with the playgrounds implementation?

Thanks!

mbrandonw:
As we make episodes permanently public we should move them to this repo so that people can see them in local development. I also fixed #181 by making a separate OSS config.

### DIFF
--- a/Sources/PointFree/PublicEpisodes/0001-Functions.swift
+++ b/Sources/PointFree/PublicEpisodes/0001-Functions.swift
@@ -1,0 +1,848 @@
+import Foundation
+import PointFree
+
+let ep1 = Episode(
+  blurb: """
+Our first episode is all about functions! We talk a bit about what makes functions special, contrasting them with the way we usually write code, and have some exploratory discussions about operators and composition.
+""",
+  codeSampleDirectory: "0001-functions",
+  id: .init(unwrap: 1),
+  exercises: [],
+  image: "https://d1hf1soyumxcgv.cloudfront.net/0001-functions/0001-poster.jpg",
+  length: 1219,
+  publishedAt: Date(timeIntervalSince1970: 1_517_206_269),
+  sequence: 1,
+  sourcesFull: [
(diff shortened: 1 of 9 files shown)


Sources/Server/main.swift
```diff
@@ -37,7 +37,22 @@ AppEnvironment.push(\.envVars .~ envVars)
 
 // Transcripts
 
-AppEnvironment.push(\.episodes .~ allEpisodes)
+#if OSS
+  private let allEpisodes = allPublicEpisodes
```
mbrandonw: this should help with #181

mbrandonw:
> Nice! Simpler than I thought! Do we need to update the README?

I don't think so. The readme says to run `make boostrap-oss` which should do everything.

mbrandonw:
Just figuring out running tests with OSS configuration and then this should be good to go.

Sources/PointFree/PublicEpisodes/0001-Functions.swift
```diff
@@ -0,0 +1,848 @@
+import Foundation
+import PointFree
```
stephencelis: Can remove this import.

mbrandonw: ah yes good catch

PR #193, opened 2018-03-28
Fix some upgrade logic

stephencelis:
Coupla small gotchas.


Sources/PointFree/Account/Change.swift
```diff
@@ -56,12 +56,20 @@ private func subscriptionChange(_ conn: Conn<StatusLineOpen, (Stripe.Subscriptio
 
     let (currentSubscription, _, _, newPricing) = conn.data
 
-    let newPrice = defaultPricing(for: newPricing.lane, billing: newPricing.billing) * 100
-    let currentPrice = currentSubscription.plan.amount.rawValue
+    let newPrice = (defaultPricing(for: newPricing.lane, billing: newPricing.billing) * 100) * newPricing.quantity
```
stephencelis: Weren't considering quantity at all in proration 🤦‍♂️

Sources/PointFree/Account/Change.swift
```diff
@@ -56,12 +56,20 @@ private func subscriptionChange(_ conn: Conn<StatusLineOpen, (Stripe.Subscriptio
...
+    let currentPrice = currentSubscription.plan.amount.rawValue * currentSubscription.quantity
 
     let prorate = newPrice > currentPrice
 
     return AppEnvironment.current.stripe.updateSubscription(currentSubscription, newPricing.plan, newPricing.quantity, prorate)
+      .flatMap { sub -> EitherIO<Error, Stripe.Subscription> in
+        if prorate {
+          parallel(AppEnvironment.current.stripe.invoiceCustomer(sub.customer).run)
```
stephencelis: I'm pretty sure I was seein automatic proration when omitting the flag entirely, but in testing was seeing that it wouldn't invoice when explicit.

Sources/PointFree/Pricing.swift
```diff
@@ -335,7 +335,7 @@ private let pricingIntervalRowView = View<Pricing> { pricing in
               Class.type.normal,
               ])
             ],
-            ["Up to 20% off the Individual Monthly plan"])
+            ["20% off the Individual Monthly plan"])
```
stephencelis: Lil copy change. It's actually more like 22% off :P

mbrandonw: BIG SAVINGS

Sources/PointFree/Account/Change.swift
mbrandonw: So this is a fire and forget invoice? What happens if it fails? Does that matter?

stephencelis: Ah i can do an email

PR #202, opened 2018-04-03
New episode email improvements

mbrandonw:
* Can now send announcements with emails
* Can test emails before sending
* New episode email no longer assumes episode is subscriber-only, and provides new copy for the times we release an episode for free from the beginning.


Sources/PointFree/Admin/Index.swift
```diff
@@ -45,90 +45,3 @@ private let adminIndexView = View<Database.User> { currentUser in
       ])
     ])
 }
-
-let showNewEpisodeEmailMiddleware =
```
mbrandonw: I moved all this stuff to a new file

Sources/PointFree/Admin/NewEpisodes.swift
```diff
@@ -0,0 +1,136 @@
...
+let showNewEpisodeEmailMiddleware =
+  requireAdmin
+    <| writeStatus(.ok)
+    >-> respond(showNewEpisodeView.contramap(lower))
+
+private let showNewEpisodeView = View<Database.User> { _ in
+  ul(
+    AppEnvironment.current.episodes()
+      .sorted(by: their(^\.sequence, >))
+      .prefix(upTo: 1)
```
mbrandonw: im only showing the most recent episode now instead of all of em

Sources/PointFree/Emails/NewEpisodeEmail.swift
```diff
@@ -10,59 +10,94 @@ import Prelude
...
+    )
+  ]
+}
+
 private let nonSubscriberCtaView = View<(Episode, isSubscriber: Bool)> { ep, isSubscriber -> [Node] in
   guard !isSubscriber else { return [] }
 
+  let blurb = ep.subscriberOnly
+    ? "This episode is for subscribers only. To access it, and all past and future episodes, become a subscriber today!"
+    : "This episode is free for everyone, made possible by our subscribers. Consider becoming a subscriber today!"
```
mbrandonw: here's the copy I came up with for when our new episodes are free from the beginning.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -159,8 +159,13 @@ private let routers: [Router<Route>] = [
...
     <¢> get %> lit("admin") %> lit("free-episode-email") <% end,
 
-  .admin <<< .newEpisodeEmail <<< .send
-    <¢> post %> lit("admin") %> lit("new-episode-email") %> pathParam(.int >>> .tagged) <% lit("send") <% end,
+  PartialIso.admin <<< PartialIso.newEpisodeEmail <<< PartialIso.send
+    <¢> post %> lit("admin") %> lit("new-episode-email") %> pathParam(.int >>> .tagged)
+    <%> formField("subscriber_announcement", .string).map(Optional.iso.some)
+    <%> formField("nonsubscriber_announcement", .string).map(Optional.iso.some)
+    <%> isTest
+    <% lit("send")
```
mbrandonw: this route is gross :/

notice that the last `/send` path param goes AFTER the form fields. I couldn't figure out how to parse out that part and discard the `Unit`. maybe we can look at this together today.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -305,3 +310,10 @@ extension PartialIso where A == (String?, Int?), B == Pricing {
     })
   }
 }
+
+private let isTest: Router<Bool?> =
+  formField("live", .string).map(isPresent >>> negate >>> Optional.iso.some)
+    <|> formField("test", .string).map(isPresent >>> Optional.iso.some)
+
+private let isPresent = PartialIso<String, Bool>(apply: { _ in true }, unapply: { $0 ? "" : nil })
+private let negate = PartialIso<Bool, Bool>(apply: { !$0 }, unapply: { !$0 })
```
mbrandonw: and this stuff down here ain't great either, but maybe this is just how it has to be. I wanna parse out the `live` and `test` form fields, and map their presence to either `true` or `false`.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -305,3 +310,10 @@ extension PartialIso where A == (String?, Int?), B == Pricing {
     })
   }
 }
+
+private let isTest: Router<Bool?> =
+  formField("live", .string).map(isPresent >>> negate >>> Optional.iso.some)
+    <|> formField("test", .string).map(isPresent >>> Optional.iso.some)
+
+private let isPresent = PartialIso<String, Bool>(apply: const(true), unapply: { $0 ? "" : nil })
+private let negate = PartialIso<Bool, Bool>(apply: (!), unapply: (!))
```
mbrandonw: and this stuff down here ain't great either, but maybe this is just how it has to be. I wanna parse out the `live` and `test` form fields, and map their presence to either `true` or `false`.

Tests/PointFreeTests/EmailTests/NewEpisodeEmail.swift
```diff
@@ -1,50 +0,0 @@
-import Html
```
mbrandonw: renamed this file to have `Test.swift` as the suffix

Sources/PointFree/Routers/PointFreeRoutes.swift
stephencelis: I thought `end` did that...

Sources/PointFree/Routers/PointFreeRoutes.swift
stephencelis: Do we wanna use a `enum: String` type for this at the least?

Sources/PointFree/Routers/PointFreeRoutes.swift
mbrandonw: I was able to figure this out, just wasn't thinking clearly before.

Sources/PointFree/Routers/PointFreeRoutes.swift
mbrandonw: intersting, but not sure it would work. it's not that a single value is coming back as either `live` or `test`, it's that either a `live` value comes back _or_ a `test` value. it's just cause of how html forms work. I have two submit buttons each with different values. open to other ways of doing this too, it's just want I came up with so that I could have a single form for both the test email and live emails.

PR #210, opened 2018-04-16
Small SQL fixes for owners

stephencelis:
This is mainly a small fix for folks that may have more than one subscription under their belt. Made a few other small fixes along the way.

### DIFF
--- a/PointFree.xcworkspace/contents.xcworkspacedata
+++ b/PointFree.xcworkspace/contents.xcworkspacedata
@@ -2,9 +2,18 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:PointFree.xcodeproj">
+      location = "group:README.md">
+   </FileRef>
+   <FileRef
+      location = "group:LICENSE">
+   </FileRef>
+   <FileRef
+      location = "group:CC-LICENSE.md">
    </FileRef>
    <FileRef
       location = "group:PointFree.playground">
+   </FileRef>
+   <FileRef
+      location = "group:PointFree.xcodeproj">
    </FileRef>
 </Workspace>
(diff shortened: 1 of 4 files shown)


PointFree.xcworkspace/contents.xcworkspacedata
```diff
@@ -2,9 +2,18 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:PointFree.xcodeproj">
+      location = "group:README.md">
```
stephencelis: Adding the `README` and whatnot to the workspace.

image

Sources/PointFree/Database.swift
```diff
@@ -174,7 +174,7 @@ private func update(stripeSubscription: Stripe.Subscription) -> EitherIO<Error,
   return firstRow(
     """
     UPDATE "subscriptions"
-    SET "stripe_subscription_status" = $1
+    SET "stripe_subscription_status" = $1, "updated_at" = NOW()
```
stephencelis: We can remove this if we ever set up some hooks.

PointFree.xcworkspace/contents.xcworkspacedata
mbrandonw: oh great!

pointfreeco.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
stephencelis: Orphan from when the workspace was renamed.

PR #212, opened 2018-04-17
Point-Free Pointers part 3

mbrandonw:
### What

Just chipping away at my checklist, trying to keep these PR's short. This contains:

* Implemented twitter meta data for the index and show page of blog
* Update router to deal with `BlogPost` directly instead of params or ids
* Don't show non-subscriber CTA to subscribers
* Replace large logo SVG that had font vectorized with a SVG that references the external google font


Sources/PointFree/SVG.swift
```diff
@@ -185,24 +185,29 @@ let twitterIconSvgBase64 = base64EncodedString <| """
...
-</g>
-</g>
+<svg width="448px" height="137px" viewBox="0 0 448 137" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <defs>
+        <rect id="path-1" x="0" y="0" width="450" height="155"></rect>
+        <style type="text/css">
+          @import url("https://fonts.googleapis.com/css?family=Permanent+Marker");
```
mbrandonw: this is where we reference the google font

Sources/PointFree/Blog/BlogPostShow.swift
```diff
@@ -113,7 +113,7 @@ let blogPostContentView = View<BlogPost> { post in
       [`class`([Class.pf.type.responsiveTitle3]),],
       [
         a(
-          [href(path(to: .blog(.show(.right(post.id.unwrap)))))],
+          [href(path(to: .blog(.show(post))))],
```
stephencelis: V nice!

PR #213, opened 2018-04-17
Add inline links to exercises

stephencelis:
Easy peasy.


stephencelis:
Would be nice to add explicit linkable links, but for now this'll at least let us link manually.

PR #214, opened 2018-04-17
Point-Free Pointers part 4

mbrandonw:
### What

Ticking more boxes in my checklist. This has:

* adds "Blog" links to nav and footer. To support this _and_ make this code deployable, I have added a very simple feature flag system.
* Render full content to atom feed.
* Adds twitter button to blog posts


Sources/PointFree/Blog/BlogAtomFeed.swift
```diff
@@ -28,6 +28,6 @@ private func atomEntry(for post: BlogPost) -> AtomEntry {
     title: post.title,
     siteUrl: url(to: .blog(.show(post))),
     updated: post.publishedAt,
-    content: [text(post.blurb)]
+    content: blogPostContentView.view(post)
```
mbrandonw: just wanna double check that you are cool with rendering the full blog post content to the atom feed.

Sources/PointFree/FeatureFlags.swift
```diff
@@ -0,0 +1,20 @@
+public struct Feature: Equatable {
```
mbrandonw: a type for feature flags. for now just wanna support things that admins have access to that others do not.

Sources/Styleguide/DesignSystems/Align.swift
```diff
@@ -4,8 +4,10 @@ import Prelude
 extension Class {
   public enum align {
     public static let middle = CssSelector.class("align-middle")
+    public static let top = CssSelector.class("align-top")
   }
 }
 
 public let alignStyles =
   Class.align.middle % verticalAlign(.middle)
+    <> Class.align.top % verticalAlign(.top)
```
mbrandonw: we were missing this from our style guide. this causes looooots of snapshot failures, but not screenshot failures.

mbrandonw:
Oh, and this PR is based off of #212

Sources/PointFree/Components.swift
```diff
@@ -52,3 +52,92 @@ func gitHubLink(text: String, type: GitHubLinkType, redirect: String?) -> Node {
...
+        via.map { URLQueryItem(name: "via", value: $0) }
+        ].compactMap(id)
+      |> ^\.url?.absoluteString
+    )
+    ?? ""
+
+  return a(
+    [
+      href(tweetHref),
+      onclick(unsafeJavascript: """
```
stephencelis: why's this onclick needed?

Sources/PointFree/Blog/BlogAtomFeed.swift
stephencelis: yup!

Sources/PointFree/Components.swift
mbrandonw: This allows us to open the tweet in a new window at a small size. Otherwise it just opens into a tab. 

If js is disabled it’ll still fallback to opening in a tab tho

PR #215, opened 2018-04-18
Point-Free Pointers part 5

mbrandonw:
### What

Adds support for inline images and videos in our transcripts. This works equally for episode transcripts and blog posts!


Sources/PointFree/Blog/BlogPost.swift
```diff
@@ -29,6 +29,11 @@ This is the blurb to a mock blog post. This should just be short and to the poin
 text, no markdown.
 """,
   contentBlocks: [
+    .init(
+      content: "",
+      timestamp: nil,
+      type: .video(poster: "", sources: [])
```
mbrandonw: this is for tests, but wish I had a way to render some mock images in tests. we could do it via `src` data and svgs, but I'll save that for another time.

Sources/PointFree/Episode/Show.swift
```diff
@@ -30,7 +30,7 @@ let episodeResponse =
           currentUser: currentUser,
           data: (permission, currentUser, subscriberState, episode),
           description: episode.blurb,
-          extraStyles: markdownBlockStyles <> pricingExtraStyles <> videoExtraStyles,
```
mbrandonw: I refactored the video styles a bit cause I needed something similar for images.

Sources/PointFree/Episode/Show.swift
```diff
@@ -794,15 +832,21 @@ private enum EpisodePermission: Equatable {
   }
 }
 
-private let pfVideoClass = CssSelector.class("pf-video")
```
mbrandonw: I got rid of these video specific styles that we rendered into the head, and instead was able to build everything off of our style guide.

Sources/PointFree/Episode/Show.swift
```diff
@@ -794,15 +832,21 @@ private enum EpisodePermission: Equatable {
...
-    width(.pct(100))
-      <> padding(bottom: .pct(56.25))
-      <> position(.relative)
-    )
-    <> (
-      (pfVideoClass > video) % (
-        position(.absolute) <> height(.pct(100))
-      )
-)
+let outerVideoContainerClass: CssSelector =
```
mbrandonw: we just have to apply these classes to the video elements, but the outer and inner containers.

mbrandonw:
this PR is based off of #214

PR #219, opened 2018-04-18
Fix word breaking 

mbrandonw:
The latest episode (https://www.pointfree.co/episodes/ep12-tagged) gets some horizontal scrolling due to a really long word in a pull quote. This PR fixes that by adding `word-wrap: break-word` to all `p` tags in our rendered markdown.


Sources/PointFree/Episode/Show.swift
```diff
@@ -723,7 +723,8 @@ private func episode(forParam param: Either<String, Int>) -> Episode? {
 private let markdownContainerClass = CssSelector.class("md-ctn")
 let markdownBlockStyles: Stylesheet =
   markdownContainerClass % (
-    a % key("text-decoration", "underline")
+    p % key("word-wrap", "break-word")
```
mbrandonw: this is the fix. we dont have these properties in the `Css` lib yet.

Sources/PointFreeTestSupport/MockEpisode.swift
```diff
@@ -100,9 +100,14 @@ private let paidEpisode = Episode(
     Episode.TranscriptBlock(
       content: """
       That was another title. See how the title create the “table of contents” under the video?
+
+      Here's a block quote with a really long word inside so that we can make sure that word breaking is
+      working how we expect:
+
+      > Fatal error: ‘try!’ expression unexpectedly raised an error: Swift.DecodingError.typeMismatch(Swift.Dictionary<Swift.Dictionary<Swift.StringSwift.Dictionary<Swift.String, Any>>>, Swift.DecodingError.Context(codingPath: [_JSONKey(stringValue: “Index 0”, intValue: 0), CodingKeys(stringValue: “email”, intValue: nil)], debugDescription: “Expected to decode Dictionary<String, Any> but found a string/data instead.”, underlyingError: nil))
```
mbrandonw: a test that showed off the problem

PR #222, opened 2018-04-19
Subscribe CTA on homepage

mbrandonw:
### What

This inserts a subscribe CTA between the 3rd and 4th episodes on the homepage. Down for radically different forms of copy, this is just what I came up with off the top of my head.


PointFree.playground/Pages/Homepage.xcplaygroundpage/Contents.swift
```diff
@@ -13,6 +13,7 @@ import Optics
 import SnapshotTesting
 
 AppEnvironment.push(const(.mock))
+AppEnvironment.push(\.episodes .~ unzurry(allPublicEpisodes))
```
mbrandonw: nice to have some public episodes to play with in playgrounds.

Sources/PointFree/Blog/BlogPostShow.swift
```diff
@@ -8,7 +8,7 @@ import Prelude
 import Styleguide
 import Tuple
 
-let blogPostShowMiddleware: (Conn<StatusLineOpen, Tuple4<Either<String, Int>, Database.User?, SubscriberState, Route?>>) -> IO<Conn<ResponseEnded, Data>> =
```
mbrandonw: I got a small boost in compilation by moving the types down to `layoutData` instead of here, so just playing with that.

Tests/PointFreeTests/HomeTests.swift
```diff
@@ -11,19 +11,86 @@ import WebKit
...
+    #if !os(Linux)
+    if #available(OSX 10.13, *), ProcessInfo.processInfo.environment["CIRCLECI"] == nil {
+      let webView = WKWebView(frame: .init(x: 0, y: 0, width: 1080, height: 2300))
+      webView.loadHTMLString(String(data: result.perform().data, encoding: .utf8)!, baseURL: nil)
+      assertSnapshot(matching: webView, named: "desktop")
+
+      webView.frame.size.width = 400
+      webView.frame.size.height = 2800
+
+      let render = XCTestExpectation()
```
mbrandonw: it seems that when making the web view _taller_ it needs an extra moment to re-render. without this wait there's just blank white space in the extra area at the bottom. it's strange that we don't need to do this when resizing the window.

do you think we should make a test helper to make this easy for times we need one extra runloop cycle to get screenshots to pass?

Tests/PointFreeTests/HomeTests.swift
```diff
@@ -11,19 +11,86 @@ import WebKit
 #endif
 
 class HomeTests: TestCase {
-  func testHomepage() {
+  override func setUp() {
+    super.setUp()
+
+    let eps = [
```
mbrandonw: I beefed up the homepage tests to use our public episodes

Tests/PointFreeTests/HomeTests.swift
```diff
@@ -11,19 +11,86 @@ import WebKit
...
+      ep10,
+      ep2,
+      ep1,
+      introduction,
+      ]
+      .suffix(4)
+      .map(\.image .~ "")
```
mbrandonw: so easy to clear out all the images on an episode so that we dont do URL requests.

Sources/PointFree/Blog/BlogPostShow.swift
stephencelis: huh

mbrandonw:
Images in the diff aren't being expanded, but here is a screenshot to show what it looks like for non-subscribers:

Desktop: https://github.com/pointfreeco/pointfreeco/blob/7ffe6fb34096f84ca6e3e4bd4660edf2fdeb6fa8/Tests/PointFreeTests/__Snapshots__/HomeTests/testHomepage_LoggedOut.2.desktop.png
Mobile: https://github.com/pointfreeco/pointfreeco/blob/7ffe6fb34096f84ca6e3e4bd4660edf2fdeb6fa8/Tests/PointFreeTests/__Snapshots__/HomeTests/testHomepage_LoggedOut.3.mobile.png

Tests/PointFreeTests/HomeTests.swift
```diff
@@ -11,19 +11,86 @@ import WebKit
...
       webView.frame.size.width = 400
-      assertSnapshot(matching: webView, named: "mobile")
+      webView.frame.size.height = 3500
+
+      let render = XCTestExpectation()
+      DispatchQueue.main.async {
+        assertSnapshot(matching: webView, named: "mobile")
+        render.fulfill()
+      }
+      guard XCTWaiter.wait(for: [render], timeout: 2.0) == .completed else {
```
stephencelis: I think you can do `waitForExpectations` and save a coupla lines.

Tests/PointFreeTests/HomeTests.swift
stephencelis: oh huh. yeah that's not a bad idea. prob could generalize using GCD to flatten things with a semaphore?

Tests/PointFreeTests/SiteMiddlewareTests.swift
```diff
@@ -16,6 +16,16 @@ private func secureRequest(_ urlString: String) -> URLRequest {
 }
 
 class SiteMiddlewareTests: TestCase {
+  override func setUp() {
+    super.setUp()
+    AppEnvironment.push(\.database .~ .mock)
```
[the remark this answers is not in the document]
stephencelis: yeah good pt!

Tests/PointFreeTests/HomeTests.swift
mbrandonw: nice, will do

PR #223, opened 2018-04-19
Add Payment History

stephencelis:
This'll make it easy for folks to grab historic invoices and add VAT numbers when they need to.


Sources/PointFree/PageLayout.swift
```diff
@@ -60,6 +60,16 @@ enum NavStyle {
 }
 
 struct SimplePageLayoutData<A> {
+  enum Style {
```
stephencelis: Promoted `NavStyle` to `Style` so we can render pages without the footer (_e.g._, the printable invoice).

Sources/PointFree/Stripe.swift
```diff
@@ -112,29 +127,59 @@ public struct Stripe {
...
+    public private(set) var number: Number
+    public private(set) var periodStart: Date
+    public private(set) var periodEnd: Date
+    public private(set) var subscription: Subscription.Id?
+    public private(set) var subtotal: Cents
+    public private(set) var total: Cents
 
-    public typealias Id = Tagged<Invoice, String>
+    public typealias Id = Tagged<(Invoice, id: ()), String>
+    public typealias Number = Tagged<(Invoice, number: ()), String>
```
stephencelis: Turns out we don't need to create one-off types for tag collisions! These two types aren't considered equal! 😄

Sources/PointFree/StripeHook.swift
```diff
@@ -10,6 +10,10 @@ import Styleguide
 
 let stripeInvoiceWebhookMiddleware =
   validateStripeSignature
+    <<< filterMap(
+      ^\Stripe.Event<Stripe.Invoice>.data.object.subscription >>> pure,
```
stephencelis: Turns out that the root `subscription` on our invoices is optional! Even though we only deal with subscriptions, invoices that deal with proration push this info down into the line items 🤷‍♂️

PointFree.playground/Pages/Invoices.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,31 @@
...
+import Prelude
+import WebKit
+import Optics
+import SnapshotTesting
+import Styleguide
+
+let page = simplePageLayout(invoicesView).view(
+  .init(
+    currentUser: .mock,
+    data: (.mock, .mock([.mock, .mock, .mock]), .mock),
```
mbrandonw: lol! type inference is great!

Sources/PointFree/Account/Account.swift
```diff
@@ -630,11 +630,14 @@ private let subscriptionPaymentInfoView = View<Stripe.Subscription> { subscripti
...
+  return currencyFormatter.string(from: dollars)
+    ?? NumberFormatter.localizedString(from: dollars, number: .currency)
+}
+
 private func totalAmount(for subscription: Stripe.Subscription) -> String {
-  let totalCents = subscription.plan.amount.rawValue * subscription.quantity
-  let totalDollars = NSNumber(value: Double(totalCents) / 100)
-  return currencyFormatter.string(from: totalDollars)
-    ?? NumberFormatter.localizedString(from: totalDollars, number: .currency)
+  return format(cents: .init(unwrap: subscription.plan.amount.rawValue * subscription.quantity))
```
mbrandonw: would `Tagged: Numeric` make this simpler?

Sources/PointFree/Account/Invoices.swift
```diff
@@ -0,0 +1,302 @@
...
+          title: "Payment history"
+        )
+    }
+)
+
+let invoiceResponse =
+  filterMap(require1 >>> pure, or: loginAndRedirect)
+    <<< requireStripeSubscription
+    <<< filterMap(
+      over3(fetchInvoice) >>> sequence3 >>> map(require3),
```
mbrandonw: do we need an additional filter to make sure that the invoice belongs to the user? it seems that any invoice id could be provided and we'd just fetch it?

Sources/PointFree/Account/Account.swift
stephencelis: Yup. I can maybe get Web and Point-Free usin Tagged after mergin this.

Sources/PointFree/Account/Invoices.swift
```diff
@@ -0,0 +1,302 @@
...
+  -> Middleware<StatusLineOpen, ResponseEnded, T2<Stripe.Subscription, A>, Data> {
+
+    return { conn in
+      let subscription = conn.data.first
+
+      return AppEnvironment.current.stripe.fetchInvoices(subscription.customer)
+        .withExcept(notifyError(subject: "Couldn't load invoices"))
+        .run
+        .flatMap {
+          switch $0 {
```
mbrandonw: I've been a big fan of the `.flatMap(either({}, {}))` style you pioneered.

Sources/PointFree/Account/Invoices.swift
```diff
@@ -0,0 +1,302 @@
...
+      )
+      ])
+    ])
+}
+
+private let titleRowView = View<Prelude.Unit> { _ in
+  gridRow([`class`([Class.padding([.mobile: [.bottom: 2]])])], [
+    gridColumn(sizes: [.mobile: 12], [
+      div([
+        h1([`class`([Class.pf.type.title2])], ["Payment history"])
```
mbrandonw: you should use `responseTitle2` so that it gets smaller on mobile.

I should deprecate all the `titleX` classes.

Sources/PointFree/Account/Invoices.swift
```diff
@@ -0,0 +1,302 @@
...
+              ["Print"]
+            )
+            ])
+          ]),
+        ])
+    }
+  )
+}
+
+let invoiceView = View<(Stripe.Subscription, Database.User, Stripe.Invoice)> { subscription, currentUser, invoice -> Node in
```
mbrandonw: totally not necessary, but I could see this view being broken down into some smaller subviews. it's p big!

we could save it for another day.

Sources/PointFree/Stripe.swift
mbrandonw: lol amazing!

Sources/PointFree/Stripe.swift
```diff
@@ -343,27 +396,25 @@ private enum Method {
...
 }
 
+private func attachMethod(_ method: Method) -> (URLRequest) -> URLRequest {
+  switch method {
+  case .get:
+    return \.httpMethod .~ "GET"
+  case let .post(params):
+    return (\.httpMethod .~ "POST")
+      <> setHeader("Idempotency-Key", UUID().uuidString)
+      <> attachFormData(params)
```
mbrandonw: fun!

Sources/PointFree/TODO.swift
```diff
@@ -286,3 +292,25 @@ public func require4<A, B, C, D, Z>(_ x: T5<A, B, C, D?, Z>) -> T5<A, B, C, D, Z
 public func require5<A, B, C, D, E, Z>(_ x: T6<A, B, C, D, E?, Z>) -> T6<A, B, C, D, E, Z>? {
   return get5(x).map { over5(const($0)) <| x }
 }
+
+// PreludeFoundation
```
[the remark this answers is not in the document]
mbrandonw: nice

Sources/PointFree/__Generated__/DerivedPartialIsos.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 0.10.1 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.11.0 — https://github.com/krzysztofzablocki/Sourcery
```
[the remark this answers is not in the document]
mbrandonw: should we vendor sourcery in `bin/`?

Sources/Styleguide/DesignSystems/Typography.swift
```diff
@@ -26,7 +26,11 @@ extension Class {
     public static let semiBold = CssSelector.class("semi-bold")
     public static let underline = CssSelector.class("underline")
 
-    public static let fontFamilyInherit = CssSelector.class("font-family-inherit")
+    public enum fontFamily {
```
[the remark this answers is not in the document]
mbrandonw: nice

Sources/PointFree/Account/Invoices.swift
stephencelis: Ah yeah. I originally thought that the customer was needed to form the request and forgot to come back to this! Good catch!

Sources/PointFree/Account/Invoices.swift
stephencelis: Ah coo. I copy-pasted this from another screen.

PR #224, opened 2018-04-19
Remove `.titleX` classes, and fix pointers logo

mbrandonw:
I removed all the `.title1`, `.title2`, etc classes in favor of their responsive cousins.

I also fixed the "Pointers" font problem in the SVG. it inflates the pages by 55kb, so maybe we should just move to cloudfront?


Sources/PointFree/Account/Change.swift
```diff
@@ -322,7 +322,7 @@ private let individualPricingColumnView = View<(Pricing.Billing, Pricing)> { bil
             ]),
           ]),
         gridColumn(sizes: [.mobile: 12], [], [
-          h6([`class`([Class.pf.type.title6, Class.pf.colors.fg.gray650, Class.display.inline])], [
+          h6([`class`([Class.pf.type.responsiveTitle7, Class.pf.colors.fg.gray650, Class.display.inline])], [
```
mbrandonw: screenshots saved my butt here! note that `title6` in the old system was the small ALL CAPS style, but in the new responsible system it's `responsiveTitle7`. I didn't realize that until I saw the different in the screenshots.

Sources/PointFree/Footer.swift
```diff
@@ -144,7 +144,7 @@ private let footerLinkClass =
     | Class.pf.type.body.regular
 
 private let columnTitleClass =
-  Class.pf.type.title6
+  Class.pf.type.responsiveTitle7
```
mbrandonw: same here

Sources/PointFree/Pricing.swift
```diff
@@ -376,7 +376,7 @@ return
             ]),
           ]),
         gridColumn(sizes: [.mobile: 12], [], [
-          h6([`class`([Class.pf.type.title6, Class.pf.colors.fg.gray650, Class.display.inline])], [
+          h6([`class`([Class.pf.type.responsiveTitle7, Class.pf.colors.fg.gray650, Class.display.inline])], [
```
mbrandonw: and here

Tests/PointFreeTests/__Snapshots__/BlogTests/testBlogIndex.1.Conn.txt
```diff
@@ -10,10 +10,10 @@
 
 ▿ Response
   Status 200 OK
-  Content-Length: 45010
+  Content-Length: 100724
```
mbrandonw: you can see just how much the blog index page is increasing here.

p cool that we have a history of the file size of all our pages.

PR #225, opened 2018-04-19
Update Tagged

stephencelis:
Let's use our extracted lib!


Sources/PointFree/Account/Account.swift
```diff
@@ -644,7 +644,7 @@ public func format(cents: Stripe.Cents) -> String {
 }
 
 private func totalAmount(for subscription: Stripe.Subscription) -> String {
-  return format(cents: .init(unwrap: subscription.plan.amount.rawValue * subscription.quantity))
+  return format(cents: subscription.plan.amount * .init(rawValue: subscription.quantity))
```
stephencelis: Cleans it up...slightly.

Sources/PointFree/Account/Invite.swift
```diff
@@ -57,7 +57,7 @@ let resendInviteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple2<Dat
       return conn
         |> redirect(
           to: .account(.index),
-          headersMiddleware: flash(.notice, "Invite sent to \(invite.email.unwrap).")
+          headersMiddleware: flash(.notice, "Invite sent to \(invite.email).")
```
stephencelis: The new `Tagged` is `CustomStringRepresentable`.

Sources/PointFree/Account/Invite.swift
```diff
@@ -79,7 +79,7 @@ let acceptInviteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple2<Dat
             errorOrInviter.right.map { inviter in
               sendEmail(
                 to: [inviter.email],
-                subject: "\(currentUser.name ?? currentUser.email.unwrap) has accepted your Point-Free team invitation!",
+                subject: "\(currentUser.displayName) has accepted your Point-Free team invitation!",
```
stephencelis: We're doin this a lot so I decided to extract it out.

PR #227, opened 2018-04-19
Make cookies last longer

### LINKED ISSUE #218: Use longer lasting cookie
Currently when you quit your browser, the login session is not persisted. It would be cool if it was persisted for some period of time to avoid having to re-login frequently.

stephencelis:
Fixes #218.


Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -24,7 +24,7 @@ extension Environment {
   public static let teamYearly = mock
     |> \.database.fetchSubscriptionTeammatesByOwnerId .~ const(pure([.mock]))
     |> \.database.fetchTeamInvites .~ const(pure([.mock]))
-    |> \.stripe.fetchSubscription .~ const(pure(.teamYearly))
+    |> (\Environment.stripe.fetchSubscription) .~ const(pure(.teamYearly))
```
stephencelis: Heh, this snuck in because it sped up compile time here quite a bit.

stephencelis:
Tested locally and my session now persists across browser sessions.

stephencelis:
Happy now @mergesort?

Sources/PointFree/Session.swift
```diff
@@ -14,7 +14,10 @@ public func writeSessionCookieMiddleware<A>(_ update: @escaping (Session) -> Ses
       return setCookie(
           key: pointFreeUserSession,
           value: value,
-          options: [.path("/")]
+          options: [
+            .expires(AppEnvironment.current.date().addingTimeInterval(60 * 60 * 24 * 365 * 10)),
```
mbrandonw: would we even do `Date.distantFuture`?

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
mbrandonw: ha I bet

Sources/PointFree/Session.swift
stephencelis: Ha, that's the year 4,000. A little optimistic, are we?

My only worry is some browsers handling this stuff weird and breaking cookies for some folks.

mergesort:
>Happy now @mergesort?

This seems a bit overly philosophical to answer in a PR.

PR #229, opened 2018-04-20
Unify bootstrap logic and introduce Runner

stephencelis:
Just a stub executable for now. But wanted to push this plumbing first.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "Sources/Server/Transcripts"]
-	path = Sources/Server/Transcripts
+	path = Sources/PointFree/Transcripts
 	url = http://github.com/mbrandonw/episode-transcripts
--- a/Package.swift
+++ b/Package.swift
@@ -1,15 +1,15 @@
 // swift-tools-version:4.0
 
-import Foundation
 import PackageDescription
 
 let package = Package(
   name: "PointFree",
   products: [
-    .executable(name: "Server", targets: ["Server"]),
-    .library(name: "Styleguide", targets: ["Styleguide"]),
     .library(name: "PointFree", targets: ["PointFree"]),
     .library(name: "PointFreeTestSupport", targets: ["PointFreeTestSupport"]),
+    .executable(name: "Runner", targets: ["Runner"]),
+    .executable(name: "Server", targets: ["Server"]),
+    .library(name: "Styleguide", targets: ["Styleguide"]),
     ],
   dependencies: [
     .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("b811904")),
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "Sources/Server/Transcripts"]
-	path = Sources/Server/Transcripts
+	path = Sources/PointFree/Transcripts
```
stephencelis: This stuff can live in the `PointFree` module now, which means our bootstrap script can in full, as well.

Package.swift
```diff
@@ -1,15 +1,15 @@
 // swift-tools-version:4.0
 
-import Foundation
```
stephencelis: Most of the noise in this file is cleanup.

Package.swift
```diff
@@ -1,15 +1,15 @@
...
 import PackageDescription
 
 let package = Package(
   name: "PointFree",
   products: [
-    .executable(name: "Server", targets: ["Server"]),
-    .library(name: "Styleguide", targets: ["Styleguide"]),
     .library(name: "PointFree", targets: ["PointFree"]),
     .library(name: "PointFreeTestSupport", targets: ["PointFreeTestSupport"]),
+    .executable(name: "Runner", targets: ["Runner"]),
```
stephencelis: Except for this.

``` sh
$ swift run Runner
```

Sources/PointFree/Bootstrap.swift
```diff
@@ -1,9 +1,11 @@
 import Either
+import Foundation
+import Optics
 import Prelude
 
 public func bootstrap() -> EitherIO<Error, Prelude.Unit> {
-
   return print(message: "⚠️ Bootstrapping PointFree...")
+    .flatMap(const(loadEnvVars))
```
stephencelis: We now load the environment in here!

Sources/Runner/main.swift
```diff
@@ -0,0 +1,9 @@
+import PointFree
```
stephencelis: Runner is so far just a "test" for if the app can boot.

Sources/Server/main.swift
```diff
@@ -1,57 +1,7 @@
-import Foundation
```
stephencelis: `Server`'s main file got reeeal short.

Sources/PointFree/Bootstrap.swift
mbrandonw: sooooooo much better

Sources/Runner/main.swift
mbrandonw: just so I understand, in the future we'd put our "cron" logic in here, like parse arguments passed in from the scheduler to figure out what logic to execute, etc...?

stephencelis: Yeah, we can start with a single task that executes, but we'll eventually wanna use a CLI argument parser. Will be fun to play around with an applicative argument printer/parser.

PR #230, opened 2018-04-20
Convert 100% widths to max widths

mbrandonw:
I was using `width="100%"` HTML attributes and turns out that isn't valid html. Oops! Easy enough to fix, and not a single screenshot failure!


Sources/PointFree/Emails/EmailLayouts.swift
```diff
@@ -48,7 +48,7 @@ func simpleEmailLayout<A>(_ bodyView: View<A>) -> View<SimpleEmailLayoutData<A>>
         body([bgcolor("#FFFFFF")], [
           span([style(preheaderStyles)], [text(layoutData.preheader)]),
 
-          emailTable([height(.pct(100)), width(.pct(100)), style(bodyTableStyles)], [
+          emailTable([height(.pct(100)), style(bodyTableStyles)], [
```
mbrandonw: wasn't even needed here!

Sources/PointFree/MountainNavView.swift
```diff
@@ -15,7 +15,7 @@ let mountainNavView = View<(NavStyle.MountainsStyle, Database.User?, SubscriberS
       gridRow([`class`([Class.grid.top(.mobile), Class.grid.between(.mobile), Class.padding([.mobile: [.top: 3], .desktop: [.top: 0]])])], [
 
         gridColumn(sizes: [.mobile: 5], [`class`([Class.padding([.mobile: [.top: 4], .desktop: [.top: 0]])]), style(lineHeight(0))], [
-          img(base64: heroMountainSvgBase64, mediaType: .image(.svg), alt: "", [width(.pct(100))])
+          img(base64: heroMountainSvgBase64, mediaType: .image(.svg), alt: "", [`class`([Class.size.width100pct])])
```
mbrandonw: we should use the `max-width: 100%` css instead of `width="100%"` html.

Issue #232, opened 2018-04-21
Bootstrap-oss error


eliotfowler:
I am trying to get the pointfree code running locally and I ran into an issue while trying to boostrap.

```
❯ make bootstrap-oss
test -f .env \
		|| cp .env.example .env
test -d Sources/Server/Transcripts/ \
		|| cp -r Transcripts.example/ Sources/Server/Transcripts/
cp: directory Sources/Server/Transcripts does not exist
make: *** [mock-transcripts] Error 1
```

I see that the git submodule for episode-transcripts lives inside Sources/Pointfree, but since that's done in bootstrap-common, it happens after the mock-transcripts task attempts to copy. I tried running `git submodule update --init --recursive` which is when I realized that it appears the [episode-transcripts](https://github.com/mbrandonw/episode-transcripts/tree/6260c6bcaf7e3a959bad435305888e92a156d762) repo is no longer public (or no longer exists). 

It's very likely I messed something up on my end. It's not urgent but I'd really appreciate a push in the right direction!

mbrandonw:
Looks like in some of our most recent work to clean things up (#229) we forgot to update the bootstrap! We'll have a fix shortly...

garricn:
I was going to report this last night but I thought maybe I was just doing something wrong! I saw Steven’s commit but still just thought I was missing something 🙈

mbrandonw:
This should be fixed by https://github.com/pointfreeco/pointfreeco/pull/233 !

Gotta step out for a bit, but when y'all get a moment give it a shot!

eliotfowler:
Confirmed fixed for me!

PR #233, opened 2018-04-21
Fix transcript references

mbrandonw:
A few more references to fix after the #229 shuffle.

Should help with #232

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,10 @@ jobs:
           command: brew install sourcery
 
       - run:
+          name: Run Bootstrap OSS
+          command: make bootstrap-oss
+
+      - run:
           name: Run Swift tests
           command: make test-swift
           environment:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -28,6 +28,10 @@ jobs:
           name: Install Sourcery
           command: brew install sourcery
 
+      - run:
+          name: Run Bootstrap OSS
+          command: make bootstrap-oss
```
mbrandonw: let's see how circle CI does with this. if it chokes I'll split it into multiple steps that skips the module map stuff.

Makefile
```diff
@@ -13,18 +13,14 @@ xcodeproj-oss:
...
 
 bootstrap: submodules bootstrap-common xcodeproj
 
 mock-env:
 	test -f .env \
 		|| cp .env.example .env
 
-mock-transcripts:
-	test -d Sources/Server/Transcripts/ \
-		|| cp -r Transcripts.example/ Sources/Server/Transcripts/
```
mbrandonw: now that we have public episodes in this repo we don't even need mock transcripts. p cool

PR #235, opened 2018-04-22
📄 README Tweaks

stephencelis:
More README tweaks!

- Better badges
- Better clarity on running Postgres
- Emojis


stephencelis:
Rich diff: https://github.com/pointfreeco/pointfreeco/blob/badges/README.md

PR #237, opened 2018-04-23
Point-Free Pointers launch

mbrandonw:
Everything for the launch...


mbrandonw:
merging cause it's late but I think everything is legit!

PR #239, opened 2018-04-28
AppEnvironment Exploration

stephencelis:
Here's an attempt at simplifying `AppEnvironment`/`Environment` given how we typically use it. I've explored some other naming, but let's bike-shed away.

### DIFF
--- a/PointFree.playground/Pages/About.xcplaygroundpage/Contents.swift
+++ b/PointFree.playground/Pages/About.xcplaygroundpage/Contents.swift
@@ -12,7 +12,7 @@ import WebKit
 import Optics
 import Styleguide
 
-AppEnvironment.push(const(.mock))
+Current = .mock
 
 let request = PointFreeTestSupport.request(to: .pricing(nil, expand: nil))
 
--- a/Sources/PointFree/Environment.swift
+++ b/Sources/PointFree/Environment.swift
@@ -3,10 +3,19 @@ import Foundation
 import Optics
 import Prelude
 
+public struct Assets {
+  public var brandonImgSrc = "https://d3rccdn33rt8ze.cloudfront.net/about-us/brando.jpg"
+  public var stephenImgSrc = "https://d3rccdn33rt8ze.cloudfront.net/about-us/stephen.jpg"
+  public var emailHeaderImgSrc = "https://d3rccdn33rt8ze.cloudfront.net/email-assets/pf-email-header.png"
+  public var pointersEmailHeaderImgSrc = "https://d3rccdn33rt8ze.cloudfront.net/email-assets/pf-pointers-header.jpg"
+}
+
 public enum CookieTransform: String, Codable {
   case plaintext
   case encrypted
 }
+
+public var Current = Environment()
 
 public struct Environment {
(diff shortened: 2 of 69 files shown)


PointFree.playground/Pages/About.xcplaygroundpage/Contents.swift
```diff
@@ -12,7 +12,7 @@ import WebKit
 import Optics
 import Styleguide
 
-AppEnvironment.push(const(.mock))
+Current = .mock
```
stephencelis: Replacing the environment is now pretty explicit mutation.

Sources/PointFree/Environment.swift
```diff
@@ -50,29 +59,3 @@ public struct Environment {
     self.stripe = stripe
   }
 }
-
-public struct AppEnvironment {
```
stephencelis: `AppEnvironment` and its static functions are replaced with `Current` and more general setter functions.

stephencelis:
``` swift
AppEnvironment.current.database
// becomes
Current.database

AppEnvironment.push(.mock)
// becomes
Current = .mock

AppEnvironment.pop()
// is gone

AppEnvironment.with((\.database .~ mock) <> (\.stripe .~ mock)) {
  // ...
}
// becomes
update(&Current, \.database .~ .mock, \.stripe .~ mock)
// or even could become
(Current.database, Current.stripe) = (.mock, .mock)
```

Tests/PointFreeTests/AuthTests.swift
```diff
@@ -17,30 +17,30 @@ class AuthTests: TestCase {
...
-      XCTAssertEqual(gitHubUserEnvelope.accessToken.accessToken, registeredUser.gitHubAccessToken)
-      XCTAssertEqual(gitHubUserEnvelope.gitHubUser.id, registeredUser.gitHubUserId)
-      XCTAssertEqual(gitHubUserEnvelope.gitHubUser.name, registeredUser.name)
-      XCTAssertEqual(1, registeredUser.episodeCreditCount)
-    }
+    update(
+      &Current,
+      \.database .~ .live,
+      \.gitHub.fetchUser .~ const(pure(gitHubUserEnvelope.gitHubUser)),
+      \.gitHub.fetchAuthToken .~ const(pure(pure(gitHubUserEnvelope.accessToken)))
```
mbrandonw: I have you found that `update(&Current, ...)` type checks better than how we were doing `env` previously?

Sources/PointFree/TODO.swift
```diff
@@ -314,3 +314,37 @@ let attachFormData =
     >>> ^\.utf8
     >>> Data.init
     >>> set(\URLRequest.httpBody)
+
+// Prelude
+
+public func concat<A>(_ fs: [(A) -> A]) -> (A) -> A {
```
mbrandonw: is concat not apart of overture too?

Tests/PointFreeTests/AuthTests.swift
stephencelis: It's about the same :/

PR #240, opened 2018-04-29
Temporarily Public Episodes

stephencelis:
This'll let us schedule subscriber-only episodes to be temporarily public.

### DIFF
--- a/Sources/PointFree/Bootstrap.swift
+++ b/Sources/PointFree/Bootstrap.swift
@@ -54,6 +54,9 @@ private let loadEnvVars =
       #else
       let allEpisodes = allPublicEpisodes + allPrivateEpisodes
       #endif
+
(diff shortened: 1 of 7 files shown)


Sources/PointFree/Bootstrap.swift
```diff
@@ -54,6 +54,9 @@ private let loadEnvVars =
       #else
       let allEpisodes = allPublicEpisodes + allPrivateEpisodes
       #endif
+
+      assert(allEpisodes.count == Set(allEpisodes.map(^\.id)).count)
+      assert(allEpisodes.count == Set(allEpisodes.map(^\.sequence)).count)
```
stephencelis: Added a good ol' runtime error! This should help us avoid the email glitch from last time.

stephencelis:
I didn't add an automatic banner because it seems like we'll want that to be personal.

PR #242, opened 2018-05-05
Pfp 0003

mbrandonw:
For monday


Sources/PointFree/Episode/Show.swift
```diff
@@ -763,6 +763,7 @@ let markdownBlockStyles: Stylesheet =
   markdownContainerClass % (
     hrMarkdownStyles
       <> aMarkdownStyles
+      <> blockquote % fontStyle(.italic)
```
mbrandonw: im using block quotes for the questions, so was nice to give it a bit different style.

PR #244, opened 2018-05-10
Upgrade normalize to 8.0.0

mbrandonw:
Normalize 8.0.0 is a bit smaller than the version we currently use, _and_ it fixes some of the bugs we've seen with input buttons being styled a bit differently than link buttons.

Main reason I'm doing this is cause it could be a fun PFP to write about...


Tests/PointFreeTests/AccountTests/__Snapshots__/AccountTests/testAccount.1.Conn.txt
```diff
@@ -9,7 +9,7 @@
 
 ▿ Response
   Status 200 OK
-  Content-Length: 38244
+  Content-Length: 37719
```
mbrandonw: Saving a whole 525 bytes! The site is gonna screeeeaaaammmmmmm

PR #246, opened 2018-05-12
Resubscribing Fixes

stephencelis:
We were making a few small bad assumptions here that prevented folks from re-subscribing after cancellation!


Sources/PointFree/Account/Account.swift
```diff
@@ -451,7 +451,7 @@ private let subscriptionPlanRows = View<Stripe.Subscription> { subscription in
 }
 
 private func mainAction(for subscription: Stripe.Subscription) -> Node {
-  if subscription.cancelAtPeriodEnd {
```
stephencelis: This value remains `true` even after cancellation.

Sources/PointFree/Pricing.swift
```diff
@@ -584,8 +584,15 @@ func redirectActiveSubscribers<A>(
           )
           ?? throwE(unit)
 
-        return (userSubscription.run.parallel <|> ownerSubscription.run.parallel)
-          .sequential
+        let race = (userSubscription.run.parallel <|> ownerSubscription.run.parallel).sequential
+
+        return EitherIO(run: race)
+          .flatMap {
+            $0.stripeSubscriptionStatus == .active
```
stephencelis: @mbrandonw Maybe I am misreading, but looks like the pricing page redirect logic was merely checking for a subscription at all, not whether or not it was active.

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -366,7 +369,7 @@ extension Stripe.Subscription {
   public static let canceling = mock
     |> \.cancelAtPeriodEnd .~ true
 
-  public static let canceled = mock
+  public static let canceled = canceling
```
stephencelis: This immediately made the account page render test for canceling subscriptions fail 😄

Tests/PointFreeTests/PricingTests.swift
```diff
@@ -76,4 +76,17 @@ class PricingTests: TestCase {
 
     assertSnapshot(matching: result.perform())
   }
+
+  func testPricingLoggedIn_CanceledSubscriber() {
```
stephencelis: Quick test for pricing redirect logic.

stephencelis:
I've deployed this branch, but will merge before monday if you don't have time to check it out.

Sources/PointFree/Pricing.swift
mbrandonw: On the road so hard to know for sure, but I believe I did that thinking the user would fix their subscription from the account page. 

Guess I’m a lil murky on when a user should go through the pricing page again

stephencelis: Gotcha. OK I think there's more nuance here to handle `pastDue` accounts! Thanks!

stephencelis:
Gonna merge now to make it easier to work other branches off these fixes.

PR #248, opened 2018-05-18
🍤 Fix Stripe Webhooks

stephencelis:
The fix is fun! The fact we had this bug is not.


Sources/PointFree/Account/Account.swift
```diff
@@ -596,7 +596,7 @@ private let subscriptionInviteMoreRowView = View<(Stripe.Subscription?, [Databas
 }
 
 private let subscriptionPaymentInfoView = View<Stripe.Subscription> { subscription -> [Node] in
-  guard let card = subscription.customer.sources.data.first else { return [] }
+  guard let card = subscription.customer.right?.sources.data.first else { return [] }
```
stephencelis: I've updated all `expand`able properties to be `Either<T.Id, T>`.

Sources/PointFree/Account/Change.swift
```diff
@@ -68,7 +68,7 @@ private func subscriptionChange(_ conn: Conn<StatusLineOpen, (Stripe.Subscriptio
       .flatMap { sub -> EitherIO<Error, Stripe.Subscription> in
         if shouldInvoice {
           parallel(
-            Current.stripe.invoiceCustomer(sub.customer)
+            Current.stripe.invoiceCustomer(sub.customer.either(id, ^\.id))
```
stephencelis: It's a nice simple fold to returns an id.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -201,8 +201,8 @@ private func render(conn: Conn<StatusLineOpen, T3<Database.Subscription?, Databa
 
     case .webhooks(.stripe(.fallthrough)):
       return conn
-        |> writeStatus(.ok)
-        >-> end
+        |> writeStatus(.internalServerError)
+        >-> respond(text: "We don't support this event.")
```
stephencelis: Decoding errors fall through here. Now we'll be more aware of failures.

Tests/PointFreeTests/StripeHookTests.swift
```diff
@@ -1,83 +0,0 @@
-import Either
```
stephencelis: This file moved to `StripeWebhooksTest.swift`. I always try to quick open typing "web" enough to rename.

Tests/PointFreeTests/StripeWebhooksTests.swift
```diff
@@ -0,0 +1,186 @@
...
+import WebKit
+#endif
+
+final class StripeWebhooksTests: TestCase {
+  override func setUp() {
+    super.setUp()
+    update(&Current, \.database .~ .mock)
+  }
+
+  func testDecoding() throws {
```
stephencelis: New test for decoding using simulated payload.

Tests/PointFreeTests/__Snapshots__/StripeWebhooksTests/testValidHook.1.Conn.txt
```diff
@@ -6,7 +6,7 @@
...
+  {"id":"evt_test","data":{"object":{"id":"in_test","amount_paid":1700,"closed":true,"lines":{"has_more":false,"data":[{"amount":1700,"id":"ii_test","quantity":1,"subscription":"sub_test","plan":{"amount":1700,"name":"Individual Monthly","id":"individual-monthly","created":1517356800,"currency":"usd","metadata":{},"interval":"month"}}]},"customer":"cus_test","charge":"ch_test","total":1700,"period_start":1517356800,"amount_remaining":0,"subtotal":1700,"date":1517356800,"period_end":1519948800,"number":"0000000-0000","subscription":"sub_test"}},"type":"invoice.payment_failed"}
 
 ▿ Response
-  Status 200 OK
\ No newline at end of file
+  Status 400 BAD REQUEST
```
stephencelis: Lookin into this...

stephencelis: Fixing these tests now. Need to update the sigs.

Sources/PointFree/Account/Change.swift
mbrandonw: very nice

Tests/PointFreeTests/__Snapshots__/StripeWebhooksTests/testValidHook.1.Conn.txt
mbrandonw: so we expect a 400 on this?

Sources/PointFree/Stripe.swift
```diff
@@ -128,7 +128,7 @@ public struct Stripe {
   public struct Invoice: Codable {
     public private(set) var amountDue: Cents
     public private(set) var amountPaid: Cents
-    public private(set) var charge: Charge?
+    public private(set) var charge: Either<Charge.Id, Charge>?
```
stephencelis: This is the root of the bug. When we added support to view historic invoices, we needed to expand information to see the charge. Stripe webhooks do _not_ expand this kind of info, so we need to support the unexpanded state. `Either` and conditional conformance makes this simple!

PR #254, opened 2018-05-30
Collapse posts on the blog index after 3.

mbrandonw:
### What

We're starting to have lots of blog posts, so let's shorten their representation on the blog index after the first 3.

image


Sources/PointFree/Blog/BlogPostIndex.swift
```diff
@@ -30,30 +30,81 @@ let blogIndexMiddleware: (Conn<StatusLineOpen, Tuple3<Database.User?, Subscriber
     }
 )
 
-private let blogIndexView = View<(Database.User?, SubscriberState)> { currentUser, subscriberState in
+private let blogIndexView = View<(Database.User?, SubscriberState)> { currentUser, subscriberState -> Node in
 
+  let allPosts = Current.blogPosts()
+    .sorted(by: their(^\.id, >))
+  let newPosts = allPosts.prefix(3)
+  let oldPosts = allPosts.dropFirst(3)
```
mbrandonw: we may want to more aggressively collapse posts. maybe everything after the first?

stephencelis: It's interesting...over time I don't think the historical posts matter in this ordinal way, but I think it's fine for now.

PR #258, opened 2018-05-31
Update 0003-StylingWithFunctions.swift

rehatkathuria:
Amend variable name


rehatkathuria:
Not the most important change, but something that stood out to me as I read the page!

stephencelis:
Good catch, thanks! 😅

PR #265, opened 2018-06-22
Extra invoice info

mbrandonw:
### What

Adds the ability for users to specify additional info to be added to their invoices. Good for people who need to expense their subscription and add a billing address, VAT, NIP, etc...

The info is stored in the `metadata` field of the stripe customer.


Sources/PointFree/Account/Account.swift
```diff
@@ -103,7 +103,7 @@ private let accountView = View<AccountData> { data in
     gridColumn(sizes: [.mobile: 12, .desktop: 8], [style(margin(leftRight: .auto))], [
       div([`class`([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]])])],
           titleRowView.view(unit)
-            <> profileRowView.view((data.currentUser, data.emailSettings))
+            <> profileRowView.view(data)
```
mbrandonw: this view grew to need more and more data from the `ProfileData`, so just passing the whole thing along now.

Sources/PointFree/Account/Account.swift
```diff
@@ -214,37 +214,61 @@ private let titleRowView = View<Prelude.Unit> { _ in
...
+    input([
+      `class`([blockInputClass]),
+      name(ProfileData.CodingKeys.email.stringValue),
+      type(.email),
+      value(data.currentUser.email.rawValue)
+      ])
+  ]
 
-          ] + emailSettingCheckboxes.view(currentEmailSettings) + [
+  let extraInvoiceInfoFields = !data.isSubscriptionOwner ? [] : [
```
mbrandonw: the extra invoice info field is only included when the user is the subscription owner.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -158,7 +158,7 @@ private func requireSubscriptionAndOwner<A>(
       <| middleware
 }
 
-private func fetchSubscription<A>(
+func fetchSubscription<A>(
```
mbrandonw: I'm able to reuse this middleware so making it non-private!

Sources/PointFree/Account/Invoices.swift
```diff
@@ -304,3 +305,26 @@ let invoiceView = View<(Stripe.Subscription, Database.User, Stripe.Invoice)> { s
...
 }
+
+private func extraInvoiceInfo(subscription: Stripe.Subscription) -> [Node] {
+  guard let extraInvoiceInfo = subscription.customer.right?.extraInvoiceInfo else { return [] }
+
+  let extraInvoiceInfoNodes = intersperse(Html.br)
+    <| extraInvoiceInfo
+      .components(separatedBy: CharacterSet.newlines)
+      .filter { !$0.isEmpty }
+      .map(Html.text)
```
mbrandonw: honestly computed this `extraInvoiceInfoNodes` was prob the toughest thing on the whole PR :/ all I wanted to do was split the newlines and insert `<br>` tags between em. I needed to add the `filter` in order to accommodate for `\r\n` type newlines. 

lemme know if you have some better ideas...

Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -22,24 +24,43 @@ extension ProfileData: Decodable {
...
+    self.extraInvoiceInfo = try? container.decode(String.self, forKey: .extraInvoiceInfo)
+    self.name = try container.decodeIfPresent(String.self, forKey: .name)
   }
 }
 
 func isValidEmail(_ email: EmailAddress) -> Bool {
   return email.rawValue.range(of: "^.+@.+$", options: .regularExpression) != nil
 }
 
+private func fetchStripeSubscription<A>(
```
mbrandonw: we have a middleware transformer that is similar to this except it was to produce an honest `Stripe.Subscription` and have an error otherwise, whereas for this screen I'm ok with not being able to find a stripe subscription. I tried to transform it in some way to be able to reuse, but couldn't figure it out.

Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -22,24 +24,43 @@ extension ProfileData: Decodable {
...
+      return Current.stripe.fetchSubscription(subscription.stripeSubscriptionId)
+        .run
+        .map(^\.right)
+        .flatMap { conn.map(const($0 .*. conn.data.second)) |> middleware }
+    }
+}
+
 let updateProfileMiddleware =
-  filterMap(require2 >>> pure, or: loginAndRedirect)
-    <<< filterMap(require1 >>> pure, or: redirect(to: .account(.index)))
```
mbrandonw: I changed the order of the connection's data to include the user first and then the `ProfileData`, so some `1`s and `2`s got flipped.

Sources/PointFree/Bootstrap.swift
```diff
@@ -79,7 +79,7 @@ private let loadEpisodes = { (_: Prelude.Unit) -> EitherIO<Error, Prelude.Unit>
 }
 
 private let connectToPostgres =
-  print(message: "  ⚠️ Connecting to PostgreSQL...")
+  print(message: "  ⚠️ Connecting to PostgreSQL at \(Current.envVars.postgres.databaseUrl)")
```
mbrandonw: this was helpful for me to confirm which database I was connecting to...

Sources/PointFree/Account/UpdateProfile.swift
mbrandonw: oh I think I know how this should be done, but don't really wanna spend too much more time on it.

the idea is that the `fetchStripeSubscription` middleware transformer in `Cancel.swift` should really just be middleware since it doesn't need to do any state changes. with that done I think it should then be possible to `map` over it and make it pluggable into what I'm doing here.

can look at this another time, just writing this for posterity.

mbrandonw:
Gonna merge this, but when yer back from vacation we can talk about it!

PR #266, opened 2018-06-22
Blog: Code and comment typos

lacyrhoades:
Pedantic but helpful changes to code and comments in blog post


lacyrhoades:
Also the blurb is "Let's create a type-safe interface to dealing with seconds ..." maybe should be "interface _for_ dealing with" IDK

PR #281, opened 2018-07-19
Use `xcrun` and document CPostgreSQL

stephencelis:
Addressing a couple potential gotchas for folks trying to get the site up and running.


Makefile
```diff
@@ -98,7 +98,7 @@ db-drop:
 
 xcodeproj-oss: check-dependencies
 	@echo "  ⚠️  Generating \033[1mPointFree.xcodeproj\033[0m..."
-	@swift package generate-xcodeproj --xcconfig-overrides=OSS.xcconfig >/dev/null \
+	@xcrun swift package generate-xcodeproj --xcconfig-overrides=OSS.xcconfig >/dev/null \
```
stephencelis: Swift.org documents adding the toolchain directory to your `PATH`, which is incompatible with some of this stuff. Using `xcrun` ensures that the currently selected Xcode's swift runs instead.

stephencelis:
(Also snuck in a link to the tour in the README.)

PR #287, opened 2018-07-25
Add next/prev links to episode page

### LINKED ISSUE #216: Add Previous & Next buttons on video page
I'm currently catching up on videos, and it would be a better experience to go to the next video once I'm done with one, rather than having to return to the homepage to find the next video.

stephencelis:
Fixes #216.


Sources/PointFree/Episode/Show.swift
```diff
@@ -13,47 +13,48 @@ import Tuple
...
 )
 
 let useCreditResponse =
   filterMap(
-    over1(episode(forParam:)) >>> require1 >>> pure,
+    over1(episodes(current:)) >>> require1 >>> pure,
     or: writeStatus(.notFound) >=> respond(episodeNotFoundView.contramap(lower))
     )
     <<< { userEpisodePermission >=> $0 }
+//    <<< map(map(over2(^\.current)))
```
stephencelis: Tried getting this to work to avoid having to make `validateCreditRequest` care about `Zipper<Episode>` but was fighting the type system a lot.

Sources/PointFree/TODO.swift
```diff
@@ -372,3 +372,40 @@ public func responseTimeout(_ interval: TimeInterval)
       }
     }
 }
+
+public struct Zipper<Element> {
```
stephencelis: Simple `Zipper` type for now. Would be nice to build out to a library with more flexibility proper conformances and whatnot, but that's prob an episode!

Tests/PointFreeTests/EmailTests/__Snapshots__/NewEpisodeEmailTests/testNewEpisodeEmail_Announcement_NonSubscriber.1.txt
```diff
@@ -1,27 +0,0 @@
-
```
stephencelis: Found a few stale snaps.

Sources/PointFree/Episode/Show.swift
mbrandonw: I think this is the first time we've done this exact thing, though a version of it is kinda baked into `filterMap` already.

we could accomplish this by defining `contramap` on middlewares:

```swift
((A) -> B) -> (Middleware<I, J, B, C>) -> Middleware<I, J, A, C>
```

and then we could do `<<< contramap(over2(^\.current))`.

Tests/PointFreeTests/EpisodeTests.swift
```diff
@@ -16,10 +16,21 @@ class EpisodeTests: TestCase {
...
     update(&Current, \.database .~ .mock)
+//    record=true
   }
 
   func testEpisodePage() {
-    let episode = request(to: .episode(.left(Current.episodes().first!.slug)), session: .loggedOut)
+    let episodes: [Episode] = [
+      .mock |> \.id .~ 1 |> \.sequence .~ 1 |> \.title .~ "Functions Part 1",
+      .mock |> \.id .~ 2 |> \.sequence .~ 2 |> \.title .~ "Functions Part 2",
+      .mock |> \.id .~ 3 |> \.sequence .~ 3 |> \.title .~ "Functions Part 3",
```
mbrandonw: might be good to have test coverage with there's no next episode and no previous episode too.

PR #288, opened 2018-07-25
Update README Screen Shot

stephencelis:
Also compressing things a bit in ImageOptim!

The code in the playground was a bit old, and we now load our public eps, so this is a nice lil update!

The main screen shot is still using mock data, but that's maybe fine. What did you use to generate it?


stephencelis:
image

PR #291, opened 2018-08-02
Discounts

mbrandonw:
### What

Adds the ability for people to enter a coupon code when subscribing to get a discount that we create in Stripe.


Sources/PointFree/Account/Account.swift
```diff
@@ -426,61 +426,86 @@ public func nextBilling(for subscription: Stripe.Subscription) -> String {
   }
 }
 
-private let subscriptionPlanRows = View<Stripe.Subscription> { subscription in
-  return div([`class`([Class.padding([.mobile: [.top: 1, .bottom: 3]])])], [
-    gridRow([
+private let subscriptionPlanRows = View<Stripe.Subscription> { subscription -> Node in
+
+  let planRow = gridRow([
```
mbrandonw: broke out some rows of this part of the account page into variables so that I could assemble them more easily.

Sources/PointFree/Account/Account.swift
```diff
@@ -426,61 +426,86 @@ public func nextBilling(for subscription: Stripe.Subscription) -> String {
...
-              ])
-            ])
+        div([`class`([Class.padding([.mobile: [.leftRight: 1]])])], [
+          p([text(nextBilling(for: subscription))])
           ])
         ])
-      ]),
+      ])
+
+  let discountRow = subscription.discount.map { discount in
```
mbrandonw: this is a new row, has the discount info if a coupon was used.

Sources/PointFree/Account/Account.swift
```diff
@@ -426,61 +426,86 @@ public func nextBilling(for subscription: Stripe.Subscription) -> String {
...
 }
 
+private func discountDescription(for discount: Stripe.Subscription.Discount) -> String {
+  var result = "\(discount.coupon.name): "
+  if let percentOff = discount.coupon.percentOff {
+    result += "\(Int(percentOff))% off"
+  } else if let amountOff = discount.coupon.amountOff {
+    result += "$\(amountOff) off"
+  }
+  return result
```
mbrandonw: example of what this looks like from screenshot test:

image

Sources/PointFree/Account/Invoices.swift
```diff
@@ -156,9 +156,25 @@ private let invoicesRowView = View<Stripe.ListEnvelope<Stripe.Invoice>> { invoic
...
 }
 
+private func discountDescription(for discount: Stripe.Subscription.Discount, invoice: Stripe.Invoice) -> String {
+  return "\(format(cents: invoice.total - invoice.subtotal)) (\(discount.coupon.name))"
+}
+
 let invoiceView = View<(Stripe.Subscription, Database.User, Stripe.Invoice)> { subscription, currentUser, invoice -> Node in
 
-  gridRow([
+  let discountRow = invoice.discount.map { discount in
```
mbrandonw: new row in the invoice. example of what this looks like from screenshot test:

image

Sources/PointFree/Blog/BlogPosts/AllBlogPosts.swift
```diff
@@ -18,4 +18,5 @@ private let _allBlogPosts: [BlogPost] = [
   post0007_openSourcingNonEmpty,
   post0008_conditionalCoding,
   post0009_6moAnniversary,
+  post0010_studentDiscounts,
```
mbrandonw: could use some proof reading on the blog post. it's kinda slim so if you have some ideas of how to beef it up I'm all ears.

Sources/PointFree/Pricing.swift
```diff
@@ -243,6 +243,25 @@ private let suggestATopic = View<Prelude.Unit> { _ in
   ]
 }
 
+private let studentDiscounts = View<Prelude.Unit> { _ in
```
mbrandonw: new FAQ on `/pricing` page

Sources/PointFree/Stripe.swift
```diff
@@ -8,7 +8,7 @@ import UrlFormEncoding
 public struct Stripe {
   public var cancelSubscription: (Subscription.Id) -> EitherIO<Swift.Error, Subscription>
   public var createCustomer: (Database.User, Token.Id, String?) -> EitherIO<Swift.Error, Customer>
-  public var createSubscription: (Customer.Id, Plan.Id, Int) -> EitherIO<Swift.Error, Subscription>
+  public var createSubscription: (Customer.Id, Plan.Id, Int, SubscribeData.Coupon?) -> EitherIO<Swift.Error, Subscription>
```
mbrandonw: can supply a coupon when subscribing now. stripe automatically handles making sure the coupon exists and that it's redeemable (can set max redemptions to 1).

Sources/PointFree/Stripe.swift
```diff
@@ -141,6 +141,7 @@ public struct Stripe {
     public private(set) var closed: Bool
     public private(set) var customer: Customer.Id
     public private(set) var date: Date
+    public private(set) var discount: Stripe.Subscription.Discount?
```
mbrandonw: invoices can have a discount payload

Sources/PointFree/Stripe.swift
```diff
@@ -232,6 +234,7 @@ public struct Stripe {
     public private(set) var currentPeriodStart: Date
     public private(set) var currentPeriodEnd: Date
     public private(set) var customer: Either<Customer.Id, Customer>
+    public private(set) var discount: Discount?
```
mbrandonw: subscriptions can have a discount payload

Sources/PointFree/Stripe.swift
```diff
@@ -250,6 +253,28 @@ public struct Stripe {
...
+      public private(set) var coupon: Coupon
+
+      public struct Coupon: Codable {
+        public typealias Id = Tagged<Coupon, String>
+
+        public private(set) var amountOff: Int?
+        public private(set) var id: Id
+        public private(set) var name: String
+        public private(set) var percentOff: Float?
+        public private(set) var valid: Bool
```
mbrandonw: these 5 fields are the minimal fields I think we need. we dont even need `amountOff` right now, but thought I would keep it around just in case.

Sources/PointFree/Stripe.swift
```diff
@@ -333,15 +359,18 @@ private func createCustomer(user: Database.User, token: Stripe.Token.Id, vatNumb
...
-    return stripeDataTask("subscriptions?expand[]=customer", .post([
-      "customer": customer.rawValue,
-      "items[0][plan]": plan.rawValue,
-      "items[0][quantity]": String(quantity),
-      ]))
+    var params: [String: Any] = [:]
+    params["customer"] = customer.rawValue
+    params["items[0][plan]"] = plan.rawValue
+    params["items[0][quantity]"] = String(quantity)
+    params["coupon"] = coupon?.rawValue
```
mbrandonw: broke this out into multiple lines cause I dont think we can put nil values directly into a `[String: Any]` initialization. have to use subscripts...

Sources/PointFree/StripeHtml.swift
```diff
@@ -68,6 +68,12 @@ extension Stripe {
             placeholder("VAT Number (EU Customers Only)"),
             type(.text),
             ]),
+          input([
+            `class`([blockInputClass]),
+            name("coupon"),
+            placeholder("Coupon Code"),
```
mbrandonw: new fields in the expanded subscription form.

Sources/PointFree/Subscribe.swift
```diff
@@ -6,19 +6,35 @@ import Prelude
 import Tuple
 
 public struct SubscribeData: Codable {
-  public let pricing: Pricing
-  public let token: Stripe.Token.Id
-  public let vatNumber: String
+  public typealias Coupon = Tagged<SubscribeData, String>
+
+  public private(set) var coupon: Coupon?
```
mbrandonw: this is technically the coupon code... do you think I should rename?

Sources/PointFree/Subscribe.swift
```diff
@@ -6,19 +6,35 @@ import Prelude
...
+      or: redirect(
+        to: .pricing(nil, expand: nil),
+        headersMiddleware: flash(.error, "An invalid subscription quantity was used.")
+      )
+    )
+    <<< filter(
+      get1 >>> validateCoupon(forSubscribeData:),
+      or: redirect(
+        to: .pricing(nil, expand: nil),
+        headersMiddleware: flash(.error, "Coupons can only be used on individual subscription plans.")
```
mbrandonw: right now I'm limiting discounts to only individual subscriptions. you think that makes sense?

Sources/PointFreeTestSupport/MockEpisode.swift
```diff
@@ -132,12 +132,12 @@ As server-side Swift becomes more popular and widely adopted, it will be importa
   codeSampleDirectory: "ep1-type-safe-html",
   id: 1,
   exercises: [],
-  image: "https://d1hf1soyumxcgv.cloudfront.net/0000-introduction/poster.jpg",
```
mbrandonw: getting rid of some live urls to make screenshots more predictable.

Sources/PointFree/Stripe.swift
stephencelis: Ah, I think we just use `[String: String]`, so we could tighten that up maybe and use `compactMapValues`.

Sources/PointFree/Subscribe.swift
stephencelis: I don't mind either way. This is consistent with Stripe at least.

Sources/PointFree/Subscribe.swift
stephencelis: Yup! Though we may wanna reuse the same mechanism for team discounts in the future, we can change it then.

PR #298, opened 2018-08-15
Update BlogPost0011_SolutionsToZipExercisesPt1.swift

timothybarraclough:
Correct minor error in Zip pt.1 exercises


mbrandonw:
Good catch!

PR #303, opened 2018-08-21
Event discount routes

mbrandonw:
Since we are doing more event discounts, I thought we should have some nice URL's to hand out to people that automatically renders the pricing screen with all the fields expanded. 

I also think we could also prob ditch the billing address fields, or move them to a different flag. Whatdya think?


Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -14,11 +14,13 @@ public enum Route: DerivePartialIsos {
   case admin(Admin)
   case appleDeveloperMerchantIdDomainAssociation
   case blog(Blog)
+  case discounts(organization: String)
```
mbrandonw: this is `/discounts/{org_name}`

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -14,11 +14,13 @@ public enum Route: DerivePartialIsos {
...
   case episode(Either<String, Int>)
   case episodes
   case expressUnsubscribe(userId: Database.User.Id, newsletter: Database.EmailSetting.Newsletter)
   case expressUnsubscribeReply(MailgunForwardPayload)
   case feed(Feed)
+  case fika
```
mbrandonw: this is just `/fika`. unfortunately we have to support this cause I already printed cards, oops!

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -116,6 +116,10 @@ private func render(conn: Conn<StatusLineOpen, T3<Database.Subscription?, Databa
       return conn.map(const(user .*. subscriberState .*. route .*. subRoute .*. unit))
         |> blogMiddleware
 
+    case .discounts:
```
mbrandonw: not currently doing anything with the organization name in the route. we could add a referrals column to subscriptions tho. whatdya think?

Sources/PointFree/__Generated__/DerivedPartialIsos.swift
```diff
@@ -139,6 +139,18 @@ import Prelude
       }
 
 
+      extension PartialIso where A == (
+        String
+      ), B == Route {
+
+        public static let discounts = parenthesize <| PartialIso(
+          apply: Route.discounts,
```
mbrandonw: sourcery is completely butchering this file now so I'm just copying and pasting these partial iso helpers.

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -451,8 +451,13 @@ extension Session {
     |> \.userId .~ Database.User.mock.id
 }
 
-public func request(to route: Route, session: Session = .loggedOut, basicAuth: Bool = false) -> URLRequest {
-  var request = router.request(for: route, base: URL(string: "http://localhost:8080"))!
+public func request(
+  with baseRequest: URLRequest,
+  session: Session = .loggedOut,
+  basicAuth: Bool = false
+  ) -> URLRequest {
```
mbrandonw: this is like the `request(to:session:basicAuth:)` helper except it takes in a request instead of a route. this is helpful for when you want to very explicitly test a URL instead of just a route value.

Tests/PointFreeTests/__Snapshots__/DiscountsTests/testDiscounts_LoggedIn.1.Conn.txt
```diff
@@ -0,0 +1,2924 @@
...
+                                  <option value="KR">
+                                    Korea, Republic of
+                                  </option>
+                                  <option value="KW">
+                                    Kuwait
+                                  </option>
+                                  <option value="KG">
+                                    Kyrgyzstan
+                                  </option>
+                                  <option valu
```
mbrandonw: this test snapshot is for when you are logged in on the /discounts/{orgname} url, and here we can see the coupon code is exposed.

Tests/PointFreeTests/__Snapshots__/DiscountsTests/testDiscounts_LoggedOut.1.Conn.txt
```diff
@@ -0,0 +1,2082 @@
...
+                  <div class="bg-white col-m col-m-12">
+                    <p class="content1 m-pb1 fg-gray400 fg-black normal h6 lh-4 w-100p center normal">
+                      20% off the Individual Monthly plan
+                    </p>
+                  </div>
+                </div>
+                <div class="bg-white row">
+                  <div class=" col-m col-m-12">
+                    <div class="m-pb3 m-pt2">
+                      <a href="/login?redirect=http://localhost:8080/pricing?expand%3Dtrue" class="medium cursor-pointer nowrap btn-normal rounded border-none text-decoration-none pf-link-white fg-white bg-black h5 m-pl2 m-pr2">
```
mbrandonw: this is the test snapshot for when we are logged out on the /discount/{orgname} url, and here we can see that the login redirect goes back to the pricing page with the expand=true flag.

Tests/PointFreeTests/__Snapshots__/DiscountsTests/testFika_LoggedIn.1.Conn.txt
```diff
@@ -0,0 +1,2924 @@
...
+                                    Korea, Republic of
+                                  </option>
+                                  <option value="KW">
+                                    Kuwait
+                                  </option>
+                                  <option value="KG">
+                                    Kyrgyzstan
+                                  </option>
+                                  <option value="LA">
+
```
mbrandonw: same as above but on the /fika url

Tests/PointFreeTests/__Snapshots__/DiscountsTests/testFika_LoggedOut.1.Conn.txt
```diff
@@ -0,0 +1,2081 @@
...
+                  <div class="bg-white col-m col-m-12">
+                    <p class="content1 m-pb1 fg-gray400 fg-black normal h6 lh-4 w-100p center normal">
+                      20% off the Individual Monthly plan
+                    </p>
+                  </div>
+                </div>
+                <div class="bg-white row">
+                  <div class=" col-m col-m-12">
+                    <div class="m-pb3 m-pt2">
+                      <a href="/login?redirect=http://localhost:8080/pricing?expand%3Dtrue" class="medium cursor-pointer nowrap btn-normal rounded border-none text-decoration-none pf-link-white fg-white bg-black h5 m-pl2 m-pr2">
```
mbrandonw: and again, same as above but on the /fika url

Sources/PointFree/SiteMiddleware.swift
stephencelis: Down if it'd be useful! Can still get the info from Stripe, though? As long as we somehow link the coupons?

Sources/PointFree/__Generated__/DerivedPartialIsos.swift
stephencelis: :(

Sources/PointFree/SiteMiddleware.swift
mbrandonw: yeah all the info should be in stripe. can either back port it or see if there's a way to generate reports with that info

Issue #297, opened 2018-08-15
Don't show team options when you aren’t on a team subscription


jasonzurita:
When a user has an individual subscription, and that user clicks to remove themselves from their own team, that user will disassociate their subscription from themselves - i.e., lose their subscription. The remove option should either be changed or removed to prevent accidentally losing a subscription (like I did 😬).

---

![screen shot 2018-08-15 at 3 19 59 pm](https://user-images.githubusercontent.com/13800855/44168493-794af580-a09f-11e8-9a26-9e103413cad6.png)

mbrandonw:
Good find, we'll have a fix soon!

mbrandonw:
It's pretty small effort. I'll give a few hints and if yer down to try then please do!

* This is the view that renders that lil fragment of HTML.
https://github.com/pointfreeco/pointfreeco/blob/48e85043005966cefb871b1622a12d29e5f24760/Sources/PointFree/Account/Account.swift#L538-L556

* That view is rendered inside this view, which stacks all of the team subscription related stuff on top of each other:
https://github.com/pointfreeco/pointfreeco/blob/48e85043005966cefb871b1622a12d29e5f24760/Sources/PointFree/Account/Account.swift#L325-L346

* And finally that view is rendered inside this view, which does the logic to check if it's the subscription owner viewing this page or not:
https://github.com/pointfreeco/pointfreeco/blob/48e85043005966cefb871b1622a12d29e5f24760/Sources/PointFree/Account/Account.swift#L314-L323

* The bug is that we only check `data.isSubscriptionOwner`, but what we really want to check is if the user is the owner of the subscription _and_ the subscription is a team subscription. Individual subscriptions can just fallthrough to the next conditional.

* That `isSubscriptionOwner` helper is defined here, and so we just need a new helper to describe the logic that we actually want:
https://github.com/pointfreeco/pointfreeco/blob/48e85043005966cefb871b1622a12d29e5f24760/Sources/PointFree/Account/Account.swift#L767-L769

If you end up fixing this bug you will also need to fix the tests. That should mean just running the `AccountTests` in record mode and verifying that everything looks correct. 

Wanna give it a shot?

jasonzurita:
This is such a great breakdown, thanks. I will start digging into this and give it a shot!

With that said, I am sure you would be able to fix this a lot faster than that I will be able to, so please don't let me hold you back from getting this fixed sooner.

mbrandonw:
[...]

PR #304, opened 2018-08-23
Don't show team options when you aren’t on a team subscription

jasonzurita:
Resolves https://github.com/pointfreeco/pointfreeco/issues/297

rr @mbrandonw


Sources/PointFree/Account/Account.swift
```diff
@@ -767,4 +767,9 @@ private struct AccountData {
   var isSubscriptionOwner: Bool {
     return self.currentUser.id == self.subscriptionOwner?.id
   }
+
+  var isTeamSubscription: Bool {
+    guard let id = stripeSubscription?.plan.id else { return false }
+    return id == Tagged.teamMonthly || id == Tagged.teamYearly
```
mbrandonw: is it possible to drop the `Tagged` type from these?

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -26,6 +26,10 @@ extension Environment {
     |> (\Environment.database.fetchSubscriptionTeammatesByOwnerId) .~ const(pure([.mock]))
     |> \.database.fetchTeamInvites .~ const(pure([.mock]))
     |> \.stripe.fetchSubscription .~ const(pure(.teamYearly))
+
+  public static let individualMonthly = mock
+    |> (\Environment.database.fetchSubscriptionTeammatesByOwnerId) .~ const(pure([.mock]))
```
mbrandonw: you should be able to drop the `Environment` from this key path.

Sources/PointFree/Account/Account.swift
```diff
@@ -767,4 +767,9 @@ private struct AccountData {
   var isSubscriptionOwner: Bool {
     return self.currentUser.id == self.subscriptionOwner?.id
   }
+
+  var isTeamSubscription: Bool {
+    guard let id = stripeSubscription?.plan.id else { return false }
```
mbrandonw: can you use `self.stripeSubscription`? we like explicit `self`s.

Sources/PointFree/Account/Account.swift
jasonzurita: yes! I ran into some issues with the type checker with an earlier version, and I accidentally left them in :(.

Sources/PointFree/Account/Account.swift
mbrandonw: we like the distinction between local/global/instance vars more than being explicit about capture. and there's not a single class in the entire Point-Free code base so no risk of capture!

jasonzurita: ahh, got ya — thanks, and the beauty of not being tied to Cocoa/Cocoa Touch!

mbrandonw:
Awesome, this looks great! Gonna merge and deploy now...

PR #305, opened 2018-09-05
Fix for subscription owners to remove themselves

mbrandonw:
### What

Fixes the issue we saw today. This more correctly computes who the owner of the subscription is so that they can still manage the team even when they remove themselves from the subscription.


Sources/PointFree/Auth.swift
```diff
@@ -86,8 +86,8 @@ public func currentSubscriptionMiddleware<A, I>(
     )
     ?? throwE(unit)
 
-  return (userSubscription.run.parallel <|> ownerSubscription.run.parallel)
-    .sequential
+  return (userSubscription <|> ownerSubscription)
+    .run
```
mbrandonw: this was causing a race condition in my tests. not sure why I raced the parallels instead of using the first success of `EitherIO`.

mbrandonw: maybe we need `EitherPar<E, A>`

PR #311, opened 2018-09-27
Upgrade Sourcery to 0.15.0

mbrandonw:
This fixes a bunch of bugs with Sourcery and Swift 4.2

I also had to make some changes to our `DerivePartialIsos.stencil`. It's _still_ not perfect, but I was able to hack it to work. I'd love to get rid of Sourcery for this and just use SwiftSyntax.


.sourcery-templates/DerivePartialIsos.stencil
```diff
@@ -9,8 +9,8 @@ import Prelude
 
       extension PartialIso where A == (
         {% for value in case.associatedValues %}
-          {% if value.type %}
-            {{ value.type.globalName }}{% if value.typeName.isOptional %}?{% endif %}
+          {% if value.type and value.type.parent %}
+            {{ value.type.name }}
```
mbrandonw: this is the only real change in this diff...

there are two places you can get a type name it seems, `value.type.name` and `value.typeName`. The latter omits generics and the former omits nested types. Let's just pray we never have nested generic types :/

PR #317, opened 2018-10-04
Podcast RSS final touches

mbrandonw:
This should be everything we need to launch the feature, other than some data entry.


Sources/PointFree/AtomFeed.swift
```diff
@@ -153,7 +167,7 @@ that day!
     description: description(episode: episode),
     dublinCore: .init(creators: ["Brandon Williams", "Stephen Celis"]),
     enclosure: enclosure(episode: episode),
-    guid: url(to: .episode(.left(episode.slug))),
+    guid: String(describing: episode.freeSince ?? episode.publishedAt),
```
mbrandonw: using the episodes free date / published date as the GUID allows us to create a new entry for episodes that were subscriber only and then switched to free.

Sources/PointFree/AtomFeed.swift
```diff
@@ -183,7 +197,14 @@ public extension Application {
 
 public func respond<A>(_ view: View<A>, contentType: MediaType = .html) -> Middleware<HeadersOpen, ResponseEnded, A, Data> {
   return { conn in
-    conn |> respond(body: view.rendered(with: conn.data), contentType: contentType)
+    conn
+      |> respond(
+        body: view.rendered(
+          with: conn.data,
+          config: Current.envVars.appEnv == .testing ? .pretty : .compact
```
mbrandonw: pretty printing for our atom feed tests

Sources/PointFree/EnvVars.swift
```diff
@@ -20,8 +20,9 @@ public struct EnvVars: Codable {
 
   public enum AppEnv: String, Codable {
     case development
-    case staging
     case production
+    case staging
+    case testing
```
mbrandonw: added a `testing` case so that we could pretty print only for tests. we could prob also get rid of staging.

Sources/PointFree/Episode/Episode.swift
```diff
@@ -28,9 +28,9 @@ public struct Episode {
   public init(
     blurb: String,
     codeSampleDirectory: String,
-    id: Id,
     exercises: [Exercise],
     fullVideo: Video? = nil,
+    id: Id,
```
mbrandonw: somehow this got out of order, had to fix all the episodes too

Sources/PointFree/FeatureFlags.swift
```diff
@@ -12,9 +12,12 @@ extension Array where Element == Feature {
...
-    return
-      feature.isEnabled
-        || (feature.isAdminEnabled && user?.isAdmin == .some(true))
+    return self
+      .first(where: { $0.name == feature.name })
+      .map {
+        $0.isEnabled
+          || ($0.isAdminEnabled && user?.isAdmin == .some(true))
+      }
+      ?? false
```
mbrandonw: ok, the whole `[Feature].hasAccess(to:)` API is so whack. I dont know what I was thinking when I wrote it. here I just did a minimal amount of work to make it sensible, but we should revamp it sometime soon.

Sources/PointFree/PageLayout.swift
```diff
@@ -121,17 +121,31 @@ func respond<A, B>(
         >=> respond(
           body: pageLayout.rendered(
             with: newLayoutData,
-            config: Current.envVars.appEnv == .production ? .compact : .pretty
+            config: Current.envVars.appEnv == .testing ? .pretty : .compact
```
mbrandonw: let's only pretty print in testing, not development.

Sources/PointFree/AtomFeed.swift
stephencelis: Is `String(describing:)` subject to weird local/locale formatting? Should we use a `DateFormatter` to be safe for testing?

Sources/PointFree/PageLayout.swift
```diff
@@ -121,17 +121,31 @@ func respond<A, B>(
...
+
+
+
+  return View { layoutData -> Node in
+
+    let hasPodcastRssFeature = Current.features.hasAccess(to: .podcastRss, for: layoutData.currentUser)
+    let episodeAtomFeed = Html.link([
+      hasPodcastRssFeature
+        ? href(url(to: .feed(.episodes)))
+        : href(url(to: .feed(.atom))),
```
mbrandonw: feature flagging the feeds in the meta tags.

Sources/PointFree/PublicEpisodes/0000-Introduction.swift
```diff
@@ -2,12 +2,12 @@ import Foundation
 
 let introduction = Episode(
   blurb: """
-Point-Free is here, bringing you videos covering functional programming concepts using the Swift language.
+Point-Free is here, bringing you videos covering functional programming concepts using the Swift language. \
```
mbrandonw: best to not have newlines for episode blurbs cause some readers will render them

Tests/PointFreeTests/AtomFeedTests.swift
```diff
@@ -1,16 +1,50 @@
...
   }
+
+  func testEpisodeFeed() {
+    let conn = connection(from: request(to: .feed(.episodes)))
+    let result = conn |> siteMiddleware
+
+    assertSnapshot(matching: result.perform())
+  }
+
+  func testEpisodeFeed_WithRecentlyFreeEpisode() {
```
mbrandonw: this test shows that an older episode that was recently made free gets bumped to the top

Tests/PointFreeTests/__Snapshots__/PrivateRssTests/testFeed_Authenticated_InActiveSubscriber.1.Conn.txt
```diff
@@ -0,0 +1,143 @@
...
+      Point-Free
+    </title>
+    <link>
+      http://localhost:8080
+    </link>
+    <language>
+      en-US
+    </language>
+    <description>
+      ‼️ The URL for this feed has been turned off by Point-Free. You can retrieve your most up-to-date private podcast URL by visiting your account page at http://localhost:8080/account. If you think this is an error, please contact support@pointfree.co.
```
mbrandonw: this is what the feed looks like for non-subscribers / in-active subscribers.

Sources/PointFree/AtomFeed.swift
mbrandonw: oh im not sure. I could also just use the underlying time interval. how's that sound?

stephencelis: Oh yeah that sounds perf!

mbrandonw: coo done 5fd134fe

PR #319, opened 2018-10-06
Add video.js player

mbrandonw:
This adds the [video.js player](http://videojs.com) for our episode videos, which allows playback speed to be changed. I've done some testing in safari, chrome and iPhone and it seems good, but would feel better with this change if you also ran this locally and gave it a whirl.

image

image


Sources/PointFree/Blog/BlogPosts/BlogPost0006_TaggedSecondsAndMilliseconds.swift
```diff
@@ -10,7 +10,7 @@ Let's create a type-safe interface for dealing with seconds and milliseconds in
     .init(
       content: "",
       timestamp: nil,
-      type: .image(src: "https://s3.amazonaws.com/pointfreeco-blog/posts/0006-tagged-seconds-and-milliseconds/poster.jpg")
+      type: .image(src: "https://d1iqsrac68iyd8.cloudfront.net/posts/0006-tagged-seconds-and-milliseconds/poster.jpg")
```
mbrandonw: found some missing cloudfront urls

Sources/PointFree/PageLayout.swift
```diff
@@ -54,6 +54,7 @@ struct SimplePageLayoutData<A> {
   private(set) var currentUser: Database.User?
   private(set) var data: A
   private(set) var description: String?
+  private(set) var extraHead: [ChildOf<Element.Head>]
```
mbrandonw: new layout field to be able to add extra things to the head

Sources/PointFree/PageLayout.swift
```diff
@@ -143,7 +146,7 @@ func simplePageLayout<A>(_ contentView: View<A>) -> View<SimplePageLayoutData<A>
       ])
 
     return document([
-      html([
+      html([lang(.en)], [
```
mbrandonw: saw an audit warning about this so thought I'd add it in

Sources/PointFree/VideoJs.swift
```diff
@@ -0,0 +1,40 @@
+import Css
+import Foundation
+import Html
+import HtmlCssSupport
+
+var videoJsHead: [ChildOf<Element.Head>] {
+  let videoJsAssets: [ChildOf<Element.Head>] = [
```
mbrandonw: this is the js and css needed to use video.js

Sources/PointFree/VideoJs.swift
```diff
@@ -0,0 +1,40 @@
...
+    link([
+      href("https://cdnjs.cloudflare.com/ajax/libs/video.js/7.2.4/alt/video-js-cdn.min.css"),
+      rel(.stylesheet)
+      ]),
+    .init(script([
+      src("https://cdnjs.cloudflare.com/ajax/libs/video.js/7.2.4/video.min.js"),
+      `defer`(true)
```
mbrandonw: I think by marking this as `defer` I can keep this script in the head instead of injecting it at the bottom of the `body`. does that sound right?

Sources/PointFree/VideoJs.swift
```diff
@@ -0,0 +1,40 @@
...
+      ]))
+  ]
+
+  return [
+    style(".vjs-subs-caps-button" % display(.none)),
```
mbrandonw: this hides the closed caption button in the video player. this is the simplest way I could find :/

Sources/PointFree/VideoJs.swift
```diff
@@ -0,0 +1,40 @@
...
+    script("window.HELP_IMPROVE_VIDEOJS = false;")
```
mbrandonw: this prevents video.js from sending stuff to google analytics.

Sources/PointFree/VideoJs.swift
```diff
@@ -0,0 +1,40 @@
...
+}
+
+let videoClasses: CssSelector =
+  ".video-js"
+    | ".vjs-default-skin"
+    | ".vjs-big-play-centered"
+
+struct VideoJsOptions: Encodable {
+  let control: Bool
+  let playbackRates: [Double]
```
mbrandonw: some video.js options that you can pass into the `data-setup` attribute of the video tag as a json string.

Tests/PointFreeTests/AccountTests/__Snapshots__/AccountTests/testAccount.1.Conn.txt
```diff
@@ -12,7 +12,7 @@ X-Permitted-Cross-Domain-Policies: none
 X-XSS-Protection: 1; mode=block
 
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
mbrandonw: lots of lil changes due to this

Sources/PointFree/VideoJs.swift
```diff
@@ -0,0 +1,40 @@
...
+      src("https://cdnjs.cloudflare.com/ajax/libs/video.js/7.2.4/video.min.js"),
+      `defer`(true)
+      ]))
+  ]
+
+  return [
+    style(".vjs-subs-caps-button" % display(.none)),
+    script("window.HELP_IMPROVE_VIDEOJS = false;")
+    ]
+    + (Current.envVars.appEnv == .testing ? [] : videoJsAssets)
```
mbrandonw: I'm not happy with this but not sure what else to do. if we serve up the video.js assets during testing we get non-deterministic screenshots cause sometimes things load in time.

Sources/PointFree/Episode/Show.swift
```diff
@@ -225,11 +226,13 @@ private let videoView = View<(Episode, isEpisodeViewable: Bool)> { episode, isEp
     [
       video(
         [
-          `class`([innerVideoContainerClass]),
+          `class`([innerVideoContainerClass, videoClasses]),
+          style(position(.absolute)),
```
mbrandonw: the `innerVideoContainerClass` styles contains a `position: absolute`, but it seems video.js overrides that with relative position. so I use this inline style to override that :/

Sources/PointFree/VideoJs.swift
stephencelis: Ha yeesh.

Sources/PointFree/VideoJs.swift
stephencelis: Seems fine to me

PR #323, opened 2018-10-08
PFP0020

mbrandonw:


Tests/PointFreeTests/AccountTests/__Snapshots__/AccountTests/testAccountCanceledSubscription.1.Conn.txt
```diff
@@ -1727,6 +1727,32 @@ html {
               </div>
             </div>
           </div>
+          <div class=" row">
```
mbrandonw: this test is for the account page with a user that has canceled their subscription, yet for some reason the podcast callout is showing. we need to look into this, but still going to merge for now. I have a trello card for it.

PR #324, opened 2018-10-10
Fix rss link for team subs.

mbrandonw:


Sources/PointFree/Account/PrivateRss.swift
```diff
@@ -9,7 +9,7 @@ let accountRssMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple2<Datab
 { fetchUser >=> $0 }
   <<< filterMap(require1 >>> pure, or: invalidatedFeedMiddleware)
   <<< validateUserAndSalt
-  <<< fetchSubscription
+  <<< fetchUserSubscription
```
mbrandonw: just a quick fix for now, but would love to move all of our non-private middleware transformers to a file so that we can properly share the.

stephencelis:
Good catch!

PR #330, opened 2018-10-21
Snap Helpers

stephencelis:
I know it'd be nice to somehow bake this into the strategy, but could be a nice stopgap till then.

### DIFF
--- a/Sources/PointFreeTestSupport/PointFreeTestSupport.swift
+++ b/Sources/PointFreeTestSupport/PointFreeTestSupport.swift
@@ -504,7 +504,7 @@ extension SnapshotTestCase {
     ) {
 
     strategies.forEach { name, strategy in
-      self.assertSnapshot(
+      assertSnapshot(
         of: strategy,
         matching: value,
         named: name,
@@ -528,10 +528,9 @@ extension SnapshotTestCase {
     ) {
 
     strategies.forEach { strategy in
-      self.assertSnapshot(
+      assertSnapshot(
         of: strategy,
         matching: value,
-        named: name,
         record: recording,
         timeout: timeout,
         file: file,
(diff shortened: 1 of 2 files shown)


Tests/PointFreeTests/AboutTests.swift
```diff
@@ -17,16 +17,12 @@ class AboutTests: TestCase {
...
-      )
-
-      assertSnapshot(
-        of: .ioConnWebView(size: .init(width: 400, height: 2300)),
-        matching: conn |> siteMiddleware,
-        named: "mobile"
+      assertSnapshots(
+        of: [
+          "desktop": .ioConnWebView(size: .init(width: 1080, height: 2300)),
+          "mobile": .ioConnWebView(size: .init(width: 400, height: 2300))
```
stephencelis: A nice thing about baking this into the strategy instead is that it could maybe account for reusing the same loaded `webView` and merely resize it. We're currently taking on some extra work whenever we use `ioConnWebView`...

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -492,6 +492,56 @@ extension Strategy {
...
+    file: StaticString = #file,
+    function: String = #function,
+    line: UInt = #line
+    ) {
+
+    strategies.forEach { strategy in
+      self.assertSnapshot(
+        of: strategy,
+        matching: value,
+        named: name,
```
mbrandonw: does an index need to be appended to this or anything to prevent overwriting?

stephencelis: Nah, named snaps don't have indices anymore.

PR #332, opened 2018-10-28
PFP0020

mbrandonw:
For tomorrow.


Sources/PointFree/Episode/Show.swift
```diff
@@ -229,7 +229,7 @@ private let videoView = View<(Episode, isEpisodeViewable: Bool)> { episode, isEp
         [
           id("episode-video"),
           Styleguide.class([innerVideoContainerClass, videoJsClasses]),
-          style(position(.absolute)),
+          style(position(.absolute) <> objectFit(.cover)),
```
mbrandonw: this makes the poster fill the video player even when it's not the exact aspect ratio. this is most noticeable on the blog where we don't have a perfect 16x9 ratio on video embeds.

mbrandonw:
Yeah alongside the episode. I think it's best to get it all out at the same time.

Issue #335, opened 2018-10-28
Fix atom feed rendering


mbrandonw:
I just noticed that our atom feed is currently broken:

https://www.pointfree.co/blog/feed/atom.xml

This happened with the `swift-html` update. I guess previously we weren't rendering `link` tags as void elements, and so now we don't render a closing tag for the XML and that is invalid. You can see the change in this link (warning: it's a big diff)

https://github.com/pointfreeco/pointfreeco/pull/326/files#diff-2473f1367f460721a7685c83d9d2f300

The relevant snapshot difference:

```diff
-  <link href="http://localhost:8080/feed/atom.xml" rel="self">
-  </link>
+  <link href="http://localhost:8080/feed/atom.xml"
+        rel="self">
```

I'm not really sure how to fix this. Do we need a separate renderer for XML documents?

If we don't want to fix the renderer we can switch from an Atom feed to an RSS feed. In RSS feeds the `link` element is used with a child text node instead of attributes
This is also affecting our RSS feed:

```
<link>https://www.pointfree.co</link>
```

I also have a temporary hacky fix for this coming in a PR...

mbrandonw:
Also worth mentioning this issue is further made complicated by the fact that the minimal renderer and pretty printer slightly differed in behavior, as show in 
https://github.com/pointfreeco/swift-web/pull/164.

It means that our RSS feeds aren't broke in production, but our Atom feeds are, and further the snapshots of both feeds are wrong.

PR #336, opened 2018-10-29
Don't close the link tag in atom feeds.

mbrandonw:
Fixes https://github.com/pointfreeco/pointfreeco/issues/335 by using the work in https://github.com/pointfreeco/swift-web/pull/164.

### DIFF
--- a/Sources/PointFree/Atom.swift
+++ b/Sources/PointFree/Atom.swift
@@ -34,8 +34,15 @@ public let atomLayout = View<AtomFeed> { atomFeed -> [Node] in
       [xmlns("http://www.w3.org/2005/Atom")],
       [
         title(atomFeed.title),
-        link([href(atomFeed.atomUrl), rel(.self)]),
-        link([href(atomFeed.siteUrl)]),
+        element(
+          "link",
+          [
+            .init("href", atomFeed.atomUrl) as Attribute<Void>,
+            .init("rel", "self")
+          ],
+          [""]
+        ),
+        element("link", [.init("href", atomFeed.siteUrl) as Attribute<Void>], [""]),
         atomFeed.entries.map(^\.updated).max().map(updated),
         id(atomFeed.siteUrl),
         author([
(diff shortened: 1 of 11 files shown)


Sources/PointFree/Atom.swift
```diff
@@ -34,8 +34,15 @@ public let atomLayout = View<AtomFeed> { atomFeed -> [Node] in
       [xmlns("http://www.w3.org/2005/Atom")],
       [
         title(atomFeed.title),
-        link([href(atomFeed.atomUrl), rel(.self)]),
-        link([href(atomFeed.siteUrl)]),
+        element(
+          "link",
+          [
+            .init("href", atomFeed.atomUrl) as Attribute<Void>,
+            .init("rel", "self")
```
mbrandonw: I can't use any of the attribute helpers because their `HasABC` restrictions don't match up for what I need. I also need to cast to `Attribute<Void>` to take care of the phantom type.

Sources/PointFree/Atom.swift
```diff
@@ -34,8 +34,15 @@ public let atomLayout = View<AtomFeed> { atomFeed -> [Node] in
...
+          ],
+          [""]
```
mbrandonw: this is the hack. I'm putting in a child node so that the closing tag is rendered :/

Tests/PointFreeTests/__Snapshots__/AtomFeedTests/testAtomFeed.1.Conn.txt
```diff
@@ -18,7 +18,11 @@ X-XSS-Protection: 1; mode=block
   </title>
   <link href="http://localhost:8080/feed/atom.xml"
         rel="self">
+    
+  </link>
   <link href="http://localhost:8080">
+    
+  </link>
```
mbrandonw: these are the closing tags we need

Sources/PointFree/Atom.swift
stephencelis: Ha, no NB?

mbrandonw: done

PR #338, opened 2018-11-04
Remove error email and expose all episodes in RSS.

mbrandonw:
Removing that `Private Rss Feed Error` email cause I think we know what is going on there. Also I'm now exposing all episodes in the private RSS feed. Seems maybe not worth doing what we are doing to combat fraudulent activity. Thoughts?

### DIFF
--- a/Sources/PointFree/Account/PrivateRss.swift
+++ b/Sources/PointFree/Account/PrivateRss.swift
@@ -132,7 +132,10 @@ with anyone else.
 private func items(forUser user: Database.User) -> [RssItem] {
   return Current
     .episodes()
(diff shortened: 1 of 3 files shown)


stephencelis:
Can we do a month-size window instead?

mbrandonw:
> Can we do a month-size window instead?

coo yeah let's try that.

mbrandonw:
> Looks good but I think still needs a `.prefix(4)` or somethin.

why's that? I'm only exposing episodes published in the last 30 days.

stephencelis:
Ah my brain skipped right past that. Looks good!

PR #340, opened 2018-11-06
Adding some references

mbrandonw:


Sources/PointFree/PublicEpisodes/References.swift
```diff
@@ -0,0 +1,273 @@
...
+""",
+    link: "https://www.pointfree.co/blog/posts/6-tagged-seconds-and-milliseconds",
+    publishedAt: Date(timeIntervalSince1970: 1531886400),
+    title: "Tagged Seconds and Milliseconds"
+  )
+
+  static let theoremsForFree = Episode.Reference(
+    author: "Philip Wadler",
+    blurb: """
+This famous paper describes the "theorems for free" result, in which if you write down a generic function
```
stephencelis: Not sure what this line is trying to convey: "'theorems for free' result" in particular. Is there a better word we can use here other than "result"?

stephencelis:
This is great! Made a quick editing pass but only really have one question.

PR #343, opened 2018-11-08
Test GitHub/Stripe Requests

stephencelis:
Linux Foundation has some buuugs that keep biiiiiting us. This'll hopefully catch some of em.


Sources/PointFree/Stripe.swift
```diff
@@ -457,48 +461,42 @@ private func attachMethod(_ method: Method) -> (URLRequest) -> URLRequest {
...
-    }
+          }
+      }
+  }
 
-    switch method {
-    case .delete, .get:
-      return task
-    case .post:
-      return task.retry(maxRetries: 3)
```
stephencelis: Btw removing this since we don't think it makes a diff.

PR #344, opened 2018-11-08
Deploy to Swift 4.2.1

mbrandonw:
For posterity, there are some Linux Foundation bugs that were fixed in Swift 4.1, regressed in Swift 4.2, and then fixed in 4.2.1 :/

Today we took down subscriptions for a few minutes because we deployed to Swift 4.2.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM swift:4.2
+FROM norionomura/swift:421
 
 # postgres
 RUN apt-get update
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM swift:4.2
+FROM norionomura/swift:421
```
stephencelis: 🙇 Thank you @norio-nomura!

PR #345, opened 2018-11-09
Swift 4.2.1

stephencelis:
Woof.

### DIFF
--- a/Sources/PointFree/GitHub.swift
+++ b/Sources/PointFree/GitHub.swift
@@ -66,7 +66,7 @@ func fetchAuthToken(with code: String) -> DecodableRequest<Either<GitHub.OAuthEr
 
   var request = URLRequest(url: URL(string: "https://github.com/login/oauth/access_token")!)
   request.httpMethod = "POST"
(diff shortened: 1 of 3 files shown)


Sources/PointFree/GitHub.swift
```diff
@@ -107,5 +107,15 @@ private func runGitHub<A>(_ gitHubRequest: DecodableRequest<A>) -> EitherIO<Erro
   return jsonDataTask(with: gitHubRequest.rawValue, decoder: gitHubJsonDecoder)
 }
 
+private let gitHubJsonEncoder: JSONEncoder = { () in
+  let encoder = JSONEncoder()
+
+  if #available(OSX 10.13, *) {
+    encoder.outputFormatting = [.sortedKeys]
```
stephencelis: I think this lil bit of extra work is fine to do in production, but if ya wanna move it into `Current` lemme know!

mbrandonw: yeahhhhh that's fine I think.

PR #350, opened 2018-11-28
Update screenshots.

mbrandonw:
I think a previous PR left some screenshots out of date.


stephencelis:
These still pass for me locally. I think because of that JavaScript that used to run to attempt to auto-size the result. I think screenshots in general will break when we update to the latest SnapshotTesting since we removed that stuff.

PR #353, opened 2018-12-02
Snapshot PFP

mbrandonw:
I think this is ready to go for tomorrow except for the cover image. I'll use something from what you do for the readme.

I had to update some prism js stuff so that I could syntax highlight diffs.


Sources/PointFree/Blog/BlogPosts/BlogPost0023_OpenSourcingSnapshotTesting.swift
```diff
@@ -0,0 +1,370 @@
...
+""",
+      timestamp: nil,
+      type: .paragraph
+    )
+
+    ],
+  coverImage: "TODO",
+  id: 23,
+  publishedAt: .init(timeIntervalSince1970: 1543827600),
+  title: "Announcing SnapshotTesting 1.0"
```
mbrandonw: don't love the -ing -ing of this title. do you have any other ideas?

PR #358, opened 2018-12-08
Better Coupon Integration

stephencelis:
Adds:

- The ability to fetch coupons from the Stripe API.
- Adds coupon duration.
- Fixes `(amountOff: Int?, percentOff: Float?)` to be `Either<Cents, Int>`
- Updates discount pricing page with all of the above


Sources/PointFree/Home.swift
```diff
@@ -46,7 +46,7 @@ let homeView = View<(Database.User?, SubscriberState)> { currentUser, subscriber
     <> episodesListView.view(secondBatch)
     <> (
       subscriberState.isNonSubscriber
-        ? pricingOptionsView.view((currentUser, .default, .minimal, nil))
+        ? pricingOptionsView.view((currentUser, .default, .minimal, nil, nil))
```
stephencelis: Now takes `Stripe.Coupon?`.

Sources/PointFree/Pricing.swift
```diff
@@ -95,22 +95,39 @@ extension Pricing: Codable {
...
           extraStyles: pricingExtraStyles <> whatToExpectStyles,
           style: .base(.minimal(.dark)),
           title: "Subscribe to Point-Free"
         )
     }
 )
 
+private func fetchCoupon<I>(
+  _ conn: Conn<I, Tuple5<Database.User?, Pricing, PricingFormStyle, Stripe.Coupon.Id?, Route?>>
+  ) -> IO<Conn<I, Tuple5<Database.User?, Pricing, PricingFormStyle, Either<String, Stripe.Coupon>?, Route?>>> {
```
stephencelis: I know there's probably a nice way of using `over4` and some `sequence`s to compose this, but was struggling a bit and just moved the `over4` inside...

Sources/PointFree/Pricing.swift
```diff
@@ -120,13 +137,12 @@ private let pricingOptionsRowClass =
     | Class.grid.center(.mobile)
     | Class.padding([.mobile: [.topBottom: 3, .leftRight: 2], .desktop: [.topBottom: 4, .leftRight: 0]])
 
-public enum PricingFormFields {
+public enum PricingFormStyle {
   case minimal
-  case coupon(String?)
```
stephencelis: Now `coupon` requires a valid one from Stripe.

Sources/PointFree/Pricing.swift
mbrandonw: should this be a middleware transformers so that we could do the flash when fetching the coupon fails?

Sources/PointFree/Pricing.swift
```diff
@@ -120,13 +137,12 @@ private let pricingOptionsRowClass =
...
   case full
 }
 
-let pricingOptionsView = View<(Database.User?, Pricing, PricingFormFields, Route?)> { currentUser, pricing, formFields, route in
+let pricingOptionsView = View<(Database.User?, Pricing, PricingFormStyle, Either<String, Stripe.Coupon>?, Route?)> { currentUser, pricing, formStyle, coupon, route in
```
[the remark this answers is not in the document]
mbrandonw: yeah I think a flash + redirect to `/pricing` makes sense.

PR #361, opened 2018-12-11
Use upcoming invoice for next billing info

stephencelis:
We were manually calculating the next invoice, which is problematic. This updates the account page to use Stripe's "upcoming invoice" API instead.

We still need to update the subscription "change" form to reflect discounts, though this is much more important to roll out first.


Sources/PointFree/Account/Account.swift
```diff
@@ -53,7 +53,11 @@ private func fetchAccountData<I>(
     .map(^\.stripeSubscriptionId)
     .flatMap(Current.stripe.fetchSubscription)
 
-  let everything = zip7(
+  let upcomingInvoice = stripeSubscription
+    .map(^\.customer >>> either(id, ^\.id))
+    .flatMap(Current.stripe.fetchUpcomingInvoice)
+
+  let everything = zip8(
```
stephencelis: 🎱

Sources/PointFree/Stripe.swift
```diff
@@ -93,7 +95,7 @@ public struct Stripe {
 
     public private(set) var duration: Duration
     public private(set) var id: Id
-    public private(set) var name: String
+    public private(set) var name: String?
```
stephencelis: Coupon name can be nil, so I've updated this to prevent coding failures.

Sources/PointFree/Stripe.swift
```diff
@@ -190,7 +192,7 @@ public struct Stripe {
     public private(set) var customer: Customer.Id
     public private(set) var date: Date
     public private(set) var discount: Discount?
-    public private(set) var id: Id
+    public private(set) var id: Id?
```
stephencelis: The id is nil for upcoming invoices...

Sources/PointFreeTestSupport/PointFreeTestSupport.swift
```diff
@@ -35,9 +35,10 @@ extension Environment {
   )
 
   public static let teamYearly = mock
-    |> (\.database.fetchSubscriptionTeammatesByOwnerId) .~ const(pure([.mock]))
-    |> \.database.fetchTeamInvites .~ const(pure([.mock]))
-    |> \.stripe.fetchSubscription .~ const(pure(.teamYearly))
+    |> (\Environment.database.fetchSubscriptionTeammatesByOwnerId) .~ const(pure([Database.User.mock]))
+    |> (\Environment.database.fetchTeamInvites) .~ const(pure([Database.TeamInvite.mock]))
+    |> (\Environment.stripe.fetchSubscription) .~ const(pure(Stripe.Subscription.teamYearly))
+    |> (\Environment.stripe.fetchUpcomingInvoice) .~ const(pure(Stripe.Invoice.upcoming |> \.amountDue .~ 640_00))
```
stephencelis: This expression was getting too complex, so had to get rid of some inference.

PR #362, opened 2018-12-11
Re-enable logging, and refactor a bit.

mbrandonw:
We've had logged turned off for a while now due to crashes with `print` on Linux (we assumed it had to do with threading, but even logging on a serial queue crashed). Looks like at some point that was maybe fixed, so I think we could re-enable. Also we are now set up with logentries which gives us 5gb of logs and 7 days of retention.

There's also a small refactoring here to use file handles. I wanna deploy this and make sure it's stable, and if it is I'd like to add a bunch of logging to various spots.

### DIFF
--- a/Sources/PointFree/Logger.swift
+++ b/Sources/PointFree/Logger.swift
@@ -1,43 +1,99 @@
+import Foundation
 #if os(Linux)
 import Glibc
 #else
 import Darwin.C
 #endif
 
-public struct Logger {
-  private(set) var level: Level = .debug
-  private(set) var logger: (String) -> () = { print($0) }
+public class StandardFileHandle: TextOutputStream {
+  fileprivate let handle: FileHandle
 
-  init(level: Level = .debug, logger: @escaping (String) -> () = { print($0) }) {
-    self.level = level
-    self.logger = logger
+  public static let error = StandardFileHandle(handle: .standardError)
+  public static let output = StandardFileHandle(handle: .standardOutput)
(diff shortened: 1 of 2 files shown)


Sources/PointFree/Logger.swift
```diff
@@ -1,43 +1,99 @@
...
   }
 
-  public func log<A>(_ level: Level, _ message: @autoclosure () -> A) {
+  public func log<A>(
+    _ level: Level,
+    _ message: @autoclosure () -> A,
+    file: StaticString = #file,
+    line: UInt = #line) {
+
+    let file = String(String(describing: file).split(separator: "/").last!)
```
mbrandonw: doing this so that logs look nice and short:

```
[debug] SiteMiddleware.swift:11: [Request] GET http://localhost:8080/
[debug] SiteMiddleware.swift:11: [Time]    88ms
```

Turns out SPM can't even handle multiple files with the same name, so I think this will sufficiently disambiguate, at least for now.

Sources/PointFree/Logger.swift
```diff
@@ -48,3 +104,20 @@ public struct Logger {
     case fatal
   }
 }
+
+import Either
+
+public func logError<A>(
```
mbrandonw: more of a proof-of-concept, not using it yet. but we could insert this into any `EitherIO` change to get error visibility. would also like to make middleware helpers that could be inserted in.

stephencelis: v cool!

PR #365, opened 2018-12-19
Prep YIR Blog

stephencelis:


Sources/PointFree/Blog/BlogPost.swift
```diff
@@ -24,12 +24,15 @@ public struct BlogPost {
 
   public enum Author {
     case brandon
+    case pointfree
```
stephencelis: Added a general `pointfree` author for shared posts and smaller promotions that don't really need authorship.

PR #366, opened 2018-12-19
Updates to episode columns and transcript images

mbrandonw:
I've done two things here:

* Made the episode columns equal size so that the video is quite a bit bigger. The "credits" at the bottom of the right column sometimes doesn't fit now, so I just killed it.

* Enchanced the `.image` transcript block to take a sizing enum `inset | fullWidth` so that we can make images look a lil nicer in the transcripts/blog posts.

I recommend running this locally to see how you like the bigger column. I think it looks nice, but def need a 2nd opinion.


Package.swift
```diff
@@ -12,10 +12,10 @@ let package = Package(
     .library(name: "Styleguide", targets: ["Styleguide"]),
     ],
   dependencies: [
-    .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("6f2fd6b")),
-    .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .revision("ca1a008")),
+    .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("8cbc934")),
+    .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", from: "1.1.0"),
```
mbrandonw: new snapshot record mode was super handy.

Sources/PointFree/Episode/Episode.swift
```diff
@@ -122,11 +122,20 @@ public struct Episode {
     public enum BlockType: Equatable {
       case code(lang: CodeLang)
       case correction
-      case image(src: String)
+      case image(src: String, sizing: ImageSizing)
       case paragraph
       case title
       case video(poster: String, sources: [String])
 
+      public static func image(src: String) -> BlockType {
```
mbrandonw: this provides backwards compatibility with all the `.image`'s out in the wild right now.

Sources/PointFree/Episode/Show.swift
```diff
@@ -811,29 +780,22 @@ let transcriptBlockView = View<Episode.TranscriptBlock> { block -> Node in
...
-              innerImageContainerClass,
-              // TODO: allow `.image` blocks to be customized for full-width vs inset
-//              Class.margin([.mobile: [.topBottom: 3]]),
-//              Class.padding([.mobile: [.leftRight: 3]]),
-//              Class.pf.colors.bg.white
-              ])
-          ]
-        )
-      ]
+      [img(src: src, alt: "", [`class`(imageClasses)])]
```
mbrandonw: we should prob add `alt` to the enum case too

Sources/PointFree/Episode/Show.swift
```diff
@@ -338,7 +334,7 @@ private let downloadsView = View<String> { codeSampleDirectory -> [Node] in
   guard !codeSampleDirectory.isEmpty else { return [] }
 
   return [
-    div([Styleguide.class([Class.padding([.mobile: [.leftRight: 3], .desktop: [.leftRight: 4]])])],
+    div([Styleguide.class([Class.padding([.mobile: [.leftRight: 3], .desktop: [.leftRight: 4]]), Class.padding([.mobile: [.bottom: 3]])])],
```
mbrandonw: screenshot test caught a regression here where the credits module had extra top padding to make up for the fact that the downloads maybe be absent. easy enough to fix

PR #370, opened 2018-12-19
Admin Mailer Fixes

stephencelis:
Still investigating the root cause, but this fixes things. The routing wasn't behaving as expected.


Sources/PointFree/Admin/SendNewBlogPostMailer.swift
```diff
@@ -88,10 +88,14 @@ func fetchBlogPost(forId id: BlogPost.Id) -> BlogPost? {
...
 private func sendNewBlogPostEmails<I>(
-  _ conn: Conn<I, Tuple4<Database.User, BlogPost, NewBlogPostFormData, Bool>>
+  _ conn: Conn<I, Tuple4<Database.User, BlogPost, NewBlogPostFormData?, Bool>>
   ) -> IO<Conn<I, Prelude.Unit>> {
 
-  let (_, post, formData, isTest) = lower(conn.data)
+  let (_, post, optionalFormData, isTest) = lower(conn.data)
+
+  guard let formData = optionalFormData else {
+    return pure(conn.map(const(unit)))
```
mbrandonw: should we send an error email to use if this happens?

PR #374, opened 2018-12-20
Remove Coupon on Subscription Change

stephencelis:
Stripe doesn't prorate coupons by their duration, and we don't support coupons for team accounts, so we should remove coupons during upgrades with a warning message.


Tests/PointFreeTests/StripeTests.swift
```diff
@@ -334,6 +334,11 @@ final class StripeTests: TestCase {
       as: .raw,
       named: "fetch-subscription"
     )
+    assertSnapshot(
+      matching: PointFree.fetchUpcomingInvoice("cus_test").rawValue,
```
stephencelis: Forgot this test in an earlier PR.

Sources/PointFree/Account/Change.swift
```diff
@@ -206,7 +206,15 @@ private let titleRowView = View<Stripe.Subscription> { subscription in
             ? ""
             : " Reactivate your subscription by submitting the form below."
           ]),
-        ])
+        ]
+        + (
+          subscription.discount?.coupon.valid == .some(true)
+            ? [
+                strong([text("Please note that changes to your subscription will remove your current discount.")])
```
mbrandonw: maybe we should put this in a `div` with some padding and a `Colors.yellow` background to pop a lil more?

Tests/PointFreeTests/__Snapshots__/StripeTests/testRequests.update-subscription.txt
```diff
@@ -2,4 +2,4 @@ POST https://api.stripe.com/v1/subscriptions/sub_test?expand%5B%5D=customer
 Authorization: Basic c2tfdGVzdDo=
 Idempotency-Key: DEADBEEF-DEAD-BEEF-DEAD-BEEFDEADBEEF
 
-items[0][id]=si_test&items[0][plan]=individual-yearly&items[0][quantity]=1
\ No newline at end of file
+coupon=&items[0][id]=si_test&items[0][plan]=individual-yearly&items[0][quantity]=1
```
mbrandonw: love these snapshot tests

PR #382, opened 2019-01-06
Prep work for episode 42

mbrandonw:
* Made a helper function that allows us to derive episode references from past episodes. Ideally there would be a special treatment for "related episodes", but for now we can just leverage the references section.
* Added a reference for railway oriented programming.

### DIFF
--- a/Sources/PointFree/Episode/Episode.swift
+++ b/Sources/PointFree/Episode/Episode.swift
@@ -150,3 +150,17 @@ public struct Episode {
     }
   }
 }
+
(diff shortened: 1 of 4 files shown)


mbrandonw:
@stephencelis made a quick change to link the title of references: 
https://github.com/pointfreeco/pointfreeco/pull/382/commits/3cb2c77f5efd26a26d215f54fb67c33cdb09ca72

stephencelis:
Looks good to me 👍

PR #383, opened 2019-01-08
Fix error in 004 transcript

timMorrill:
While reading through the transcript of this episode I noticed this sentence had an error in it.


stephencelis:
👀 Nice catch! Thanks!

PR #386, opened 2019-01-12
Less qualification of functions

mbrandonw:
We have a bunch of `Styleguide.` and `Html.` qualifications that are not necessary. I think we added these while updating to swift-html and the compiler was just pointing to all types of lines as being problematic when they really weren't.


mbrandonw:
worked through some circle problems too. should fix master when this merges.

PR #387, opened 2019-01-13
swift-html Fragment Node

stephencelis:
This is a branch to play around with given the fragment node we're testing in swift-html.

Compile times seem on par: a little slower on Travis and a little faster on Circle. So the main thing is error and overload diagnostics (which are still mixed, given that we haven't ditched any of the overloads). I'll flag some interesting things in a review.


Sources/PointFree/Episode/Show.swift
```diff
@@ -628,75 +622,55 @@ private let transcriptView = View<([Episode.TranscriptBlock], Bool)> { blocks, i
...
-                  " so that you get access to this episodes and more!",
-                  ]
-              )
-            ]
-          )
-        ]
+  return gridRow(
+    gridColumn(
+      sizes: [.mobile: 12],
+      [style(margin(leftRight: .auto))] as [Attribute<Tag.Div>],
```
stephencelis: Not sure why this annotation is required. But the compiler just errors for me otherwise.

Sources/PointFree/Episode/Show.swift
```diff
@@ -896,25 +826,22 @@ let transcriptBlockView = View<Episode.TranscriptBlock> { block -> Node in
...
-      ]
+      video(
+        [
+          `class`([innerVideoContainerClass]),
+          controls(true),
+          playsinline(true),
+          autoplay(false),
+          Html.poster(poster),
+          style(objectFit(.cover))
+        ] as [Attribute<Tag.Video>],
```
stephencelis: Same here. Annotation required :/

Sources/PointFree/Episode/Show.swift
```diff
@@ -896,25 +826,22 @@ let transcriptBlockView = View<Episode.TranscriptBlock> { block -> Node in
...
+        ChildOf.fragment(sources.map { source(src: $0) })
```
stephencelis: Also needed the additional `ChildOf` here to disambiguate for some reason. The prefix `...` didn't have that problem, strangely enough.

PR #388, opened 2019-01-17
Shorten the default timeout to 30 seconds.

mbrandonw:
The default is 60 seconds which is pretty long. So let's try this out.

### DIFF
--- a/Sources/PointFree/TODO.swift
+++ b/Sources/PointFree/TODO.swift
@@ -49,10 +49,14 @@ public func requireSome<A>(
 
 // TODO: Move to PreludeFoundation?
 
+private let sessionConfig = URLSessionConfiguration.default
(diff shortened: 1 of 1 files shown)


Sources/PointFree/TODO.swift
```diff
@@ -49,10 +49,14 @@ public func requireSome<A>(
 
 // TODO: Move to PreludeFoundation?
 
+private let sessionConfig = URLSessionConfiguration.default
+  |> \.timeoutIntervalForRequest .~ 30
+  |> \.timeoutIntervalForResource .~ 30
```
mbrandonw: maybe this should be 25 since Heroku's timeout is 30, and so we prob won't have enough time to recover.

stephencelis: yeah let’s do it

PR #390, opened 2019-02-05
Fix modify subscription javascript.

mbrandonw:
If you change the number of team members when modify a subscription it does not live update the total $, and so leads to a confusing user experience.


mbrandonw:
Updating swift-web and swift-html caused a few snapshot failures that I also fixed.

mbrandonw:
yeah works in both. I think the main problem here is that there's a lot of logic duplicated in the pricing page and the modify subscription page, so things may have been updated in one place but not the other.

PR #391, opened 2019-02-06
Send an admin email if we don't invoke a callback in dataTask.

mbrandonw:
Just want to rule out a possibility for the strange errors we are seeing in prod.


Sources/PointFree/TODO.swift
```diff
@@ -56,16 +56,46 @@ private let sessionConfig = URLSessionConfiguration.default
 public func dataTask(with request: URLRequest) -> EitherIO<Error, (Data, URLResponse)> {
   return .init(
     run: .init { callback in
+
+      let startTime = Current.date().timeIntervalSince1970
+      Current.logger.debug("[Data Task Started] \(request.httpMethod ?? "UNKNOWN") \(request.url?.absoluteString ?? "nil")")
```
mbrandonw: added some logging to help us debug.

PR #395, opened 2019-02-14
Fix spacing

stephencelis:
Seems like multiline literals have some gotchas to consider.


Tests/PointFreeTests/AccountTests/AccountTests.swift
```diff
@@ -15,7 +15,7 @@ final class AccountTests: TestCase {
   override func setUp() {
     super.setUp()
     update(&Current, \.database .~ .mock)
-//    record = true
+    record = true
```
mbrandonw: comment this out?

```suggestion
//    record = true
```

PR #396, opened 2019-02-18
Support solutions for exercises

mbrandonw:
Adds a `solution` field to `Episode.Exercise` for us to be able to add solutions to exercises, and exposes it in the episode with a `details`/`summary` tag.

From the tests:

image

### Todo for later

* Solve a bunch of exercises. Figuring we can just do a handful per week to slowly backfill.

* Improve the rendering of markdown code blocks in exercises, it's pretty gross right now:

image

* All of my account screenshots changed very slightly, sometimes just a few pixels, other times no discernible change at all. Are our configurations different, or did some screenshots not get updated from #395?


Sources/PointFree/Episode/Episode.swift
```diff
@@ -95,10 +95,12 @@ public struct Episode {
   }
 
   public struct Exercise {
-    public private(set) var body: String
+    public var problem: String
```
mbrandonw: renamed `body` to `problem`. caused a lot of churn, but think it makes more sense. I've got the changes in the transcripts directory in a branch: https://github.com/mbrandonw/episode-transcripts/compare/model-update?expand=1

stephencelis:
Awesome!

> All of my account screenshots changed very slightly, sometimes just a few pixels, other times no discernible change at all. Are our configurations different, or did some screenshots not get updated from #395?

I wonder if dark mode or other Mojave settings cause screen shots to be slightly different...I've noticed my shots slightly change the existing ones as well...

Sources/PointFree/Episode/Show.swift
```diff
@@ -811,19 +811,45 @@ private let exercisesView = View<[Episode.Exercise]> { exercises -> [Node] in
...
+    ]
+  )
+}
+
+private func solution(to exercise: Episode.Exercise) -> [Node] {
+  guard let solution = exercise.solution else { return [] }
+
+  return [
+    details([
+      summary(["Solution"]).rawValue, // todo: update details to take children of details
```
mbrandonw: seems that `details` doesn't take `ChildOf<Tag.Details>`, but `summary` does return that. 

We can't fix this immediately because we still need to update to the newest swift-html in this repo, but something to come back to.

Tests/PointFreeTests/EpisodeTests.swift
```diff
@@ -398,19 +398,24 @@ class EpisodeTests: TestCase {
...
     #if !os(Linux)
     if #available(OSX 10.13, *), ProcessInfo.processInfo.environment["CIRCLECI"] == nil {
-      assertSnapshots(
-        matching: conn |> siteMiddleware,
-        as: [
-          "desktop": .ioConnWebView(size: .init(width: 1100, height: 1600)),
-          "mobile": .ioConnWebView(size: .init(width: 500, height: 1900))
-        ]
-      )
+      let webView = WKWebView(frame: .init(x: 0, y: 0, width: 1100, height: 1600))
```
mbrandonw: creating this web view explicitly so that I can execute JS in it for opening the details of one of the exercises.

Sources/PointFree/Episode/Show.swift
stephencelis: Hm it's a bit more complicated. This is valid:

``` html
<details>
  <summary>Solution</summary>
  <p>Lorem</p>
  <p>Ipsum</p>
</details>
```

Maybe we change `details` to include a `summary:` argument and change `summary` to return a plain ole `Node` or remove the helper entirely.

``` swift
details(
  summary: "Solution",
  p("Lorem"),
  p("Ipsum")
)
```

mbrandonw: oh yeah, maybe we just remove the restriction on summary then

PR #398, opened 2019-02-19
More modules

stephencelis:
This begins to modularize things a bit more. Added:

- `Logger`: Extracted logging to its own module.
- `PointFreePrelude`: A collection of utility stuff that could move further up the chain (to swift-web, swift-prelude, etc.), but for now is a convenient dumping ground. Previously was all in `TODO.swift`.
- `GitHub`: The GitHub API client.

Also stubbed out `Stripe` and `Database` modules.


.sourcery-templates/DerivePartialIsos.stencil
```diff
@@ -1,5 +1,6 @@
 import ApplicativeRouter
 import Either
+import PointFreePrelude
```
stephencelis: This generated file needs access to `EmailAddress` to work. Not sure how modular generated Isos should be distributed yet...

Package.swift
```diff
@@ -20,6 +25,34 @@ let package = Package(
...
+        "Either",
+        "GitHub",
+        "PostgreSQL",
+        "Prelude",
+        "Stripe",
+        ]
+    ),
+
+    .target(
+      name: "GitHub",
```
stephencelis: - [ ] Make `GitHubTests` too.

Sources/GitHub/GitHub.swift
```diff
@@ -1,23 +1,28 @@
...
   public var fetchUser: (AccessToken) -> EitherIO<Error, User>
+}
 
-  static let live = GitHub(
-    fetchAuthToken: PointFree.fetchAuthToken >>> runGitHub,
-    fetchEmails: PointFree.fetchEmails >>> runGitHub,
-    fetchUser: PointFree.fetchUser >>> runGitHub
-  )
+extension GitHub {
+  public init(clientId: String, clientSecret: String) {
```
stephencelis: GitHub clients are now convenience-initialized with their dependencies.

Sources/GitHub/GitHub.swift
```diff
@@ -59,40 +64,48 @@ public struct GitHub {
   public struct UserEnvelope: Codable {
     public private(set) var accessToken: AccessToken
     public private(set) var gitHubUser: User
+
+    public init(accessToken: AccessToken, gitHubUser: User) {
```
stephencelis: Need a public initializer to rebuild this.

Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -81,7 +82,7 @@ let updateProfileMiddleware =
         updateFlash = flash(.notice, "We’ve updated your profile!")
       }
 
-      let updateCustomerExtraInvoiceInfo = zip2(
+      let updateCustomerExtraInvoiceInfo = zip(
```
stephencelis: We only have a single overload on Optional...if this is an issue we can go back to numbering.

Sources/PointFree/Bootstrap.swift
```diff
@@ -51,7 +53,12 @@ private let loadEnvVars = { (_: Prelude.Unit) -> EitherIO<Error, Prelude.Unit> i
     .flatMap { try? decoder.decode(EnvVars.self, from: $0) }
     ?? Current.envVars
 
-  update(&Current, \.envVars .~ envVars)
+  Current.envVars = envVars
+  Current.gitHub = GitHub(
```
stephencelis: This part is slightly clunky. We manually have to do this work on bootstrap, but at least the side effect is localized...

Sources/PointFree/Database.swift
```diff
@@ -301,7 +303,7 @@ private func fetchSubscription(id: Database.Subscription.Id) -> EitherIO<Error,
     ORDER BY "created_at" DESC
     LIMIT 1
     """,
-    [id.rawValue.rawValue]
+    [id.rawValue.uuidString]
```
stephencelis: I removed the `RawRepresentable` conformance I think I added to UUID. Seems like a bad idea now.

Sources/PointFree/Emails/Mailgun.swift
```diff
@@ -78,7 +77,7 @@ private func mailgunSend(email: Email) -> EitherIO<Error, Mailgun.SendEmailRespo
     |> \.allHTTPHeaderFields %~ attachedMailgunAuthorization
     |> \.httpBody .~ Data(urlFormEncode(value: params).utf8)
 
-  return jsonDataTask(with: request)
+  return jsonDataTask(with: request, logger: Current.logger)
```
stephencelis: Making logger a dependency of these helpers, but we really should just add logging as a bracket around IO...maybe we can't.

Sources/PointFree/Environment.swift
```diff
@@ -1,22 +1,25 @@
...
-  public private(set) var envVars = EnvVars()
-  public private(set) var episodes = { [Episode]() }
-  public private(set) var features = [Feature].allFeatures
-  public private(set) var gitHub = GitHub.live
-  public private(set) var logger = Logger()
-  public private(set) var mailgun = Mailgun.live
-  public private(set) var renderHtml: ([Node]) -> String = Html.render
-  public private(set) var stripe = Stripe.live
-  public private(set) var uuid: () -> UUID = UUID.init
+  public var assets = Assets()
```
stephencelis: Preemptively removing this since it won't work in Swift 5.

Sources/PointFree/Environment.swift
```diff
@@ -1,22 +1,25 @@
...
+  public var blogPosts = allBlogPosts
+  public var cookieTransform = CookieTransform.encrypted
+  public var database = Database.live
+  public var date: () -> Date = Date.init
+  public var envVars = EnvVars()
+  public var episodes = { [Episode]() }
+  public var features = [Feature].allFeatures
+  public var gitHub = GitHub(clientId: "", clientSecret: "")
```
stephencelis: Not a great default, but...

Sources/PointFree/Account/UpdateProfile.swift
mbrandonw: we should also prob just bring in overture. I'd like to start using it more anyway. I'll add a Trello card

Sources/PointFree/Database.swift
mbrandonw: haha yeah

Sources/PointFreePrelude/EmailAddress.swift
```diff
@@ -0,0 +1,3 @@
+import Tagged
+
+public typealias EmailAddress = Tagged<(email: (), String), String>
```
mbrandonw: ha maybe this should just be in tagged

.sourcery-templates/DerivePartialIsos.stencil
mbrandonw: yeah, I bet the command line tool for generating partial isos will need to take an argument for extra imports or something. I don't think SwiftSyntax can give us module names of types.

Sources/PointFree/Environment.swift
```diff
@@ -1,22 +1,25 @@
...
-  public private(set) var uuid: () -> UUID = UUID.init
+  public var assets = Assets()
+  public var blogPosts = allBlogPosts
+  public var cookieTransform = CookieTransform.encrypted
+  public var database = Database.live
+  public var date: () -> Date = Date.init
+  public var envVars = EnvVars()
+  public var episodes = { [Episode]() }
+  public var features = [Feature].allFeatures
+  public var gitHub = GitHub.Client(clientId: "", clientSecret: "")
```
mbrandonw: I guess one way around this is to make `var Current: Environment!` and we can just construct it in the bootstrap?

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -199,7 +200,7 @@ private let routers: [Router<Route>] = [
     <¢> post %> lit("invites") %> formField("email", Optional.iso.some >>> opt(.rawRepresentable)) <% end,
 
   .invite <<< .show
-    <¢> get %> lit("invites") %> pathParam(.rawRepresentable >>> .rawRepresentable) <% end,
+    <¢> get %> lit("invites") %> pathParam(.uuid >>> .rawRepresentable) <% end,
```
mbrandonw: I guess this could also be `.uuid >>> .tagged` which might read a bit nicer.

```suggestion
    <¢> get %> lit("invites") %> pathParam(.uuid >>> .tagged) <% end,
```

PR #401, opened 2019-02-20
Remove SiteTag

stephencelis:
We can revisit in the future.


Sources/PointFree/Tag.swift
```diff
@@ -1,69 +0,0 @@
...
-        Class.pf.colors.bg.white,
-        Class.type.textDecorationNone,
-        Class.pf.colors.border.gray900
-        ])
-    ],
-    [.text(tag.name)]
-  )
-}
-
-func slug(for string: String) -> String {
```
mbrandonw: this is used elsewhere so should be moved out. I know cause I had to move it with some of my current work

mbrandonw:
there were some ambiguous uses of `assertSnapshot`, and while I don't understand why that would be happening we plan on updating snapshot testing to 1.2.0 anyway, so I've done that. if the build goes green I'll merge.

PR #402, opened 2019-02-21
Extract some models to a module

mbrandonw:
### What

This extracts `Episode`, `BlogPost`, and all transcripts to a `Models` module. With this done we should be able to start extracting the database and router code into their own modules.


.gitignore
```diff
@@ -71,6 +71,6 @@ Packages/*
 *.xcodeproj/*
 Package.resolved
 .env
-Sources/PointFree/Transcripts
+Sources/Models/Transcripts
```
mbrandonw: Moved the transcripts submodule to the Models directory

.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "Sources/PointFree/Transcripts"]
-	path = Sources/PointFree/Transcripts
+[submodule "Sources/Models/Transcripts"]
+	path = Sources/Models/Transcripts
```
mbrandonw: after pulling this you'll prob have to do some submodule clean up on yer end :/

Package.swift
```diff
@@ -21,11 +22,13 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("8cbc934")),
     .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", from: "1.1.0"),
     .package(url: "https://github.com/pointfreeco/swift-html.git", from: "0.2.1"),
+    .package(url: "https://github.com/pointfreeco/swift-tagged.git", from: "0.2.0"),
```
mbrandonw: the Models package will need tagged for id's

Sources/Models/BlogPosts/BlogPost0001_Announcement.swift
```diff
@@ -10,7 +10,7 @@ even deeper into functional programming, showing real world use cases and more!
   contentBlocks: [
     .init(
       content: """
-      It's hard to believe it's been nearly 3 months since we launched [Point-Free](\(url(to: .home))), but
+      It's hard to believe it's been nearly 3 months since we launched [Point-Free](/), but
```
mbrandonw: since the router no longer lives in the same module as the models we can't use it, so I've hard coded the url's.

I don't think this is a huge deal since 1) these url's shouldn't change, we should only add new url's, and 2) eventually we probably want to move to have transcripts as markdown that we parse rather than embedded in swift files, and in that case we wouldn't have access to the router anyway.

Sources/Models/BlogPosts/BlogPost0020_PodcastRSS.swift
```diff
@@ -21,18 +21,31 @@ favorite podcast app!
...
+""",
+      timestamp: nil,
+      type: .paragraph
+    ),
+    .init(
+      content: """
+Since the time we launched this feature we have changed this slightly. Only yearly subscribers get access to the entire catalogue of episodes in their podcast RSS feed, and monthly subscribers get access to the last 4 episodes.
+""",
+      timestamp: nil,
+      type: .correction
```
mbrandonw: went ahead and added this correction to the blog post while I was here:

image

Sources/PointFree/Blog/BlogPost.swift
```diff
@@ -1,98 +1,21 @@
...
-    public var twitterUrl: String {
-      switch self {
-      case .brandon:
-        return PointFree.twitterUrl(to: .mbrandonw)
-      case .pointfree:
-        return PointFree.twitterUrl(to: .pointfreeco)
-      case .stephen:
-        return PointFree.twitterUrl(to: .stephencelis)
-      }
+extension BlogPost.Author {
```
mbrandonw: most things in this file moved to `Models` except for some stuff that had  to stay in `PointFree`

Sources/Models/BlogPosts/BlogPost0004_OvertureSetters.swift
```diff
@@ -20,7 +20,7 @@ transformations out of smaller units.
 
 > This week's [episode](https://www.pointfree.co/episodes/ep15-setters-ergonomics-performance) explored providing a friendlier API to functional setters, and improved
 their performance by leveraging Swift's value mutation semantics. To make these ideas accessible to everyone
-we have updated our [Swift Overture](\(gitHubUrl(to: .repo(.overture)))) library to add these
+we have updated our [Swift Overture](https://www.github.com/pointfreeco/swift-overture) library to add these
```
stephencelis: I think "www" redirects, but we may want to use plain ole "github.com".

mbrandonw: true true, done

PR #405, opened 2019-02-26
Extract database to module

mbrandonw:
### What

Extracts all the database logic into its own module, which will help us to build feature specific modules soon.

I have a few TODO's left and I need to get tests building. Might need some help with tests...


Sources/Database/Database.swift
```diff
@@ -1 +1,845 @@
...
+  fileprivate func execute(
+    _ query: String,
+    _ representable: [PostgreSQL.NodeRepresentable] = []
+    ) -> EitherIO<E, PostgreSQL.Node> {
+
+      return self.flatMap { conn in
+        return EitherIO<E, PostgreSQL.Node>.wrap { () -> Node in
+          //let uuid = UUID().uuidString
+          //let startTime = Date().timeIntervalSince1970
+          // TODO: logger
```
mbrandonw: need to thread the logger through. just a note to myself...

Sources/PointFree/Account/Account.swift
```diff
@@ -842,15 +843,15 @@ let blockSelectClass =
...
+  let currentUser: User
+  let emailSettings: [EmailSetting]
+  let episodeCredits: [EpisodeCredit]
   let stripeSubscription: Stripe.Subscription?
   let subscriberState: SubscriberState
-  let subscription: Database.Subscription?
-  let subscriptionOwner: Database.User?
-  let teamInvites: [Database.TeamInvite]
-  let teammates: [Database.User]
+  let subscription: Models.Subscription?
```
mbrandonw: we talked about this irl but just for posterity: we discussed changing `Stripe.Susbcription` to StripeSubscription` so that we could then remove these `Models` qualifications.

Sources/PointFree/Routers/PointFreeRoutes.swift
```diff
@@ -75,23 +75,24 @@ public enum Route: DerivePartialIsos, Equatable {
...
   public enum Team: DerivePartialIsos, Equatable {
     case leave
-    case remove(Database.User.Id)
+    case remove(User.Id)
   }
 
   public enum Webhooks: DerivePartialIsos, Equatable {
-    case stripe(Stripe)
+    // TODO: what to do about this _
+    case stripe(_Stripe)
```
mbrandonw: that would also help with this. I needed to make this `_Stripe`.

PR #406, opened 2019-03-01
Set timeout on URLRequest too.

mbrandonw:
Just curious if this matters at all.

Starting to think maybe we bring back the `requestTimeout` middleware transformer...

### DIFF
--- a/Sources/PointFreePrelude/URLRequest.swift
+++ b/Sources/PointFreePrelude/URLRequest.swift
@@ -40,9 +40,10 @@ public let attachFormData =
     >>> Data.init(_:)
     >>> set(\URLRequest.httpBody)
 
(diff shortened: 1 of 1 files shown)


Sources/PointFreePrelude/URLRequest.swift
```diff
@@ -57,6 +58,7 @@ public func dataTask(
       logger?.debug("[Data Task] \(uuid) \(request.url?.absoluteString ?? "nil") \(request.httpMethod ?? "UNKNOWN")")
 
       let session = URLSession(configuration: sessionConfig)
+      let request = request |> \.timeoutInterval .~ timeoutInterval
```
stephencelis: The default is 60, so not sure it matters much.

PR #414, opened 2019-03-06
Extract About Views

stephencelis:
Just a WIP for extracting views into their own modules. Biggest issue will be feeding the base URI through to every middleware that needs to send email.


Package.swift
```diff
@@ -302,5 +305,39 @@ let package = Package(
       dependencies: [
         "Syndication",
         ]),
+
+    .target(
+      name: "Views",
+      dependencies: [
```
stephencelis: - [ ] These dependencies are eager based on `PointFree`'s. Pare down with just what's needed before merging.

Sources/PointFreeRouter/Routes.swift
```diff
@@ -1,10 +1,27 @@
...
 import Prelude
 import Stripe
 
+public enum EncryptedTag {}
+public typealias Encrypted<A> = Tagged<EncryptedTag, A>
+
+extension Tagged where Tag == EncryptedTag, RawValue == String {
+  public init?(_ text: String, with secret: String) {
+    guard
+      let string = encrypted(text: text, secret: secret)
```
mbrandonw: since we're going with removing all encryption stuff from the routes, we could move all of this back to `PointFree` and remove the `HttpPipeline` dependency from `PointFreeRouter`. would speed up cold builds for the router a lot.

Sources/PointFree/__Generated__/DerivedPartialIsos.swift
```diff
@@ -0,0 +1,12 @@
+// Generated using Sourcery 0.15.0 — https://github.com/krzysztofzablocki/Sourcery
```
mbrandonw: oh ha forgot to delete this (and the corresponding makefile commands) after everything was moved to PointFreeRouter.

Tests/PointFreeTests/__Snapshots__/NewslettersTests/testExpressUnsubscribeReply_IncorrectSignature.1.Conn.txt
```diff
@@ -3,1999 +3,10 @@ Cookie: pf_session={}
 
 recipient=unsub-f9c46e50cb32c3f12369e92c8bb9d9db09edf2cce5a0307b4e8516ac36340b470d3fed478e73515f3ad396db9cb71d21c0abe5f9872e3e792de15f4b1ae5795e%40pointfree.co&sender=hello%40pointfree.co&signature=this%20is%20an%20invalid%20signature&timestamp=1517356800&token=deadbeef
 
-404 Not Found
-Content-Length: 45057
-Content-Type: text/html; charset=utf-8
+406 Not Acceptable
```
mbrandonw: for posterity this is a request that was previously 404'ing because of failed decryption in the router, but not the request gets to the middleware and we `406` it when decryption fails.

PR #416, opened 2019-03-12
Functional CSS module

mbrandonw:
### What

I split off all the library level, functional css stuff into its own module. This could potentially be split into its own repo someday as it has no PF specific stuff, just all of the atomic classes and styles.

I also moved `Components.swift` to `Styleguide`, which contains some common html fragments we use (like for GitHub and twitter buttons).

I did this on the flight last night...


Sources/PointFree/Account/Invite.swift
```diff
@@ -181,7 +182,11 @@ private let showInviteLoggedOutView = View<(TeamInvite, User)> { invite, inviter
           ]),
 
         p([`class`([Class.padding([.mobile: [.top: 3]])])], [
-          gitHubLink(text: "Login with GitHub", type: .black, redirectRoute: .invite(.show(invite.id)))
+          gitHubLink(
+            text: "Login with GitHub",
+            type: .black,
+            redirect: url(to: .invite(.show(invite.id)))
```
mbrandonw: previous this `gitHubLink` helper could take the redirect route, but now it needs the redirect url since it's been put in the `Styleguide` module and no longer has access to the router. I guess I could have made a `gitHubLink` overload in `PointFree` that takes a route...

PR #420, opened 2019-03-13
Tag Safely

stephencelis:
A recent refactor almost led to a bug because we hadn't tagged things like our app secret and API keys. Luckily a snapshot test saved us, but let's make things even safer!

### DIFF
--- a/Sources/Mailgun/Client.swift
+++ b/Sources/Mailgun/Client.swift
@@ -10,7 +10,7 @@ import Tagged
 import UrlFormEncoding
 
 public struct Client {
-  public typealias ApiKey = MailgunForwardPayload.ApiKey
+  public typealias ApiKey = Tagged<(Client, apiKey: ()), String>
   public typealias Domain = Tagged<(Client, domain: ()), String>
 
   private let appSecret: String
@@ -72,6 +72,14 @@ public struct Client {
 
         return (userId, newsletter)
     }
+  }
+
+  public func verify(payload: MailgunForwardPayload, with apiKey: ApiKey) -> Bool {
+    let digest = hexDigest(
+      value: "\(payload.timestamp)\(payload.token)",
+      asciiSecret: apiKey.rawValue
+    )
+    return payload.signature == digest
   }
 }
 
--- a/Sources/Mailgun/Client.swift
+++ b/Sources/Mailgun/Client.swift
@@ -13,12 +13,12 @@ public struct Client {
   public typealias ApiKey = Tagged<(Client, apiKey: ()), String>
   public typealias Domain = Tagged<(Client, domain: ()), String>
 
-  private let appSecret: String
+  private let appSecret: AppSecret
 
   public var sendEmail: (Email) -> EitherIO<Error, SendEmailResponse>
 
   public init(
-    appSecret: String,
+    appSecret: AppSecret,
     sendEmail: @escaping (Email) -> EitherIO<Error, SendEmailResponse>) {
     self.appSecret = appSecret
     self.sendEmail = sendEmail
@@ -26,7 +26,7 @@ public struct Client {
 
   public init(
     apiKey: ApiKey,
-    appSecret: String,
+    appSecret: AppSecret,
     domain: Domain,
     logger: Logger) {
     self.appSecret = appSecret
@@ -43,7 +43,7 @@ public struct Client {
 
     guard let payload = encrypted(
       text: "\(userId.rawValue.uuidString)\(boundary)\(newsletter.rawValue)",
-      secret: self.appSecret
+      secret: self.appSecret.rawValue
       ) else { return nil }
 
     return .init(rawValue: "unsub-\(payload)@pointfree.co")
@@ -62,7 +62,7 @@ public struct Client {
       .map(String.init)
 
     return payload
-      .flatMap { decrypted(text: $0, secret: self.appSecret) }
+      .flatMap { decrypted(text: $0, secret: self.appSecret.rawValue) }
       .map { $0.components(separatedBy: boundary) }
       .flatMap { components in
         guard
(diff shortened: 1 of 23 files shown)


Sources/Mailgun/Client.swift
```diff
@@ -6,9 +6,13 @@ import Models
 import Optics
 import PointFreePrelude
 import Prelude
+import Tagged
 import UrlFormEncoding
 
 public struct Client {
+  public typealias ApiKey = MailgunForwardPayload.ApiKey
```
stephencelis: Because `Mailgun` depends on `Model`s, the tag needs to be there.

Sources/PointFree/Auth.swift
```diff
@@ -221,7 +221,7 @@ private func refreshStripeSubscription(for user: Models.User) -> EitherIO<Error,
 private func gitHubAuthorizationUrl(withRedirect redirect: String?) -> String {
   return gitHubUrl(
     to: .authorize(
-      clientId: Current.envVars.gitHub.clientId,
+      clientId: Current.envVars.gitHub.clientId.rawValue,
```
stephencelis: I wanted to put `GitHub.Client.Id` directly in the route, but adding `import GitHub` to `DerivePartialIsos.swift` leads to `GitHub.User`/`Models.User` ambiguity :(

Not sure if there's an easy fix in Sourcery to prefix types with their module names, or maybe we need to stop overlapping and always prefix our types (`GitHubUser`, etc.).

Sources/PointFree/Bootstrap.swift
```diff
@@ -67,7 +67,7 @@ private let loadEnvVars = { (_: Prelude.Unit) -> EitherIO<Error, Prelude.Unit> i
   )
   Current.mailgun = .init(
     apiKey: Current.envVars.mailgun.apiKey,
-    appSecret: Current.envVars.appSecret,
+    appSecret: Current.envVars.appSecret.rawValue,
```
stephencelis: `Mailgun` doesn't know about `EnvVars`, so escape-hatching the initializer here.

Sources/PointFree/TODO.swift
```diff
@@ -61,19 +61,6 @@ public func requireSome<A>(
...
 
-extension PartialIso where A == String, B == String {
-  public static func decrypted(withSecret secret: String) -> PartialIso<String, String> {
-    return PartialIso(
-      apply: { HttpPipeline.decrypted(text: $0, secret: secret) },
-      unapply: { encrypted(text: $0, secret: secret) }
-    )
-  }
-
-  public static var appDecrypted: PartialIso<String, String> {
```
stephencelis: Was going to convert this to use `Encrypted<String>`, but we weren't using this.

Sources/PointFree/Auth.swift
mbrandonw: I think `GitHubUser` is good!

Sources/PointFree/EnvVars.swift
```diff
@@ -1,8 +1,25 @@
...
+
+public typealias MailgunApiKey = Mailgun.Client.ApiKey
+public typealias MailgunDomain = Mailgun.Client.Domain
+
+public typealias StripeEndpointSecret = Stripe.Client.EndpointSecret
+public typealias StripePublishableKey = Stripe.Client.PublishableKey
+public typealias StripeSecretKey = Stripe.Client.SecretKey
 
 public struct EnvVars: Codable {
+  public typealias Secret = Tagged<EnvVars, String>
```
mbrandonw: I'd be down to even call this `AppSecret`

Sources/Mailgun/Client.swift
mbrandonw: do you think it would make sense to move `verify(with:)` from `MailgunForwardPayload.swift` to this file? Then `MailgunForwardPayload` wouldn't even need an `ApiKey` I don't think...

Sources/PointFree/Bootstrap.swift
mbrandonw: could we move the tagged `AppSecret` to models so that everyone gets access to it?

Sources/PointFree/EnvVars.swift
```diff
@@ -1,8 +1,25 @@
...
+
+public typealias GitHubClientId = GitHub.Client.Id
+public typealias GitHubClientSecret = GitHub.Client.Secret
+
+public typealias MailgunApiKey = Mailgun.Client.ApiKey
+public typealias MailgunDomain = Mailgun.Client.Domain
+
+public typealias StripeEndpointSecret = Stripe.Client.EndpointSecret
+public typealias StripePublishableKey = Stripe.Client.PublishableKey
+public typealias StripeSecretKey = Stripe.Client.SecretKey
```
mbrandonw: I think I'd even be fine dropping this type aliases. do you think they help clean things up?

stephencelis: Can't unfortunately. We have `Stripe`, etc., structs inside `EnvVars` and it doesn't know how to disambiguate.

mbrandonw: arg. should we just go all out with `StripeClient`, `MailGunClient` and `GitHubClient`?

Sources/Mailgun/Client.swift
```diff
@@ -6,9 +6,13 @@ import Models
...
 import PointFreePrelude
 import Prelude
+import Tagged
 import UrlFormEncoding
 
 public struct Client {
+  public typealias ApiKey = Tagged<(Client, apiKey: ()), String>
+  public typealias Domain = Tagged<(Client, domain: ()), String>
+
   private let appSecret: String
```
mbrandonw: I think this can now be `AppSecret` right?

Sources/PointFree/EnvVars.swift
stephencelis: yeah...I'm thinking we need a renaming PR that rethinks any potential collision :/

mbrandonw: yeah we can do that later, this looks good tho!

PR #421, opened 2019-03-14
Add logging on unrecognized Stripe hooks

stephencelis:
A few have snuck in recently so let's start troubleshooting.

### DIFF
--- a/Sources/PointFreeRouter/Routes.swift
+++ b/Sources/PointFreeRouter/Routes.swift
@@ -69,7 +69,7 @@ public enum Route: DerivePartialIsos, Equatable {
 
     public enum _Stripe: DerivePartialIsos, Equatable {
       case event(Event<Either<Invoice, Stripe.Subscription>>)
-      case `fallthrough`
+      case `fallthrough`(Event<Prelude.Unit>)
     }
   }
 }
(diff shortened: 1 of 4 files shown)


Sources/PointFreeRouter/Routes.swift
```diff
@@ -190,7 +190,13 @@ private let routers: [Router<Route>] = [
     <% end,
 
   .webhooks <<< .stripe <<< .fallthrough
-    <¢> post %> lit("webhooks") %> lit("stripe") <% end,
+    <¢> post %> lit("webhooks") %> lit("stripe")
+    %> jsonBody(
```
mbrandonw: in a follow up let's also make this optional in case stripe is just sending bad json data

Sources/PointFreeRouter/Routes.swift
```diff
@@ -68,8 +68,9 @@ public enum Route: DerivePartialIsos, Equatable {
     case stripe(_Stripe)
 
     public enum _Stripe: DerivePartialIsos, Equatable {
-      case event(Event<Either<Invoice, Stripe.Subscription>>)
-      case `fallthrough`
+      case knownEvent(Event<Either<Invoice, Stripe.Subscription>>)
+      case unknownEvent(Event<Prelude.Unit>)
+      case fatal
```
mbrandonw: nice nice I like that

PR #422, opened 2019-03-15
Bundle base URL with the router in the PointFreeRouter module

mbrandonw:
This allows us to bundle the base URL with the router in the `PointFreeRouter`, which will allow us to extract views into their own module more easily. Once that is done we'll work on a reader/context based way of threading base URL through the view.

### DIFF
--- a/Sources/PointFreeRouter/PointFreeRouter.swift
+++ b/Sources/PointFreeRouter/PointFreeRouter.swift
@@ -1,0 +1,35 @@
+import ApplicativeRouter
+import Foundation
+
+public struct PointFreeRouter {
+  public let baseUrl: URL
+  public let router: Router<Route>
+
+  public init(baseUrl: URL = URL(string: "http://localhost:8080")!) {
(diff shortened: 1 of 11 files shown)


Sources/PointFreeRouter/PointFreeRouter.swift
```diff
@@ -0,0 +1,35 @@
...
+public struct PointFreeRouter {
+  public let baseUrl: URL
+  public let router: Router<Route>
+
+  public init(baseUrl: URL = URL(string: "http://localhost:8080")!) {
+    self.baseUrl = baseUrl
+    self.router = routers.reduce(.empty, <|>)
+  }
+
+  public init(baseUrl: URL = URL(string: "http://localhost:8080")!, router: Router<Route>) {
```
mbrandonw: I can't do a default argument of `router: Router<Route> = routers.reduce(.empty, <|>)` since `routes` is internal :/

PR #423, opened 2019-03-15
Move a few views to the new views module

mbrandonw:
Just moving a few views over to see how it feels what problems we might have. So far it's very smooth!


Sources/PointFree/Admin/FreeEpisodes.swift
```diff
@@ -12,35 +12,12 @@ import Prelude
...
-  [
-    p([
-      .text(ep.title),
-      form([action(path(to: .admin(.freeEpisodeEmail(.send(ep.id))))), method(.post)], [
-        input([type(.submit), value("Send email!")])
-        ])
-      ])
-  ]
-}
+    >=> respond(freeEpisodeView(episodes: Current.episodes(), today: Current.date()))
```
mbrandonw: I experimented with converting this view to a function instead of `View`. it works, we just need a new `respond` middleware to deal with it, which I haven't done yet. this view didnt need any custom data from the route, so I could do it more simply by just calling the function immediately.

Sources/PointFree/Blog/BlogMiddleware.swift
```diff
@@ -21,7 +21,7 @@ func blogMiddleware(
       |> blogAtomFeedResponse
 
   case .index:
-    return conn.map(const(user .*. subscriberState .*. route .*. unit))
+    return conn.map(const(Current.blogPosts() .*. user .*. subscriberState .*. route .*. unit))
```
mbrandonw: just need to be more explicit with what data the view needs

PR #425, opened 2019-03-26
Swift 5

stephencelis:
Don't have to merge soon, but wanted to open this.


Package.swift
```diff
@@ -27,14 +27,14 @@ let package = Package(
...
-    .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .exact("1.2.0")),
+    .package(url: "https://github.com/ianpartridge/swift-backtrace.git", .exact("1.1.0")),
+    .package(url: "https://github.com/pointfreeco/Ccmark.git", .branch("master")),
     .package(url: "https://github.com/pointfreeco/swift-html.git", .exact("0.2.1")),
+    .package(url: "https://github.com/pointfreeco/swift-prelude.git", .branch("swift-5")),
+    .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", from: "1.5.0"),
     .package(url: "https://github.com/pointfreeco/swift-tagged.git", .revision("926e8e0")),
-    .package(url: "https://github.com/pointfreeco/swift-web.git", .revision("2c3d440")),
-    .package(url: "https://github.com/pointfreeco/Ccmark.git", .branch("master")),
+    .package(url: "https://github.com/pointfreeco/swift-web.git", .revision("swift-5")),
```
mbrandonw: should we merge all the deps in before merging this?

stephencelis: Yeah I think so.

Sources/PointFree/Environment.swift
```diff
@@ -18,6 +19,7 @@ public struct Environment {
   public var date: () -> Date = Date.init
   public var envVars = EnvVars()
   public var episodes = { [Episode]() }
+  public var eventLoopGroup: EventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
```
mbrandonw: we can revert this right?

stephencelis: Done!

stephencelis:
@mbrandonw I think this is finally ready to go!

Package.swift
stephencelis: Done!

PR #432, opened 2019-04-13
First pass for "tale of two flatmap" episode solutions

dannyhertz:


Sources/Models/PublicEpisodes/0010-ATaleOfTwoFlatMaps.swift
```diff
@@ -27,33 +27,130 @@ operation to other structures and derive new, useful code!
...
+```
 """),
-  Episode.Exercise(problem:
-    """
+  Episode.Exercise(
+    problem: """
 Is it possible to define `partitionMap` on `Either`?
+""",
+    solution: """
+  // Uhhhhh
```
dannyhertz: I struggled with this one as you can see. And because of that it felt like no you cannot define it on Either. Here is the signature I tried to play with but nothing really came of it. Any more guidance?

```
extension Either {

    func partitionMap<C, D>(_ transforms: ((A) -> Either<C, D>, (B) -> Either<C, D>)) -> (lefts: Either<C, D>, rights: Either<C, D>) {

    }

}
```

Sources/Models/PublicEpisodes/0010-ATaleOfTwoFlatMaps.swift
```diff
@@ -27,33 +27,130 @@ operation to other structures and derive new, useful code!
...
+  Episode.Exercise(
+    problem: """
 Define `filtered` as a function from `[A?]` to `[A]`.
-"""),
-  Episode.Exercise(problem:
-    """
+""",
+    solution: """
+```
+func filtered<A>() -> ([A?]) -> [A] {
```
stephencelis: I think we want the signature to be `([A?]) -> [A]` rather than `() -> ([A?]) -> [A]`, but I'm also unsure of if we wanna mix functions and methods. @mbrandonw do you have a preference on using methods everywhere vs. functions in these exercises?

mbrandonw: @dannyhertz yoooo did you wanna make these changes?? if you flatten these extra `() -> ` functions in a few of your solutions, I think it might be g2g!

PR #433, opened 2019-04-22
A few todos for enterprise

mbrandonw:
* Adds some padding around the login link on the landing page.
* Threads the subscriber state through to the landing page so that the "Subscribe" link doesn't show if not applicable.


mbrandonw:
I just added another commit for adding link color to the account page for enterprise links.

Also realizing we don't have test coverage on that page, we should prob add some.

PR #434, opened 2019-04-24
Add Backtrace

stephencelis:
Not that we have many crashes, but this'll hopefully help us debug things during upgrades, etc.

### DIFF
--- a/Sources/Runner/main.swift
+++ b/Sources/Runner/main.swift
@@ -1,8 +1,11 @@
+import Backtrace
 import Either
 import Foundation
 import PointFree
 
 // Bootstrap
+
+Backtrace.install()
 
 _ = try! PointFree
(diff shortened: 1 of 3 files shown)


Sources/Runner/main.swift
```diff
@@ -1,9 +1,12 @@
+import Backtrace
 import Either
 import Foundation
 import PointFree
 
 // Bootstrap
 
+Backtrace.install()
```
mbrandonw: I guess this could also be a step in the `bootstrap` process? think that's worth it?

stephencelis: I thought about that but liked the explicitness of it being called in `main.swift` as a dependency. Don't feel super strongly about it, though. Want me to move?

PR #435, opened 2019-05-08
Break all in reference links.

mbrandonw:
A long link in a reference will break the page and cause it to scroll horizontally. Adding this bit of css fixes it. Looks like we don't have support for `word-break` in our CSS lib, and we should def add it, but I wanted to quickly fix this...


Tests/PointFreeTests/EpisodePageTests.swift
```diff
@@ -435,5 +435,7 @@ class EpisodePageTests: TestCase {
       assertSnapshot(matching: webView, as: .image, named: "mobile")
     }
     #endif
+
+    assertSnapshot(matching: conn |> siteMiddleware, as: .ioConn)
```
mbrandonw: we didn't have html snapshot coverage on the episode page when references are present, so adding it.

stephencelis: Ah cool. No change to the screen shots?

mbrandonw: nah, but I could add a really long link to test it out. I'll do that

mbrandonw: unfortunately I wasn't able to reproduce the breaking problem in a test, but nonetheless I updated the mock to have a really long url.

PR #436, opened 2019-05-09
Blob post for enterprise subscriptions.

mbrandonw:
Blog post for tomorrow


Sources/Models/BlogPosts/BlogPostXYZW_TODO.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+public let postXYZW_TODO = BlogPost(
```
stephencelis: What’s this for?

PR #439, opened 2019-05-10
Add enterprise subscription to FAQ.

mbrandonw:


mbrandonw:
ah good catch, re-snapping now

PR #441, opened 2019-05-20
Link to Previous Episodes

stephencelis:
It'd be nice to build out a more robust support for episode collections/tags, but this should be a nice stop-gap in the meantime.


Sources/PointFree/Episode/Show.swift
```diff
@@ -487,10 +487,36 @@ private let episodeInfoView = View<(EpisodePermission, Episode)> { permission, e
...
+
+  return [
+    div(
+      [`class`([Class.padding([.mobile: [.top: 2], .desktop: [.top: 3]]), Class.pf.colors.bg.white])],
+      [
+        div([
+          "This episode builds on concepts introduced previously:"]),
+        ul(
+          previousEps.map {
+            li([
```
mbrandonw: We should try hiding the bullet from the `li`. I think that would be nicer since it already has the `#X` prefix.

 you can do it with the utility class `Class.type.list.styleNone`.

Sources/PointFree/Episode/Show.swift
```diff
@@ -487,10 +487,36 @@ private let episodeInfoView = View<(EpisodePermission, Episode)> { permission, e
...
+private func previousEpisodes(of ep: Episode) -> [Node] {
+  let previousEps = ep.previousEpisodes
+  guard !previousEps.isEmpty else { return [] }
+
+  return [
+    div(
+      [`class`([Class.padding([.mobile: [.top: 2], .desktop: [.top: 3]]), Class.pf.colors.bg.white])],
+      [
+        div([
+          "This episode builds on concepts introduced previously:"]),
```
mbrandonw: I think it would be nice if the text in this callout was the same style as the episode blurb, which I think can be done with `Class.pf.type.body.regular`.

mbrandonw:
Another thought I had, should we expose this info in the podcast feed too? feel like more and more people are watching videos that way...

stephencelis:
> Another thought I had, should we expose this info in the podcast feed too? feel like more and more people are watching videos that way...

Ah like show notes? I can try that in a followup PR.

stephencelis:
Gonna merge manually from the command line. Something strange with the submodule.

stephencelis:
Nevermind. Figured it out!

PR #442, opened 2019-06-05
Xcode 10.2 Support

stephencelis:
This gets all dependencies, etc., in order so we can finally `rm` Xcode 10.1.

### DIFF
--- a/PointFree.playground/Pages/Database.xcplaygroundpage/Contents.swift
+++ b/PointFree.playground/Pages/Database.xcplaygroundpage/Contents.swift
@@ -7,28 +7,28 @@ PlaygroundPage.current.needsIndefiniteExecution = true
 
 Current.envVars.postgres.databaseUrl
 
-//Current.database.migrate().run.perform()
-//
-//Current.database.fetchFreeEpisodeUsers()
-//  .run
-//  .perform()
-//  .right!
-//
-//let user = Current.database.registerUser(.mock, "blob@pointfree.co")
-//  .run
-//  .perform()
-//  .right!!
-//
-//user.episodeCreditCount
-//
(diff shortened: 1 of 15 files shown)


PointFree.playground/Pages/Database.xcplaygroundpage/Contents.swift
```diff
@@ -7,28 +7,28 @@ PlaygroundPage.current.needsIndefiniteExecution = true
...
-  .perform()
-  .right!
-
-let updatedUser = Current.database.fetchUserById(user.id)
-  .run
-  .perform()
-  .right!!
-
-updatedUser.episodeCreditCount
+//Current.database.migrate().run.perform()
```
mbrandonw: should we revert this?

Sources/PointFree/Account/Account.swift
```diff
@@ -40,7 +40,7 @@ private func fetchAccountData<I>(
 
   let (user, subscriberState) = lower(conn.data)
 
-  let userSubscription = user.subscriptionId
+  let userSubscription: EitherIO<Error, Models.Subscription> = user.subscriptionId
```
mbrandonw: type inference probs?

PointFree.playground/Pages/Database.xcplaygroundpage/Contents.swift
stephencelis: ah yeah. snuck in sorry!

Sources/PointFree/Account/Account.swift
stephencelis: yeah same as with Swift 5 :( thinks it's an optional

PR #446, opened 2019-06-11
Xcode 11 Support

stephencelis:
Requires a swift-web bump and a few small changes around type checking. Also ditches all our `xcodeproj` stuff. Nice!


.gitignore
```diff
@@ -38,6 +38,7 @@ playground.xcworkspace
 # Packages/
 # Package.pins
 .build/
+.swiftpm/
```
stephencelis: New directory that has Xcode project-related metadata.

Makefile
```diff
@@ -13,63 +13,17 @@ bootstrap:
 bootstrap-oss:
 	@echo "  ⚠️  Bootstrapping open-source Point-Free..."
 	@set -e; set -o pipefail; $(MAKE) .env | sed "s/make\[1\]: \`\.env'/\  ✅ $$(tput bold).env$$(tput sgr0)/"
-	@$(MAKE) xcodeproj-oss
```
stephencelis: No longer need to deal with an Xcode project.

Makefile
```diff
@@ -13,63 +13,17 @@ bootstrap:
...
 	@echo "  👀 Bootstrapping Point-Free (private)..."
-	@$(MAKE) xcodeproj
-	@$(MAKE) install-mm
+	@$(MAKE) check-dependencies
 	@echo "  ✅ Bootstrapped! Opening Xcode..."
 	@sleep 1 && xed .
 
-uninstall: uninstall-mm db-drop
-
-install-mm:
```
stephencelis: I think we lose playground support...but Xcode previews should work even better for our purposes (when we're 5.1-compatible).

Package.swift
```diff
@@ -90,6 +90,7 @@ let package = Package(
       dependencies: [
         "CssTestSupport",
         "FunctionalCss",
+        "Html",
```
stephencelis: SwiftPM dependency resolution has improved and we were missing this.

PR #448, opened 2019-07-03
Fix inline video.

mbrandonw:
@krzyzanowskim [brought](https://twitter.com/krzyzanowskim/status/1146355372342611969) it to my attention that videos don't play inline on mobile. We use the `playsinline` attribute on the `video` tag, but apparently video.js overrides that, so I add to also add it to the video.js config.


mbrandonw:
@stephencelis I think you are traveling this morning, so I'm going to go ahead and merge this. I think it's pretty straightforward.

stephencelis:
Thanks for takin care of this!

PR #449, opened 2019-07-16
Vimeo Support

stephencelis:
Still working on tests.

### DIFF
--- a/Sources/PointFree/Episode/Show.swift
+++ b/Sources/PointFree/Episode/Show.swift
@@ -186,27 +186,7 @@ private let episodeView = View<(EpisodePermission, User?, SubscriberState, Episo
           )
         ]
       )
-      ]),
-
-    script(
-      """
-      var hasPlayed = false;
-      var video = document.getElementsByTagName('video')[0];
-      video.addEventListener('play', function () {
-        hasPlayed = true;
-      });
-      document.addEventListener('keypress', function (event) {
-        if (hasPlayed && event.key === ' ') {
-          if (video.paused) {
-            video.play();
-          } else {
-            video.pause();
-          }
-          event.preventDefault();
-        }
-      });
-      """
-    )
+    ])
   ]
 }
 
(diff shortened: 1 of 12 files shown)


Sources/PointFree/Episode/Show.swift
```diff
@@ -186,27 +186,7 @@ private let episodeView = View<(EpisodePermission, User?, SubscriberState, Episo
           )
         ]
       )
-      ]),
-
-    script(
```
stephencelis: Moved this logic to the video player file.

Sources/PointFree/Episode/Show.swift
```diff
@@ -253,15 +233,15 @@ private let tocChapterLinkView = View<(title: String, timestamp: Int, active: Bo
       [
         div([`class`([Class.hide(.mobile)])], [
           a(
-            timestampLinkAttributes(timestamp: timestamp, useAnchors: true) +
```
stephencelis: Removed this argument and am always linking directly to the video timestamp now.

Sources/Views/EpisodeVideoView.swift
```diff
@@ -7,35 +7,105 @@ import Styleguide
...
-          data("setup", VideoJsOptions.default.jsonString)
-        ],
-        [
-          source(
-            src: isEpisodeViewable
-              ? episode.fullVideo.streamingSource
-              : episode.trailerVideo?.streamingSource ?? "",
-            [type(.application(.init(rawValue: "vnd.apple.mpegurl")))]
-          )
+    episodeSource.hasPrefix("https://player.vimeo.com")
```
stephencelis: We can incrementally adopt it.

Sources/Views/TranscriptBlockView.swift
```diff
@@ -120,25 +120,10 @@ public func timestampLabel(for timestamp: Int) -> String {
   return "\(minuteString):\(secondString)"
 }
 
-public func timestampLinkAttributes(timestamp: Int, useAnchors: Bool) -> [Attribute<Tag.A>] {
+public func timestampLinkAttributes(timestamp: Int) -> [Attribute<Tag.A>] {
 
   return [
-    useAnchors
```
stephencelis: All this logic moved/updated in the video view.

PR #454, opened 2019-07-30
Update corrections for beta 5

stephencelis:


Sources/Models/PublicEpisodes/0065-SwiftUIAndStateManagementPt1.swift
```diff
@@ -66,6 +66,13 @@ These problems can be fixed by creating a generic class `Store<A>` that wraps ac
 ]
 
 private let transcriptBlocks: [Episode.TranscriptBlock] = [
+  Episode.TranscriptBlock(
+    content: """
+This episode was recorded with Xcode 11 beta 3. Xcode 11 beta 5 introduced a lot of changes. While we note these changes inline below, we also went over them in detail [on our blog](/blog/posts/30-swiftui-and-state-management-corrections).
```
mbrandonw: Might read better to avoid the duplication of "Xcode 11 beta":

```suggestion
This episode was recorded with Xcode 11 beta 3, and a lot has changed in recent betas. While we note these changes inline below, we also went over them in detail [on our blog](/blog/posts/30-swiftui-and-state-management-corrections).
```

Sources/Models/PublicEpisodes/0065-SwiftUIAndStateManagementPt1.swift
```diff
@@ -727,7 +734,13 @@ var didChange: AppState.PublisherType
   ),
   Episode.TranscriptBlock(
     content: """
-This episode was recorded with Xcode 11 beta 3, and a change has been made to the `BindableObject` protocol in beta 4 and later versions of Xcode. The protocol now requires a `willChange` publisher, and you are supposed to ping this publisher _before_ you make any mutations to your model.
+This episode was recorded with Xcode 11 beta 3. In Xcode 11 beta 5 and later versions, SwiftUI's `BindableObject` protocol was deprecated in favor of an `ObservableObject` protocol that was introduced to the Combine framework. This protocol utilizes an `objectWillChange` property of `ObservableObjectPublisher`, which is pinged _before_ (not after) any mutations are made to your model:
```
mbrandonw: Avoiding the "Xcode 11 beta" duplication like you did below:

```suggestion
In Xcode 11 beta 5 and later versions, SwiftUI's `BindableObject` protocol was deprecated in favor of an `ObservableObject` protocol that was introduced to the Combine framework. This protocol utilizes an `objectWillChange` property of `ObservableObjectPublisher`, which is pinged _before_ (not after) any mutations are made to your model:
```

Sources/Models/PublicEpisodes/0065-SwiftUIAndStateManagementPt1.swift
```diff
@@ -727,7 +734,13 @@ var didChange: AppState.PublisherType
...
+
+```
+let objectDidChange = ObservableObjectPublisher()
+```
+
+This boilerplate is also not necessary, as the `BindableObject` protocol will synthesize a default publisher for you automatically.
```
mbrandonw:
```suggestion
This boilerplate is also not necessary, as the `ObservableObject` protocol will synthesize a default publisher for you automatically.
```

Sources/Models/PublicEpisodes/0066-SwiftUIAndStateManagementPt2.swift
```diff
@@ -387,7 +394,19 @@ var favoritePrimes: [Int] = [] {
   ),
   Episode.TranscriptBlock(
     content: """
-This episode was recorded with Xcode 11 beta 3, and a change has been made to the `BindableObject` protocol in beta 4 and later versions of Xcode. The protocol now requires a `willChange` publisher, and you are supposed to ping this publisher _before_ you make any mutations to your model.
+This episode was recorded with Xcode 11 beta 3. In Xcode 11 beta 5 and later versions, SwiftUI's `BindableObject` protocol was deprecated in favor of an `ObservableObject` protocol that was introduced to the Combine framework. This protocol utilizes an `objectWillChange` property of `ObservableObjectPublisher`, which is pinged _before_ (not after) any mutations are made to your model. Because of this, `willSet` should be used instead of `didSet`:
```
mbrandonw: same comments as above...

Sources/Models/PublicEpisodes/0067-SwiftUIAndStateManagementPt3.swift
```diff
@@ -795,6 +835,23 @@ Unfortunately, it came with a _lot_ of boilerplate, but let's make sure it actua
...
+
+```
+var state: Binding<(favoritePrimes: [Int], activityFeed: [AppState.Activity])>
+""",
+    timestamp: nil,
+    type: .correction
+  ),
+  Episode.TranscriptBlock(
+    content: """
+Let's make sure it works. First, we can replace the object binding in our favorite primes view.
```
mbrandonw: is this new transcript block supposed to be here?

stephencelis: It was combined above, but I wanted to break it in half with the correction so that it more directly applied to the previous text.

mbrandonw: ah I see, should that part of the text be removed from the other transcript block then?

stephencelis: Dang, I swear I did that. Xcode's been crashing a bunch for me.

PR #455, opened 2019-08-02
Remove some padding from related episodes.

mbrandonw:
I noticed the left padding on related episodes was a bit much and caused some word wrap that wasn't necessary. so I reduced that, and there's a few other miscellaneous things in here....


Sources/Models/PublicEpisodes/References.swift
```diff
@@ -61,6 +61,16 @@ This question on the Swift forums brings up an interesting discussion on how to
     title: "Difficulties With Efficient Large File Parsing"
   )
 
+  public static let elmHomepage = Episode.Reference(
```
mbrandonw: references for next week

stephencelis: 😯

PR #456, opened 2019-08-06
Fixes for JS error and word breaks.

mbrandonw:
Fixes a few problems we have on the episode page right now:

* JS error because of some old JS that dealt with video.js
* Long code lines in corrections and exercises cause the page to overflow horizontally


Sources/Views/Markdown.swift
```diff
@@ -33,6 +33,7 @@ public let markdownBlockStyles: Stylesheet =
       <> blockquoteMarkdownStyles
       <> pMarkdownStyles
       <> codeMarkdownStyles
+      <> key("word-break", "break-word")
```
mbrandonw: decided to just break words on all markdown containers. fixes the problem in a bunch of places at once

Sources/PointFree/Episode/Show.swift
```diff
@@ -34,7 +34,6 @@ let episodeResponse =
           currentUser: currentUser,
           data: (permission, currentUser, subscriberState, episode),
           description: episode.blurb,
-          extraHead: videoJsHead,
```
stephencelis: This'll temporarily making video.js videos less featured, right?

Sources/Views/Markdown.swift
stephencelis: Nice, thanks for doing this!

Sources/PointFree/Episode/Show.swift
mbrandonw: oh yes, I forgot we still have some video.js. I'll bring it back for now

PR #457, opened 2019-08-07
Stubs for new pricing page

mbrandonw:
Creates the route, middleware and view for the new pricing page.

### DIFF
--- a/Sources/PointFree/Account/Cancel.swift
+++ b/Sources/PointFree/Account/Cancel.swift
@@ -10,6 +10,7 @@ import Mailgun
 import Models
 import Optics
 import PointFreePrelude
+import PointFreeRouter
 import Prelude
 import Styleguide
 import Stripe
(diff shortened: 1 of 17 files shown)


Sources/PointFree/Account/Cancel.swift
```diff
@@ -10,6 +10,7 @@ import Mailgun
 import Models
 import Optics
 import PointFreePrelude
+import PointFreeRouter
```
mbrandonw: I moved the `url(to:)` and `path(to:)` helpers to `PointFreeRouter` (where it belongs), so some extra imports here. I didn't end up needing to do that, but seemed worth keeping.

Sources/PointFree/SubscribeLanding.swift
```diff
@@ -0,0 +1,31 @@
...
+import View
+import Views
+
+public let subscribeLanding: Middleware<
+  StatusLineOpen,
+  ResponseEnded,
+  Tuple3<User?, SubscriberState, Route>,
+  Data
+  > =
+  requireAdmin
```
mbrandonw: only us admins will be able to access this page.

PR #458, opened 2019-08-08
Header for new pricing page

mbrandonw:
This gets the basics of the new pricing hero in place. I've decided to just completely hide the CTA column when you are logged out / not an active subscriber. Not sure what else we could do.


Sources/FunctionalCss/Spacing.swift
```diff
@@ -105,7 +105,8 @@ private let spacings: [Size] = [
   .rem(0.5),
   .rem(1.0),
   .rem(2.0),
-  .rem(4.0)
+  .rem(4.0),
+  .rem(8.0)
```
mbrandonw: We need a new size of spacing for the hero area of the pricing page, since on desktop it has a healthy amount of padding (~128px)

Sources/PointFree/MinimalNavView.swift
```diff
@@ -11,6 +11,7 @@ import PointFreeRouter
 import Styleguide
 import Prelude
 import View
+import Views
```
mbrandonw: this import is now needed because I moved our SVG's to the views framework. I'll comment more about this in a moment.

Sources/PointFree/MinimalNavView.swift
```diff
@@ -101,6 +102,8 @@ private let logInLinkView = View<(NavStyle.MinimalStyle, Route?)> { style, curre
 
 private func gitHubLinkType(for style: NavStyle.MinimalStyle) -> GitHubLinkType {
   switch style {
+  case .black:
+    return .white
```
mbrandonw: I added a new style of minimal nav for a completely black nav. this will need some separate work because @amotion's design calls for a completely different nav, but I didn't want to tackle that yet.

Sources/PointFree/SVG.swift
```diff
@@ -1,227 +0,0 @@
-import Foundation
```
mbrandonw: I'm deleting this file because somehow there was also a copy of this file in the Views framework. maybe when we were modularizing we forgot to delete this...

stephencelis:
Wanna get an image snapshot test in place?

Sources/Views/Svg.swift
```diff
@@ -5,7 +5,7 @@ private func base64EncodedString(_ string: String) -> String {
   return Data(string.utf8).base64EncodedString()
 }
 
-func newWindowSvgBase64(fill: String) -> String {
+public func newWindowSvgBase64(fill: String) -> String {
```
mbrandonw: making all of these public so that we can access them outside the framework.

Sources/Views/Svg.swift
```diff
@@ -111,13 +111,7 @@ let pointFreeHeroSvgBase64 = base64EncodedString <| """
 </svg>
 """
 
-func gitHubSvgBase64(fill: String) -> String {
```
mbrandonw: this svg is already in the styleguide framework.

mbrandonw:
> Wanna get an image snapshot test in place?

Added some tests here https://github.com/pointfreeco/pointfreeco/pull/458/commits/8a74ae8dd2c03a374d37efcd392fca23c141ec80

PR #459, opened 2019-08-08
Add backtrace library

mbrandonw:
This adds the backtrace library to our bootstrap process. I had to make a few other unrelated changes too...


.gitignore
```diff
@@ -70,7 +70,7 @@ fastlane/test_output
 Packages/*
 *.xcodeproj/*
 Package.resolved
-.env
+.pf-env
```
mbrandonw: apparently newer versions of Docker use the `.env` file in the root for environment stuff so we can no longer use it as the place for our environment stuff. I looked for ways to get docker to ignore it, like the `.dockerignore` file but nothing worked.

Dockerfile
```diff
@@ -19,6 +19,6 @@ RUN make -C cmark INSTALL_PREFIX=/usr
 RUN make -C cmark install
 
 RUN swift package update
-RUN swift build --product Server --configuration release && \
-    swift build --product Runner --configuration release
+RUN swift build --product Server --configuration release -Xswiftc -g && \
+    swift build --product Runner --configuration release -Xswiftc -g
```
mbrandonw: this flag builds with debug symbols.

PR #462, opened 2019-08-08
Footer for new pricing page.

mbrandonw:
Adds the footer module to the new pricing page. I can't snapshot this because it only appears if you are logged out and we require admins right now, but I will get coverage on this once we launch. Here's what it looks like:

image


Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -348,6 +350,7 @@ private let colorStyles: Stylesheet =
   Class.pf.colors.bg.black % backgroundColor(Colors.black)
     <> Class.pf.colors.bg.blue900 % backgroundColor(Colors.blue900)
     <> Class.pf.colors.bg.dark % backgroundColor(Colors.black)
+    <> Class.pf.colors.bg.gray150 % backgroundColor(Colors.gray150)
```
mbrandonw: new atomic style for parking backgrounds, will cause lots of snapshots to update

PR #463, opened 2019-08-08
Don't play video on non-timestamped links

stephencelis:
Accidentally introduced this with Vimeo.


Sources/Views/EpisodeVideoView.swift
```diff
@@ -35,20 +35,20 @@ public func videoView(forEpisode episode: Episode, isEpisodeViewable: Bool) -> N
...
     var target = event.target;
     if (target.tagName != "A") { return; }
     var hash = new URL(target.href).hash;
-    jump(hash);
-    player.play();
+    jump(hash, true);
   });
 
-  function jump(hash) {
+  function jump(hash, play) {
```
stephencelis: This `play` parameter prevents a JavaScript error. Playing videos can only work from JS fired from a direct user action, and window load isn't considered one.

PR #464, opened 2019-08-08
"What to expect" section for new pricing page

mbrandonw:
Adds the "what to expect" module to the new pricing page. The images don't show up in the screenshot tests (because they are on s3), so here is what the whole page looks like on desktop:

localhost_8080_subscribe


mbrandonw:
> Ha yer on a roll. Gonna save any of this for me?

ha well there is a whole other page that is a lot more complicated than this one.

PR #465, opened 2019-08-10
Plans on new pricing plan

mbrandonw:
Adds the pricing plans module. The copy and prices still need to be tweaked, just getting the main pieces in place.

There's a lot of churn in this PR because I made an update to some font sizes that affect lots of pages.

localhost_8080_subscribe (1)


Sources/FunctionalCss/TypeScale.swift
```diff
@@ -65,7 +65,7 @@ public let typescale =
     <> Class.h3 % fontSize(.rem(2))    // 32  28
     <> Class.h4 % fontSize(.rem(1.5))  // 24  21
     <> Class.h5 % fontSize(.rem(1.0))  // 16  14
-    <> Class.h6 % fontSize(.rem(0.75)) // 12  11
+    <> Class.h6 % fontSize(.rem(0.875)) // 14  12.25
```
mbrandonw: There was way too big of a gap between our 2nd smallest and smallest font, so I adjusted. I audited the few places we use this small font, and each place this bump made things a lot more readable. This causes a lot of changes in our snapshots.

PR #470, opened 2019-08-16
Adds a max width container to the pricing landing

mbrandonw:
Been meaning to do this. Pretty straightforward, but I couldn't do one big container for the whole page since each module has a background that bleeds out of the margins. so I had to apply max width containers to each module.

 Here's what it looks like when the browser is made bigger than the max width:

localhost_8080__pricing

### DIFF
--- a/Sources/Views/PricingLanding.swift
+++ b/Sources/Views/PricingLanding.swift
@@ -108,17 +108,19 @@ private func hero(currentUser: User?, subscriberState: SubscriberState) -> [Node
       [
         `class`([
           Class.pf.colors.bg.black,
-          Class.padding([.mobile: [.leftRight: 3, .topBottom: 4], .desktop: [.all: 5]])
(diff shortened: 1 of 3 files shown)


Sources/Views/PricingLanding.swift
```diff
@@ -352,70 +355,75 @@ private func pricingPlan(_ plan: PricingPlan) -> ChildOf<Tag.Ul> {
...
 
 private let whatToExpect = [
-  gridRow(
-    [
-      `class`([
-        Class.padding([.mobile: [.leftRight: 2, .topBottom: 3], .desktop: [.all: 4]])
-        ]),
-      style(backgroundColor(.other("#fafafa")))
-    ],
+  div(
```
mbrandonw: this file has a lot of churn due to some indentation, so may be best to review with "ignore whitespace" on

PR #472, opened 2019-08-18
Add+invite teammates at once

mbrandonw:
This is PR is based off of the `sub-confirmation-stubs` branch.

It adds the ability to simultaneously increase the team subscription by 1 and invite a teammate from the account page.


Sources/PointFree/Account/Account.swift
```diff
@@ -772,7 +773,7 @@ private func mainAction(for subscription: Stripe.Subscription) -> Node {
           action(path(to: .account(.subscription(.change(.update(nil)))))),
           method(.post),
           onsubmit(unsafe: """
-if (!confirm("Upgrade to yearly billing? You will be charged \(formattedAmount ?? "") immediately with a pro-rated refund for the time remaining in your billing period.")) {
+if (!confirm("Upgrade to yearly billing? You will be charged \(formattedAmount ?? "") immediately with a prorated refund for the time remaining in your billing period.")) {
```
mbrandonw: I think it's more correct to use "prorated" over "pro-rated". is that true?

Sources/PointFree/Account/Account.swift
```diff
@@ -906,6 +907,64 @@ private func inviteRowView(_ invite: TeamInvite) -> Node {
     ])
 }
 
+private func addTeammateToSubscriptionRow(_ data: AccountData) -> [Node] {
```
mbrandonw: this is the module to allow adding a teammate:

image

it only shows if you have no invites remaining on your subscription.

Sources/PointFree/Account/Account.swift
```diff
@@ -906,6 +907,64 @@ private func inviteRowView(_ invite: TeamInvite) -> Node {
...
+  guard !data.subscriberState.isEnterpriseSubscriber else { return [] }
+  guard let subscription = data.stripeSubscription else { return [] }
+  let invitesRemaining = subscription.quantity - data.teamInvites.count - data.teammates.count
+  guard invitesRemaining == 0 else { return [] }
+  guard let amount = subscription.plan.tiers.min(by: { $0.amount < $1.amount })?.amount else { return [] }
+
+  let interval = subscription.plan.interval == .some(.year) ? "year" : "month"
+  let amountPerPeriod = "$\(amount.rawValue / 100) per \(interval)"
```
mbrandonw: can you double check how I am computing the amount and interval in the above 3 lines of code?

Sources/PointFree/Account/Account.swift
```diff
@@ -952,6 +1011,11 @@ private func subscriptionInviteMoreRowView(_ data: AccountData) -> [Node] {
   ]
 }
 
+private func inviteTeammatesDescription(invitesRemaining: Int) -> Node {
+  let seats = invitesRemaining == 1 ? "1 open seat": "\(invitesRemaining) open seats"
+  return .text("You have \(seats) on your team. Invite a team member below:")
```
mbrandonw: just fixing a pluralization error

Sources/PointFree/Account/Change.swift
```diff
@@ -12,82 +12,72 @@ import PointFreeRouter
...
-      )
-    )
+    <<< filterMap(require2 >>> pure, or: invalidSubscriptionErrorMiddleware)
     <<< fetchSeatsTaken
     <<< requireStripeSubscription
     <<< requireActiveSubscription
     <<< requireValidSeating
-    <| map(lower)
-    >>> subscriptionChange
+    <| changeSubscription(
```
mbrandonw: I generalized `subscriptionChange` so that it can take custom middleware to run in the success and failure cases of modifying the subscription. this allowed me to reuse all of this logic easily for my new route

Sources/PointFree/Account/Change.swift
```diff
@@ -108,8 +98,24 @@ func requireActiveSubscription<A>(
...
+func subscriptionModificationErrorMiddleware<A>(
+  _ conn: Conn<StatusLineOpen, A>
+  ) -> IO<Conn<ResponseEnded, Data>> {
+
+  return conn |> redirect(
+    to: .account(.index),
+    headersMiddleware: flash(
+      .error,
+      """
+      We couldn’t modify your subscription at this time. Please try again or contact
```
mbrandonw: this flash message is getting used in a few spots now so I extracted the middleware out

Sources/PointFree/Account/Change.swift
```diff
@@ -108,8 +98,24 @@ func requireActiveSubscription<A>(
...
+      <support@pointfree.co>.
+      """
+    )
+  )
+}
+
 private func requireValidSeating(
-  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, Tuple4<Stripe.Subscription, User, Int, Pricing>, Data>
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, (Stripe.Subscription, Pricing), Data>
```
mbrandonw: this middleware was dealing with more data than necessary so I focused it a bit more. this turned out to not really be necessary (I thought I was gonna need to share this at one point), but thought I might as well keep it in

Sources/PointFree/Account/Change.swift
```diff
@@ -160,15 +168,15 @@ private func fetchSeatsTaken<A>(
     }
 }
 
-private func defaultPricing(for lane: Pricing.Lane, billing: Pricing.Billing) -> Int {
-  switch (lane, billing) {
+private func defaultPricing(for pricing: Pricing) -> Cents<Int> {
```
mbrandonw: upgraded to cents to strength the code around the prorating logic: https://github.com/pointfreeco/pointfreeco/pull/472/files#diff-bab38e3e97cb96ef710c6056a81b5248R51

Sources/PointFree/Account/Invite.swift
```diff
@@ -122,6 +122,32 @@ let acceptInviteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple2<Tea
         .flatMap(const(conn |> redirect(to: path(to: .account(.index)))))
 }
 
+let addTeammateViaInviteMiddleware: Middleware<
```
mbrandonw: this is the middleware that updates the subscription to increase the quantity by 1 and then invites the teammate

Sources/PointFree/Account/Invite.swift
```diff
@@ -122,6 +122,32 @@ let acceptInviteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple2<Tea
...
+    <<< filterMap(require2 >>> pure, or: invalidSubscriptionErrorMiddleware)
+    <<< requireStripeSubscription
+    <<< requireActiveSubscription
+    <| { (conn: Conn<StatusLineOpen, Tuple3<Stripe.Subscription, User, EmailAddress>>) in
+
+      let (stripeSubscription, inviter, email) = lower(conn.data)
+      let newPricing = Pricing(
+        billing: stripeSubscription.plan.interval == .month ? .monthly : .yearly,
+        quantity: stripeSubscription.quantity + 1
+      )
```
mbrandonw: are the above lines of code a legit way to compute the new pricing? we should maybe combine `Plan.Interval` and `Pricing.Billing` into a single enum

Sources/PointFree/Account/Invite.swift
```diff
@@ -122,6 +122,32 @@ let acceptInviteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple2<Tea
...
+
+      return conn
+        .map(const((stripeSubscription, newPricing)))
+        |> changeSubscription(
+          error: subscriptionModificationErrorMiddleware,
+          success: map(const(email .*. inviter .*. unit)) >>> sendInviteMiddleware
```
mbrandonw: here we get to run the shared middleware `changeSubscription` and on success we then run the `sendInviteMiddleware` middleware.

Sources/PointFree/Account/Invite.swift
```diff
@@ -162,6 +188,20 @@ let showInviteView = View<(TeamInvite, User, User?)> { teamInvite, inviter, curr
...
+func invalidSubscriptionErrorMiddleware<A>(
+  _ conn: Conn<StatusLineOpen, A>
+  ) -> IO<Conn<ResponseEnded, Data>> {
+
+  return conn
+    |> redirect(
+      to: .account(.index),
+      headersMiddleware: flash(
+        .error,
+        "Invalid subscription data. Please try again or contact <support@pointfree.co>."
```
mbrandonw: this flash message was also used a few times so extracted it

Sources/PointFreeRouter/Routes.swift
```diff
@@ -63,6 +63,7 @@ public enum Route: DerivePartialIsos, Equatable {
 
   public enum Invite: DerivePartialIsos, Equatable {
     case accept(TeamInvite.Id)
+    case addTeammate(EmailAddress?)
```
mbrandonw: I added the route here, but maybe it makes sense to be in `Route.Account.Subscription.Change` instead? what do you think?

mbrandonw:
Tests are gonna fail on this PR, but I'm not gonna fix cause there's going to be more things to fix in the main branch. All of the `ChangeTests` tests still pass though, hopefully there's good coverage on all of that.

Sources/PointFree/Account/Account.swift
```diff
@@ -1066,7 +1129,6 @@ private struct AccountData {
   }
 
   var isTeamSubscription: Bool {
-    guard let id = self.stripeSubscription?.plan.id else { return false }
-    return id == .teamMonthly || id == .teamYearly
+    return (self.stripeSubscription?.quantity ?? 0) > 1
```
mbrandonw: is this the right logic for this?

mbrandonw: also I made a trello card for us to remove references to `.teamMonthly` and `.teamYearly`

Sources/PointFree/Account/Invite.swift
```diff
@@ -46,7 +46,18 @@ let revokeInviteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple2<Tea
...
         .run
-        .flatMap(const(conn |> redirect(to: path(to: .account(.index)))))
+        .flatMap(
+          const(
+            conn
+              |> redirect(
+                to: path(to: .account(.index)),
+                headersMiddleware: flash(
+                  .notice,
+                  "Invite to \(get1(conn.data).email.rawValue) has been revoked."
```
mbrandonw: unrelated to this PR but I realized I dont show a flash when revoking an invite, so adding that.

Sources/PointFree/Account/Account.swift
stephencelis: ah yup

Sources/PointFree/Account/Account.swift
stephencelis: don’t you have to account for if the current quantity is 1 it will be $14 for the additional seat, otherwise $16? so the difference between the max and min

Sources/PointFree/Account/Change.swift
stephencelis: interesting change away from Tuple. in general do you think we should move away from the structural types?

Sources/PointFree/Account/Change.swift
stephencelis: nice. should we add conversion methods between `Dollar`s and maybe even helpers for formatting to TaggedMoney? We got a lot of mileage out of TaggedTime having them along with `timeInterval`

Sources/PointFree/Account/Invite.swift
stephencelis: this should be legit for how we structure plans right now

Sources/PointFreeRouter/Routes.swift
stephencelis: i think here is good. `Change` should probably go away soon in favor of explicit `upgrade` and `downgrade` routes. would simplify a lot

Sources/PointFree/Account/Account.swift
stephencelis: this looks right to me

PR #475, opened 2019-08-20
Show more discount info on confirmation page

mbrandonw:


Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -548,23 +563,41 @@ window.addEventListener("load", function() {
             sizes: [.mobile: 12],
             [`class`([Class.padding([.mobile: [.top: 3, .bottom: 2]])])],
             [
-              p(
+              span(
```
mbrandonw: switched to span so that I could flow another span next to it for the discount info. this causes a bit of churn in the screenshots since spans have different default margins than `p`s do

PR #477, opened 2019-08-21
Discount redirect

stephencelis:
This will properly redirect card failures back to a discount page.


Sources/PointFree/SiteMiddleware.swift
```diff
@@ -83,8 +83,14 @@ private func render(conn: Conn<StatusLineOpen, T3<(Models.Subscription, Enterpri
       return conn.map(const(user .*. subscriberState .*. route .*. subRoute .*. unit))
         |> blogMiddleware
 
-    case let .discounts(couponId):
-      return conn.map(const(user .*. route .*. subscriberState .*. .personal .*. nil .*. couponId .*. unit))
+    case let .discounts(couponId, billing):
+      let subscribeData = SubscribeData(
```
stephencelis: Still need to swap this data type out for something simpler per another Trello.

PR #479, opened 2019-08-21
Simplify Subscribe Confirmation Data

stephencelis:
https://trello.com/c/r2kxgFLV/373-subscribeconfirmation-route-should-not-use-subscribedata

### DIFF
--- a/Sources/PointFree/SubscribeConfirmation.swift
+++ b/Sources/PointFree/SubscribeConfirmation.swift
@@ -12,7 +12,7 @@ import Views
 public let subscribeConfirmation: Middleware<
   StatusLineOpen,
   ResponseEnded,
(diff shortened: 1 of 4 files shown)


Sources/PointFree/SubscribeConfirmation.swift
```diff
@@ -12,7 +12,7 @@ import Views
 public let subscribeConfirmation: Middleware<
   StatusLineOpen,
   ResponseEnded,
-  Tuple6<User?, Route, SubscriberState, Pricing.Lane, SubscribeData?, Stripe.Coupon?>,
+  Tuple6<User?, Route, SubscriberState, Pricing.Lane, SubscribeConfirmationData?, Stripe.Coupon?>,
```
mbrandonw: it seems like this can be non-optional now right?

PR #482, opened 2019-08-21
Remove expanded stripe form

stephencelis:
Never really ended up needing/using this hidden feature.


stephencelis:
I'll cherry pick and force push things after the other is merged.

stephencelis:
Diff should be readable now.

PR #484, opened 2019-08-23
Bare bones API endpoints.

mbrandonw:
Just wanted to get some basic ideas down for API endpoints. I'm 100% open to suggestions and ideas to take this further. I think there are a lot of interesting directions we could take this, but I was being a little overwhelmed by the possibilities so I just wanted to get something down that we could iterate off of.

There are lots of TODO's in the code for future things we can do.

### DIFF
--- a/Tests/PointFreeTests/__Snapshots__/ApiTests/testEpisode.1.Conn.txt
+++ b/Tests/PointFreeTests/__Snapshots__/ApiTests/testEpisode.1.Conn.txt
@@ -2,7 +2,7 @@ GET http://localhost:8080/api/episodes/1
 Cookie: pf_session={}
 
 200 OK
-Content-Length: 8253
+Content-Length: 8251
 Content-Type: application/json
 Referrer-Policy: strict-origin-when-cross-origin
 X-Content-Type-Options: nosniff
@@ -209,8 +209,8 @@ X-XSS-Protection: 1; mode=block
     }
   ],
   "video" : {
-    "bytesLength" : 5000000,
-    "downloadUrl" : "https:\/\/s3.amazonaws.com\/pointfreeco\/trailer.mp4",
-    "streamingSource" : "https:\/\/s3.amazonaws.com\/pointfreeco\/trailer.m3u8"
+    "bytesLength" : 500000000,
+    "downloadUrl" : "https:\/\/s3.amazonaws.com\/pointfreeco\/video.mp4",
+    "streamingSource" : "https:\/\/s3.amazonaws.com\/pointfreeco\/video.m3u8"
   }
 }


Sources/PointFree/ApiMiddleware.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    init(episode: Episode, currentDate: Date) {
+      let subscriberOnly = episode.isSubscriberOnly(currentDate: currentDate)
+
+      self.blurb = episode.blurb
+      self.codeSampleDirectory = episode.codeSampleDirectory
+      self.id = episode.id
+      self.image = episode.image
+      self.length = episode.length
+      self.previousEpisodesInCollection = [] // TODO
```
mbrandonw: just leaving this as a later exercise

Sources/PointFree/ApiMiddleware.swift
```diff
@@ -0,0 +1,120 @@
...
+      .map {
+        Api.EpisodeDetail(
+          episode: $0,
+          currentDate: Current.date()
+        )
+    }
+
+    return conn.map(const(episode))
+      |> (
+        filterMap(pure, or: routeNotFoundMiddleware) // TODO: make a JSON 404 payload?
```
mbrandonw: right now im showing the html 404 page when we can't fetch the episode, but I guess we should have a better form of json erroring

Sources/PointFree/ApiMiddleware.swift
```diff
@@ -0,0 +1,120 @@
...
+
+public func respondJson<A: Encodable>(
+  _ conn: Conn<HeadersOpen, A>
+  ) -> IO<Conn<ResponseEnded, Data>> {
+
+  let encoder = JSONEncoder()
+  if #available(OSX 10.13, *), Current.envVars.appEnv == .testing {
+    encoder.outputFormatting = [.sortedKeys, .prettyPrinted]
+  }
+  let data = try! encoder.encode(conn.data) // TODO: 400 on badly formed data
```
mbrandonw: so I don't know how it's possible for this encoder to crash, but I also don't know how to properly handle this in middleware. by the time we've gotten to this point the status has already been written. perhaps it needs to work when status is open and it takes an extra headers middleware to run ¯\_(ツ)_/¯

Sources/PointFree/RouteNotFoundMiddleware.swift
```diff
@@ -11,21 +11,25 @@ import Styleguide
...
-      layoutData: { subscription, currentUser in
-        SimplePageLayoutData(
-          currentUser: currentUser,
-          data: unit,
-          title: "Page not found"
-        )
-    }
-)
+func routeNotFoundMiddleware<A>(
+  _ conn: Conn<StatusLineOpen, A>
```
mbrandonw: I need to generalize this middleware over `A` so that I could plug it into places that don't only work with `Unit`.

Sources/PointFreeRouter/AccountRoutes.swift
```diff
@@ -35,7 +35,7 @@ public enum Account: DerivePartialIsos, Equatable {
   }
 }
 
-public let accountRouter
```
mbrandonw: doesn't need to be public, just fixing

Sources/PointFree/ApiMiddleware.swift
stephencelis: Yeah I dunno if anything would throw normally here.

PR #493, opened 2019-09-02
Ghosting 👻

mbrandonw:
This gives admin the ability to ghost users (i.e. log in as them) so that we can debug things. When we do this a new cookie is saved that shows we are ghosting, and a large banner appears at the top of the screen so that we are aware of what we are doing.

I did some small refactors along the way, so will comment about that.

Screenshot 2019-09-02 11 15 34

### DIFF
--- a/Sources/PointFree/Account/AccountMiddleware.swift
+++ b/Sources/PointFree/Account/AccountMiddleware.swift
@@ -8,7 +8,7 @@ import Prelude
 import Stripe
 import Tuple
 
-func renderAccount(conn: Conn<StatusLineOpen, Tuple4<Models.Subscription?, User?, SubscriberState, Account>>)
+func accountMiddleware(conn: Conn<StatusLineOpen, Tuple4<Models.Subscription?, User?, SubscriberState, Account>>)
   -> IO<Conn<ResponseEnded, Data>> {
 
     let (_, user, subscriberState, account) = lower(conn.data)
--- a/Sources/PointFree/Account/UpdateProfile.swift
+++ b/Sources/PointFree/Account/UpdateProfile.swift
@@ -63,6 +63,7 @@ let updateProfileMiddleware: Middleware<
           )
           .run({ _ in })
       } else {
+        // TODO: why is unicode ‘ not encoded correctly?
         updateFlash = flash(.notice, "We've updated your profile!")
       }
 
--- a/Sources/PointFree/Account/UpdateProfile.swift
+++ b/Sources/PointFree/Account/UpdateProfile.swift
@@ -29,7 +29,12 @@ private func fetchStripeSubscription<A>(
     }
 }
 
-let updateProfileMiddleware =
+let updateProfileMiddleware: Middleware<
+  StatusLineOpen,
+  ResponseEnded,
+  Tuple2<User?, ProfileData?>,
+  Data
+  > =
   filterMap(require1 >>> pure, or: loginAndRedirect)
     <<< filterMap(require2 >>> pure, or: redirect(to: .account(.index)))
     <<< filter(
--- a/Sources/PointFree/Account/UpdateProfile.swift
+++ b/Sources/PointFree/Account/UpdateProfile.swift
@@ -52,7 +52,7 @@ let updateProfileMiddleware: Middleware<
 
       let updateFlash: Middleware<HeadersOpen, HeadersOpen, Prelude.Unit, Prelude.Unit>
       if data.email.rawValue.lowercased() != user.email.rawValue.lowercased() {
-        updateFlash = flash(.warning, "We’ve sent an email to \(user.email) to confirm this change.")
+        updateFlash = flash(.warning, "We've sent an email to \(user.email) to confirm this change.")
         parallel(
           sendEmail(
             to: [user.email],
@@ -63,7 +63,7 @@ let updateProfileMiddleware: Middleware<
           )
           .run({ _ in })
       } else {
-        updateFlash = flash(.notice, "We’ve updated your profile!")
+        updateFlash = flash(.notice, "We've updated your profile!")
       }
 
       let updateCustomerExtraInvoiceInfo = zip(
--- a/Sources/PointFree/Admin/AdminMiddleware.swift
+++ b/Sources/PointFree/Admin/AdminMiddleware.swift
@@ -3,9 +3,7 @@ import Foundation
 import HttpPipeline
 import Models
 import PointFreeRouter
-import PointFreePrelude
 import Prelude
-import Stripe
 import Tuple
 
 let adminMiddleware =
--- a/Sources/PointFree/Admin/AdminMiddleware.swift
+++ b/Sources/PointFree/Admin/AdminMiddleware.swift
@@ -1,0 +1,71 @@
+import ApplicativeRouter
+import Foundation
+import HttpPipeline
+import Models
+import PointFreeRouter
+import PointFreePrelude
+import Prelude
+import Stripe
+import Tuple
+
+let adminMiddleware =
+  requireAdmin
+    <| _adminMiddleware
+
+private func _adminMiddleware(
+  conn: Conn<StatusLineOpen, Tuple2<User, Admin>>
+  )
+  -> IO<Conn<ResponseEnded, Data>> {
+
+    let (user, route) = lower(conn.data)
+
+    switch route {
+    case let .episodeCredits(.add(userId: userId, episodeSequence: episodeSequence)):
+      return conn.map(const(user .*. userId .*. episodeSequence .*. unit))
+        |> redeemEpisodeCreditMiddleware
+
+    case .episodeCredits(.show):
+      return conn.map(const(user .*. unit))
+        |> showEpisodeCreditsMiddleware
+
+    case .index:
+      return conn.map(const(user .*. unit))
+        |> adminIndex
+
+    case .freeEpisodeEmail(.index):
+      return conn.map(const(user .*. unit))
+        |> indexFreeEpisodeEmailMiddleware
+
+    case let .freeEpisodeEmail(.send(episodeId)):
+      return conn.map(const(user .*. episodeId .*. unit))
+        |> sendFreeEpisodeEmailMiddleware
+
+    case .ghost(.end):
+      fatalError()
+
+    case .ghost(.index):
+      return conn.map(const(unit))
+        |> ghostIndexMiddleware
+
+    case let .ghost(.start(userId)):
+      return conn.map(const(user .*. userId .*. unit))
+        |> ghostStartMiddleware
+
+    case let .newBlogPostEmail(.send(blogPostId, formData, isTest)):
+      return conn.map(const(user .*. blogPostId .*. formData .*. isTest .*. unit))
+        |> sendNewBlogPostEmailMiddleware
+
+    case .newBlogPostEmail(.index):
+      return conn.map(const(user .*. unit))
+        |> showNewBlogPostEmailMiddleware
+
+    case let .newEpisodeEmail(.send(episodeId, subscriberAnnouncement, nonSubscriberAnnouncement, isTest)):
+      return conn.map(const(user .*. episodeId .*. subscriberAnnouncement .*. nonSubscriberAnnouncement .*. isTest .*. unit))
+        |> sendNewEpisodeEmailMiddleware
+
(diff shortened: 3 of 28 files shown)


Sources/PointFree/Account/AccountMiddleware.swift
```diff
@@ -8,7 +8,7 @@ import Prelude
 import Stripe
 import Tuple
 
-func renderAccount(conn: Conn<StatusLineOpen, Tuple4<Models.Subscription?, User?, SubscriberState, Account>>)
+func accountMiddleware(conn: Conn<StatusLineOpen, Tuple4<Models.Subscription?, User?, SubscriberState, Account>>)
```
mbrandonw: seemed like a more appropriate name

Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -47,7 +52,7 @@ let updateProfileMiddleware =
 
       let updateFlash: Middleware<HeadersOpen, HeadersOpen, Prelude.Unit, Prelude.Unit>
       if data.email.rawValue.lowercased() != user.email.rawValue.lowercased() {
-        updateFlash = flash(.warning, "We’ve sent an email to \(user.email) to confirm this change.")
+        updateFlash = flash(.warning, "We've sent an email to \(user.email) to confirm this change.")
```
mbrandonw: something about the way we encode json for cookies breaks when we us this unicode character. we should fix it asap, but I didn't want it to hold us up. I could prob write a test for this that uses a WKWebView 🤔

Sources/PointFree/Admin/AdminMiddleware.swift
```diff
@@ -0,0 +1,66 @@
...
+import Foundation
+import HttpPipeline
+import Models
+import PointFreeRouter
+import Prelude
+import Tuple
+
+let adminMiddleware =
+  requireAdmin
+    <| _adminMiddleware
```
mbrandonw: I pulled out all admin related routes/middleware into a new file. this also allows me to `requireAdmin` a single time, which allowed me to remove a bunch of `requireAdmin`'s in other middlewares

Sources/PointFree/Admin/AdminMiddleware.swift
```diff
@@ -0,0 +1,66 @@
...
+
+private func _adminMiddleware(
```
mbrandonw: this is mostly just cut-pasted from `siteMiddleware`

Sources/PointFree/Admin/EpisodeCredit.swift
```diff
@@ -15,14 +15,22 @@ import Tuple
 import View
 import Views
 
-let showEpisodeCreditsMiddleware: Middleware<StatusLineOpen, ResponseEnded, Tuple1<User?>, Data> =
-  requireAdmin
-    <| writeStatus(.ok)
+let showEpisodeCreditsMiddleware: Middleware<
+  StatusLineOpen,
+  ResponseEnded,
+  Tuple1<User>,
```
mbrandonw: was able to make all the admin middlewares take an honest user instead of an optional, and also changed the formatting of these middleware decelerations to be a little more readable.

Sources/PointFree/Session.swift
```diff
@@ -18,7 +18,7 @@ public func writeSessionCookieMiddleware<A>(_ update: @escaping (Session) -> Ses
       let value = update(conn.request.session)
       guard value != conn.request.session else { return pure(conn) }
       return setCookie(
-          key: pointFreeUserSession,
+          key: pointFreeUserSessionCookieName,
```
mbrandonw: just a rename to be clearer

Sources/PointFree/Session.swift
```diff
@@ -50,11 +50,127 @@ extension URLRequest {
   }
 }
 
-public struct Session: Codable, Equatable {
+public struct Session: Equatable {
```
mbrandonw: the `Session` type changed quite a bit. I refactored to not let `userId` be the defining quality of the session, but rather a new enum `User = standard | ghosting`. I needed to do a decent amount of work to support backwards compatibility with the old session (luckily snapshot tests of middleware caught all regressions). I would be interested if you see any spots to simplify further.

Sources/PointFree/Session.swift
```diff
@@ -50,11 +50,122 @@ extension URLRequest {
...
+    case user
+    case userId
+  }
+
+  public func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    try container.encodeIfPresent(self.flash, forKey: .flash)
+    switch self.user {
+    case let .some(.standard(userId)):
+      try container.encode(userId, forKey: .userId)
```
mbrandonw: I need to encode `User.standard` in the session as just a single value `userId` in order to be backwards compatible to how sessions used to be encoded.

Sources/PointFree/Session.swift
```diff
@@ -50,11 +50,122 @@ extension URLRequest {
...
+
+    self.flash = try container.decodeIfPresent(Flash.self, forKey: .flash)
+    do {
+      self.user = .standard(try container.decode(Models.User.Id.self, forKey: .userId))
+    } catch {
+      do {
+        self.user = try container.decode(Session.User.self, forKey: .user)
+      } catch {
+        self.flash = nil
+        self.user = nil
```
mbrandonw: we need this last catch all to make sure we get a decoded session, which can just be `nil` for both fields.

Sources/PointFreeRouter/Routes.swift
```diff
@@ -19,6 +19,7 @@ public enum Route: DerivePartialIsos, Equatable {
   case appleDeveloperMerchantIdDomainAssociation
   case blog(Blog)
   case discounts(code: Stripe.Coupon.Id, Pricing.Billing?)
+  case endGhosting
```
mbrandonw: this route is not in the admin routes because you will not be logged in as an admin when you need to end ghosting

Tests/PointFreeTests/GhostTests.swift
```diff
@@ -0,0 +1,191 @@
...
+import SnapshotTesting
+import XCTest
+
+final class GhostTests: TestCase {
+  override func setUp() {
+    super.setUp()
+    record=true
+  }
+
+  func testStartGhosting_HappyPath() {
```
mbrandonw: these tests have some pretty complicated mocking needs in order to get the environment into shape for properly testing ghosting. would love to find ways of simplifying.

Tests/PointFreeTests/GhostTests.swift
```diff
@@ -0,0 +1,191 @@
...
+    _assertInlineSnapshot(matching: conn, as: .conn, with: """
+POST http://localhost:8080/admin/ghost/start
+Cookie: pf_session={"userId":"12121212-1212-1212-1212-121212121212"}
+
+user_id=10101010-DEAD-BEEF-DEAD-BEEFDEADBEEF
+
+302 Found
+Location: /
+Referrer-Policy: strict-origin-when-cross-origin
+Set-Cookie: pf_session={"user":{"ghosteeId":"10101010-DEAD-BEEF-DEAD-BEEFDEADBEEF","ghosterId":"12121212-1212-1212-1212-121212121212"}}; Expires=Sat, 29 Jan 2028 00:00:00 GMT; Path=/
```
mbrandonw: this test shows our session transitioning into a ghosting session when we hit the `/admin/ghost/start` endpoint.

Tests/PointFreeTests/GhostTests.swift
```diff
@@ -0,0 +1,191 @@
...
+      |> Prelude.perform
+
+    _assertInlineSnapshot(matching: conn, as: .conn, with: """
+POST http://localhost:8080/ghosting/end
+Cookie: pf_session={"user":{"ghosteeId":"10101010-DEAD-BEEF-DEAD-BEEFDEADBEEF","ghosterId":"12121212-1212-1212-1212-121212121212"}}
+
+302 Found
+Location: /
+Referrer-Policy: strict-origin-when-cross-origin
+Set-Cookie: pf_session={"userId":"12121212-1212-1212-1212-121212121212"}; Expires=Sat, 29 Jan 2028 00:00:00 GMT; Path=/
```
mbrandonw: this test shows us transitioning back to our original account when ending ghosting.

Sources/PointFree/Session.swift
stephencelis: Yeah this was my main thought reading so far.

stephencelis: Also, what the behavior of breaking? Does it just re-log you in via GitHub redirect? Is it worth letting the breakage happen if so?

Sources/PointFree/Session.swift
```diff
@@ -50,11 +50,122 @@ extension URLRequest {
...
+public struct Session: Equatable {
   public var flash: Flash?
-  public var userId: User.Id?
+  public var user: User?
 
-  public static let empty = Session(flash: nil, userId: nil)
+  public static let empty = Session(flash: nil, user: nil)
+
+  public var userId: Models.User.Id? {
+    get {
```
stephencelis: Minor: nesting you could kill...

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -92,6 +60,10 @@ private func render(conn: Conn<StatusLineOpen, T3<(Models.Subscription, Enterpri
       return conn.map(const(user .*. route .*. subscriberState .*. .personal .*. subscribeData .*. couponId .*. unit))
         |> discountSubscribeConfirmation
 
+    case .endGhosting:
```
stephencelis: Ghost buster.

Tests/PointFreeTests/SessionTests.swift
```diff
@@ -0,0 +1,82 @@
...
+import XCTest
+
+final class SessionTests: TestCase {
+  override func setUp() {
+    super.setUp()
+//    record=true
+  }
+
+  func testEncodable() {
+    var session: Session
```
mbrandonw: im doing some local variable rewrite stuff so that I can keep the `_assertInlineSnapshot` nice and tidy. I had started by using new lines with `_assertInlineSnapshot` but then it doesn't find the `with` argument.

Sources/PointFree/Session.swift
```diff
@@ -50,11 +50,122 @@ extension URLRequest {
...
+          ghosteeId: try container.decode(Models.User.Id.self, forKey: .ghosteeId),
+          ghosterId: try container.decode(Models.User.Id.self, forKey: .ghosterId)
+        )
+      }
+    }
+  }
+}
+
+extension Session: Codable {
+  private enum CodingKeys: CodingKey {
```
stephencelis: Wow these don't need to "inherit" from `String`? TIL.

Sources/PointFree/Session.swift
mbrandonw: it would only re-login if you are on a page that requires authentication. otherwise you will just be silently logged out.

Sources/PointFree/Session.swift
mbrandonw: ah yeah, I had a deprecated `set` for backwards compatibility but was able to kill it last minute

Sources/PointFree/SiteMiddleware.swift
mbrandonw: lolll

Sources/PointFree/Session.swift
mbrandonw: I think only when you need to change their names

Sources/PointFree/Session.swift
stephencelis: Since we don't care about the error what if we did:
``` swift
self.user = (try? container.decode(Models.User.Id.self, forKey: .userId)).map(User.standard)
  ?? (try container.decode(Session.User.self, forKey: .user))
```

Sources/PointFree/SiteMiddleware.swift
mbrandonw: lolll

Sources/PointFree/Session.swift
stephencelis: K. I just finished reading and the extra code is not so bad.

Sources/PointFree/Session.swift
mbrandonw: 👍 done

PR #494, opened 2019-09-02
Add a foreign key for user's subscription id.

mbrandonw:
I think this is just something we overlooked a long time ago. The users table `subscription_id` wasn't an actual foreign key pointing to `subscriptions.id`. This is never really problematic in production because we never delete subscriptions from the table, but it's still nice to have this constraint, and it can be useful for local development when messing with data manually.

https://trello.com/c/OFf9g5Qk


mbrandonw:
> Looks good... any worry about the migration messing things up?

nah I exported/imported the production data base locally and ran, and we're good

PR #495, opened 2019-09-02
Remove old RSS feature.

mbrandonw:
When we did the private RSS feature long ago we made it a feature so that we could roll it out silently. We can now delete the old RSS code since everyone is now using the new stuff.


Sources/PointFree/AtomFeed.swift
```diff
@@ -96,6 +77,12 @@ private func items() -> [RssItem] {
 
 private func item(episode: Episode) -> RssItem {
 
+  func title(episode: Episode) -> String {
+    return episode.subscriberOnly
+      ? episode.title
+      : "🆓 \(episode.title)"
```
mbrandonw: makes our free episodes pop a bit more

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -92,11 +92,7 @@ private func render(conn: Conn<StatusLineOpen, T3<(Models.Subscription, Enterpri
       return conn.map(const(payload))
         |> expressUnsubscribeReplyMiddleware
 
-    case .feed(.atom):
-      return conn.map(const(Current.episodes()))
-        |> atomFeedResponse
-
-    case .feed(.episodes):
+    case .feed(.atom), .feed(.episodes):
```
mbrandonw: we'll still support the old url, we'll just render the new feed.

Tests/PointFreeTests/AccountTests/AccountTests.swift
```diff
@@ -309,7 +288,11 @@ final class AccountTests: TestCase {
   }
 
   func testAccountCanceledSubscription() {
-    update(&Current, \.stripe.fetchSubscription .~ const(pure(.canceled)))
+    update(
+      &Current,
+      \.stripe.fetchSubscription .~ const(pure(.canceled)),
+      \.database.fetchSubscriptionById .~ const(pure(.canceled))
```
mbrandonw: ive had an old Trello card to fix this test and I finally am. basically this snapshot would render with the "private RSS feed" link on the account page, even tho the subscription had been canceled. I wasn't mocking the right thing.

Tests/PointFreeTests/__Snapshots__/AtomFeedTests/testAtomFeed.1.Conn.txt
```diff
@@ -2,71 +2,228 @@ GET http://localhost:8080/feed/atom.xml
 Cookie: pf_session={}
 
 200 OK
-Content-Length: 1562
-Content-Type: application/atom+xml
+Content-Length: 7488
+Content-Type: text/xml; charset=utf-8
```
stephencelis: Is this a change we want?

stephencelis: I guess we've been doing `text/xml` for the public feature huh.

mbrandonw: yeah seems to be how the RSS feed has been working for awhile. we can switch back if it makes a difference, but doesn't seem too.

PR #496, opened 2019-09-02
Force macOS 10.13 for the project.

mbrandonw:
While working on #495 I realized that we weren't running all tests on linux, which made me realize that some tests were failing on linux due to json encoding. We can just force the generated project to be macOS 10.13 now so that we can properly sort json encoding keys.

### DIFF
--- a/Sources/GitHub/Client.swift
+++ b/Sources/GitHub/Client.swift
@@ -91,11 +91,7 @@ private func runGitHub<A>(_ logger: Logger?) -> (DecodableRequest<A>) -> EitherI
 
 private let gitHubJsonEncoder: JSONEncoder = { () in
   let encoder = JSONEncoder()
-
-  if #available(OSX 10.13, *) {
-    encoder.outputFormatting = [.sortedKeys]
-  }
-
+  encoder.outputFormatting = [.sortedKeys]
   return encoder
 }()
 
--- a/Sources/PointFreeTestSupport/TestCase.swift
+++ b/Sources/PointFreeTestSupport/TestCase.swift
@@ -45,7 +45,7 @@ open class TestCase: XCTestCase {
   }
 
   public var isScreenshotTestingAvailable: Bool {
-    if #available(OSX 10.13, *), ProcessInfo.processInfo.environment["CIRCLECI"] == nil {
+    if ProcessInfo.processInfo.environment["CIRCLECI"] == nil {
       return true
     } else {
       return false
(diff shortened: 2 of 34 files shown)


Tests/PointFreeTests/ApiTests.swift
```diff
@@ -24,7 +24,9 @@ final class ApiTests: TestCase {
       |> siteMiddleware
       |> Prelude.perform
 
+    #if !os(Linux)
     assertSnapshot(matching: conn, as: .conn)
+    #endif
```
mbrandonw: I'm getting this weird diff in linux for this test:

```diff
   "id" : 1,
   "image" : "",
   "length" : 1380,
   "previousEpisodesInCollection" : [
-
+    ¬
   ],
   "publishedAt" : 519652800,
   "references" : [
-
+    ¬
   ],
   "sequence" : 1,
   "subscriberOnly" : false,
   "title" : "Type-Safe HTML in Swift",
```

linux seems to add spaces to the empty array??!!

Sources/GitHub/Client.swift
```diff
@@ -91,11 +91,7 @@ private func runGitHub<A>(_ logger: Logger?) -> (DecodableRequest<A>) -> EitherI
 
 private let gitHubJsonEncoder: JSONEncoder = { () in
   let encoder = JSONEncoder()
-
-  if #available(OSX 10.13, *) {
-    encoder.outputFormatting = [.sortedKeys]
-  }
-
+  encoder.outputFormatting = [.sortedKeys]
```
mbrandonw: found a few more spots we can remove 10.13 checking

Tests/PointFreeTests/ApiTests.swift
```diff
@@ -24,7 +24,10 @@ final class ApiTests: TestCase {
       |> siteMiddleware
       |> Prelude.perform
 
+    #if !os(Linux)
+    // Necessary because of https://bugs.swift.org/browse/SR-11410
```
mbrandonw: filed a bug report

Sources/PointFreeTestSupport/TestCase.swift
```diff
@@ -43,4 +43,12 @@ open class TestCase: XCTestCase {
     super.tearDown()
     record = false
   }
+
+  public var isScreenshotTestingAvailable: Bool {
+    if ProcessInfo.processInfo.environment["CIRCLECI"] == nil {
```
mbrandonw: created a helper for this logic and dropped the macOS 10.13 availability check

PR #498, opened 2019-09-04
Easier team owner removal from subscription confirmation page

mbrandonw:
This makes it easy for owners to remove themselves from the team before subscribing. It does this by adding a "Remove yourself" button next to the owner's name, and clicking that button will just swap out that row for a standard teammate row.

image


Package.swift
```diff
@@ -240,7 +240,9 @@ let package = Package(
     .testTarget(
       name: "PointFreeRouterTests",
       dependencies: [
+        "Models",
         "PointFreeRouter",
+        "SnapshotTesting",
```
mbrandonw: unrelated to the PR, but we were missing some dependencies. I guess everything just worked by chance ¯\\_(ツ)_/¯

Sources/Database/Database.swift
```diff
@@ -14,7 +14,7 @@ public struct Client {
   public var createEnterpriseAccount: (String, EnterpriseAccount.Domain, Models.Subscription.Id) -> EitherIO<Error, EnterpriseAccount?>
   public var createEnterpriseEmail: (EmailAddress, User.Id) -> EitherIO<Error, EnterpriseEmail?>
   public var createFeedRequestEvent: (FeedRequestEvent.FeedType, String, Models.User.Id) -> EitherIO<Error, Prelude.Unit>
-  public var createSubscription: (Stripe.Subscription, Models.User.Id) -> EitherIO<Error, Models.Subscription?>
+  public var createSubscription: (Stripe.Subscription, Models.User.Id, Bool) -> EitherIO<Error, Models.Subscription?>
```
mbrandonw: this bool represents if the owner of the subscription is going to take a seat on the subscription.

Sources/Database/Database.swift
```diff
@@ -276,7 +278,9 @@ private struct _Client {
         ])
 
       return subscription.flatMap { subscription in
-        self.execute(
+        guard isOwnerTakingSeat else { return pure(subscription) }
```
mbrandonw: and here is where we implement that logic in the database. 

it kinda seems weird to have a single database endpoint do two queries with logic tbh. maybe this should be split up.

Sources/Models/SubscribeConfirmationData.swift
```diff
@@ -3,13 +3,16 @@ import Stripe
 
 public struct SubscribeConfirmationData: Equatable {
   public var billing: Pricing.Billing
+  public var isOwnerTakingSeat: Bool
```
mbrandonw: this is to allow restoring this piece of state when we redirect back to the subscription confirmation page in case of a stripe error.

Sources/Models/SubscribeData.swift
```diff
@@ -3,24 +3,28 @@ import Stripe
 
 public struct SubscribeData: Equatable {
   public var coupon: Stripe.Coupon.Id?
+  public var isOwnerTakingSeat: Bool
```
mbrandonw: this is to represent this state in the data that is submitted with the subscription form. we can use this to determine if we assign the subscription to the owner in the DB or not.

Sources/Models/SubscribeData.swift
```diff
@@ -3,24 +3,28 @@ import Stripe
...
   ) {
     self.coupon = coupon
+    self.isOwnerTakingSeat = isOwnerTakingSeat
     self.pricing = pricing
     self.teammates = teammates
     self.token = token
   }
 
-  private enum CodingKeys: String, CodingKey {
+  public enum CodingKeys: String, CodingKey {
```
mbrandonw: made it public so that I could use these values when building form inputs. helps keep those things in 1-1.

Sources/PointFree/Subscribe.swift
```diff
@@ -54,7 +54,7 @@ private func subscribe(_ conn: Conn<StatusLineOpen, Tuple2<SubscribeData, User>>
           let parallel = sequence(
             subscribeData.teammates
               .filter { email in email.rawValue.contains("@") && email != user.email }
-              .prefix(subscribeData.pricing.quantity - 1)
+              .prefix(subscribeData.pricing.quantity - (subscribeData.isOwnerTakingSeat ? 1 : 0))
```
mbrandonw: we need to truncate this data differently depending on whether or not the owner is taking a seat. a test actually caught this which is nice.

Sources/PointFree/Subscribe.swift
```diff
@@ -112,13 +112,19 @@ private func validateCoupon(forSubscribeData subscribeData: SubscribeData) -> Bo
 
 private func subscribeConfirmationWithSubscribeData(_ subscribeData: SubscribeData?) -> Route {
   guard let subscribeData = subscribeData else {
-    return .subscribeConfirmation(.team, .yearly, [""])
+    return .subscribeConfirmation(
+      lane: .team,
+      billing: .yearly,
+      isOwnerTakingSeat: true,
+      teammates: [""]
+    )
```
mbrandonw: I added labels to this enum case's arguments

Sources/PointFree/SubscribeConfirmation.swift
```diff
@@ -33,7 +33,7 @@ public let subscribeConfirmation: Middleware<
             coupon,
             currentUser,
             Current.stripe.js,
-            Current.envVars.stripe.publishableKey.rawValue
+            Current.envVars.stripe.publishableKey
```
mbrandonw: I upgraded the middleware to take a `Stripe.Client.PublishableKey` instead of just the raw string value.

Sources/PointFreeRouter/Routes.swift
```diff
@@ -300,6 +306,11 @@ private let subscriberDataIso = PartialIso<String, SubscribeData?>(
         return nil
     }
 
+    let isOwnerTakingSeat = keyValues
+      .first { key, value in key == SubscribeData.CodingKeys.isOwnerTakingSeat.rawValue }?.1
+      .flatMap(Bool.init)
+      ?? false
```
mbrandonw: just adding onto this partial iso hack :/ but I've written some tests for it.

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -29,7 +29,7 @@ public func subscriptionConfirmation(
         style(maxWidth(.px(900)) <> margin(leftRight: .auto)),
       ],
       header(lane)
-        + (lane == .team ? teamMembers(currentUser, subscribeData) : [])
+        + teamMembers(lane: lane, currentUser: currentUser, subscribeData: subscribeData)
```
mbrandonw: I moved the `lane == .team` logic into this view function. I did this because in the case of individual subscriptions I need to render a hidden input (for `isOwnerTakingSeat`) and this view function seemed like a natural place to put it.

up for other ideas or if you think this function should be renamed!

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -81,7 +81,22 @@ private func header(_ lane: Pricing.Lane) -> [Node] {
...
+  subscribeData: SubscribeConfirmationData
+  ) -> [Node] {
+
+  guard lane == .team else {
+    return [
+      input([
+        name(SubscribeData.CodingKeys.isOwnerTakingSeat.rawValue),
+        `type`(.hidden),
+        value("true")
+        ])
```
mbrandonw: I need to include this hidden input for individual subscriptions so that it gets properly set in the `SubscribeData` posted to the server.

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -111,6 +126,11 @@ private func teamMembers(_ currentUser: User, _ subscribeData: SubscribeConfirma
                   [("id", "team-member-template")],
                   [teamMemberTemplate("", withRemoveButton: true)]
                 ),
+                .element(
+                  "template",
+                  [("id", "team-member-template-without-remove")],
+                  [teamMemberTemplate("", withRemoveButton: false)]
+                ),
```
mbrandonw: I need a template of the team member row without a remove button because that is what will replace the owner if they decide to remove themselves.

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -147,18 +167,30 @@ updateSeats()
               ])
           ],
           [
-            "You can add additional team members at anytime from your account page."
+            """
+You must have at least two seats for your team subscription. You can add additional team members at anytime
+from your account page.
```
mbrandonw: I don't support an owner adding themselves back, but maybe it's not worth supporting. but open to other ways of phrasing this copy

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -147,18 +167,30 @@ updateSeats()
...
 }
 
-private func teamOwner(_ currentUser: User) -> Node {
+private func teamOwner(currentUser: User, subscribeData: SubscribeConfirmationData) -> Node {
+  guard subscribeData.isOwnerTakingSeat else {
+    return input([
+      name(SubscribeData.CodingKeys.isOwnerTakingSeat.rawValue),
+      `type`(.hidden),
+      value("false")
+      ])
```
mbrandonw: this won't render the owner row if they aren't taking a seat. this is used when redirecting back to this page after a failed subscription.

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -147,18 +167,30 @@ updateSeats()
...
+  }
+
   return gridColumn(
     sizes: [.mobile: 12],
     [
+      id("team-owner"),
```
mbrandonw: not ideal to add more ids to the page, but I needed this so that I could check if the owner removed themselves from javascript.

an alternative is to look up the input element `input[name=isOwnerTakingSeat]`. thoughts on approaches?

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -187,7 +224,31 @@ private func teamOwner(_ currentUser: User) -> Node {
...
+              ]),
+            onclick("""
+var ownerRow = this.parentNode.parentNode
+ownerRow.parentNode.removeChild(ownerRow)
+
+var teamMember = document.getElementById("team-member-template-without-remove").content.cloneNode(true)
+var teamMembersContainer = document.getElementById("team-members")
+teamMembersContainer.insertBefore(teamMember, teamMembersContainer.firstChild)
+
+updateSeats()
```
mbrandonw: could use some 👀 on this js. it removes the owner row and then adds a new team member row.

Tests/PointFreeTests/SubscribeConfirmationTests.swift
```diff
@@ -237,6 +344,42 @@ class SubscriptionConfirmationTests: TestCase {
...
+
+    #if !os(Linux)
+    if self.isScreenshotTestingAvailable {
+      let webView = WKWebView(frame: .init(x: 0, y: 0, width: 1100, height: 1600))
+      let html = String(decoding: result.perform().data, as: UTF8.self)
+      webView.loadHTMLString(html, baseURL: nil)
+
+      assertSnapshot(
+        matching: webView,
+        as: .image(afterEvaluatingJavascript: "document.getElementById('remove-yourself-button').click()"),
```
mbrandonw: this tests what happens when a user actually removes themselves from the subscription.

Tests/PointFreeTests/SubscribeTests.swift
```diff
@@ -152,16 +152,6 @@ final class SubscribeTests: TestCase {
     #endif
   }
 
-  // TODO: dont know how to get this route to recognize.
-  //  func testMissingSubscriberData() {
```
mbrandonw: this seems to be old. cool to remove?

Sources/Views/SubscriptionConfirmation.swift
stephencelis: Nah I think this is fine for now.

Sources/Views/SubscriptionConfirmation.swift
stephencelis: It'd be nice but can be done later.

Sources/Views/SubscriptionConfirmation.swift
stephencelis: I think id is just fine. I'll think on other approaches...but really revamping the way we do JS seems like a larger issue.

Sources/Views/SubscriptionConfirmation.swift
stephencelis: Ha looks right but gonna check out the branch and play with it after initial review.

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -147,18 +167,30 @@ updateSeats()
               ])
           ],
           [
-            "You can add additional team members at anytime from your account page."
+            """
+You must have at least two seats for your team subscription. You can add additional team members at anytime
```
stephencelis: Sorry, snaps will change too if you take this:
```suggestion
You must have at least two seats for your team subscription. You can add additional team members at any time
```

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -187,7 +224,31 @@ private func teamOwner(_ currentUser: User) -> Node {
...
+var ownerRow = this.parentNode.parentNode
+ownerRow.parentNode.removeChild(ownerRow)
+
+var teamMember = document.getElementById("team-member-template-without-remove").content.cloneNode(true)
+var teamMembersContainer = document.getElementById("team-members")
+teamMembersContainer.insertBefore(teamMember, teamMembersContainer.firstChild)
+
+updateSeats()
+""")
+            ], [.raw("Remove&nbsp;yourself")])
```
stephencelis: Maybe something else, like "Don't include yourself as a seat", "Don't include yourself", or "I don’t want a seat".
```suggestion
            ], [.raw("Don’t&nbsp;include&nbsp;yourself&nbsp;as&nbsp;a&nbsp;seat")])
```

stephencelis: I guess it looks nicer and makes more sense when you add more rows: 
image

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -147,22 +167,35 @@ updateSeats()
...
   return gridColumn(
     sizes: [.mobile: 12],
     [
+      id("team-owner"),
       `class`([
         Class.border.all,
         Class.pf.colors.border.gray850,
-        Class.padding([.mobile: [.all: 2]])
+        Class.padding([.mobile: [.all: 2]]),
+        Class.margin([.mobile: [.top: 1]]),
```
stephencelis: Added this to match the padding that the team cells have. Makes clicking "remove self" a bit more of a seamless transition. (Did not regenerate snaps...but can if no more changes are coming.)

PR #500, opened 2019-09-04
More webhook failure messaging.

mbrandonw:
We're getting a weird web hook failure that doesn't have any error messaging because the error is erased to `Unit`. I'm adding some more logging so that we can debug this.

### DIFF
--- a/Sources/Database/Database.swift
+++ b/Sources/Database/Database.swift
@@ -297,7 +297,7 @@ private struct _Client {
     UPDATE "subscriptions"
     SET "stripe_subscription_status" = $1
     WHERE "subscriptions"."stripe_subscription_id" = $2
-    RETURNING "id", "stripe_subscription_id", "stripe_subscription_status", "user_id"
+    RETURNING *
     """,
       [
         stripeSubscription.status.rawValue,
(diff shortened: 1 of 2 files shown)


Sources/Database/Database.swift
```diff
@@ -297,7 +297,7 @@ private struct _Client {
     UPDATE "subscriptions"
     SET "stripe_subscription_status" = $1
     WHERE "subscriptions"."stripe_subscription_id" = $2
-    RETURNING "id", "stripe_subscription_id", "stripe_subscription_status", "user_id"
+    RETURNING *
```
mbrandonw: this just seems like a good idea in general

Sources/PointFree/StripeWebhooks.swift
```diff
@@ -94,11 +94,14 @@ private func handleFailedPayment(
   -> IO<Conn<ResponseEnded, Data>> {
 
     return Current.stripe.fetchSubscription(conn.data)
+      .withExcept(notifyError(subject: "Stripe Hook failed: Couldn't find stripe subscription."))
       .flatMap(Current.database.updateStripeSubscription)
       .mapExcept(requireSome)
+      .withExcept(notifyError(subject: "Stripe Hook failed: Couldn't find updated subscription."))
```
mbrandonw: it's the `requireSome`'s that are turning the error into `Prelude.Unit`:

https://github.com/pointfreeco/pointfreeco/blob/a57a4c8b1aa4dd183ab78cb566024fef0e13b65d/Sources/PointFreePrelude/Either.swift#L18

PR #501, opened 2019-09-06
Slight simplification to subscribe logic.

mbrandonw:
While working on the subscribe stuff for #498 I was seeing that the subscribe middleware could be slightly simplified. First, it could be flattened by not starting off with a pure `EitherIO` value. Then, I believe we should first create the subscription in the DB and _then_ try sending out the invites. Otherwise an error in sending emails will mess up creating the DB row. Also, I think sending the invites shouldn't result result in any errors, i.e. if an email fails to send it shouldn't cause any errors to show.

### DIFF
--- a/Tests/PointFreeTests/__Snapshots__/SubscribeTests/testHappyPath_Team.2.txt
+++ b/Tests/PointFreeTests/__Snapshots__/SubscribeTests/testHappyPath_Team.2.txt
@@ -1,6 +1,6 @@
 ▿ Subscription
-  ▿ id: 00000000-0000-0000-0000-000000000007
-    - rawValue: 00000000-0000-0000-0000-000000000007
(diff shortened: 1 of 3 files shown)


Tests/PointFreeTests/__Snapshots__/SubscribeTests/testHappyPath_Team.2.txt
```diff
@@ -1,6 +1,6 @@
 ▿ Subscription
-  ▿ id: 00000000-0000-0000-0000-000000000007
-    - rawValue: 00000000-0000-0000-0000-000000000007
+  ▿ id: 00000000-0000-0000-0000-000000000003
+    - rawValue: 00000000-0000-0000-0000-000000000003
```
mbrandonw: had to update snaps. these are because the subscription is now created before the team invites.

PR #503, opened 2019-09-07
Remove "View/Func" dependency

stephencelis:
We talked about getting our "Html" version up-to-date and eliminating this abstraction/dependency seems like a good first step. Started this last night. Mostly a mindless task but took longer than I expected.

### DIFF
--- a/Sources/PointFree/Account/Cancel.swift
+++ b/Sources/PointFree/Account/Cancel.swift
@@ -206,24 +206,23 @@ private func sendCancelEmail(to owner: User, for subscription: Stripe.Subscripti
     return sendEmail(
       to: [owner.email],
       subject: "Your subscription has been canceled",
-      content: inj2(cancelEmailView.view((owner, subscription)))
+      content: inj2(cancelEmailView((owner, subscription)))
     )
 }
 
-let cancelEmailView = simpleEmailLayout(cancelEmailBodyView)
-  .contramap { owner, subscription in
-    SimpleEmailLayoutData(
-      user: nil,
-      newsletter: nil,
-      title: "Your subscription has been canceled",
-      preheader: """
-      Your \(subscription.plan.nickname) subscription has been canceled and will remain active through
-      \(dateFormatter.string(from: subscription.currentPeriodEnd)).
-      """,
-      template: .default,
-      data: (owner, subscription)
-    )
-}
+let cancelEmailView = { owner, subscription in
+  SimpleEmailLayoutData(
+    user: nil,
+    newsletter: nil,
+    title: "Your subscription has been canceled",
+    preheader: """
+    Your \(subscription.plan.nickname) subscription has been canceled and will remain active through
+    \(dateFormatter.string(from: subscription.currentPeriodEnd)).
+    """,
+    template: .default,
+    data: (owner, subscription)
+  )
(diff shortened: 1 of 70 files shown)


Sources/PointFree/Admin/FreeEpisodes.swift
```diff
@@ -52,7 +51,7 @@ private func sendEmail(forFreeEpisode episode: Episode, toUsers users: [User]) -
 
   // A personalized email to send to each user.
   let freeEpisodeEmails = users.map { user in
-    lift(IO { inj2(freeEpisodeEmail.view((episode, user))) })
+    lift(IO { inj2(freeEpisodeEmail((episode, user))) })
```
stephencelis: Email views are still pullback-composed, so the tupling prevails.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -207,44 +206,45 @@ private func sendCancelEmail(to owner: User, for subscription: Stripe.Subscripti
...
+    newsletter: nil,
+    title: "Your subscription has been canceled",
+    preheader: """
+    Your \(subscription.plan.nickname) subscription has been canceled and will remain active through
+    \(dateFormatter.string(from: subscription.currentPeriodEnd)).
+    """,
+    template: .default,
+    data: (owner, subscription)
+  )
+  } >>> simpleEmailLayout(cancelEmailBodyView)
```
mbrandonw: super nitty, but what do you think about doing `simpleEmailLayout(cancelEmailBodyView) <<<` at the beginning instead? kinda of nice to have the layout at the top

stephencelis: Can do. Just was `>>>`ing everything.

Sources/PointFree/Account/PrivateRss.swift
```diff
@@ -159,11 +158,13 @@ private func fetchStripeSubscriptionForUser<A>(
     }
 }
 
-private let privateEpisodesFeedView = itunesRssFeedLayout <| View<(Stripe.Subscription?, User)> { subscription, user -> Node in
-  node(
-    rssChannel: privateRssChannel(user: user),
-    items: items(forUser: user, subscription: subscription)
-  )
+private let privateEpisodesFeedView = itunesRssFeedLayout { (data: (subscription: Stripe.Subscription?, user: User)) -> [Node] in
```
mbrandonw: oh interesting. I never thought about using this syntax for a layout. could we even replace the other `layout <<< { ... }`'s with just `layout { ... }` ?

stephencelis: We'd need to un-curry `simpleEmailLayout`. Worth doing but let's kick to another PR.

Sources/PointFree/Admin/FreeEpisodes.swift
mbrandonw: yeahhh

mbrandonw:
oh I guess the `Package.swift` could be updated too?

stephencelis:
So nice to have snapshot tests! One gnarly part of the refactor got caught. There may be more, but as soon as all the snaps pass will merge!

mbrandonw:
Wow very nice to catch that footer bug.

PR #504, opened 2019-09-07
Add upgraded Html dependency

stephencelis:
I've forked `Html` and renamed it to `HtmlUpgrade` so we can incrementally make changes. This PR merely adds the dependency.

The path forward to me seems to be:

1. Add a second HTML renderer for the newer library to `Current`
2. Duplicate any helper methods that render the older Html nodes.
3. Start piece-meal upgrading things.


Package.swift
```diff
@@ -33,6 +33,7 @@ let package = Package(
     .package(url: "https://github.com/ianpartridge/swift-backtrace.git", .exact("1.1.0")),
     .package(url: "https://github.com/pointfreeco/Ccmark.git", .branch("master")),
     .package(url: "https://github.com/pointfreeco/swift-html.git", .exact("0.2.1")),
+    .package(url: "https://github.com/stephencelis/swift-html-1.git", .branch("pointfreeco-upgrade")),
```
stephencelis: I guess I may have already had a fork? SPM requires the git URL to be different so couldn't merely add another branch to `pointfreeco/swift-html` even if the module names don't overlap.

PR #506, opened 2019-09-14
Proof-of-concept for HTML upgrade path

mbrandonw:
(NB: this is based off of #504)

Shows how we can incremental update our views to the newest HTML library. It does the following:

* Introduces a new render function for upgraded HTML nodes
* Provides a function to "downgrade" nodes from the lib to the old lib
* Provides a new `respond` function for upgraded nodes, and under the hood it just downgrades the nodes and passes it along to the current `respond`
* Converts the about screen to the new nodes


Sources/FunctionalCss/GridHelpersUpgrade.swift
```diff
@@ -0,0 +1,53 @@
+import Css
+import HtmlUpgrade
+
+// TODO: extract to grid helpers in design systems?
+public func _gridRow(_ attribs: [Attribute<Tag.Div>], _ content: [Node]) -> Node {
```
mbrandonw: we need to duplicate any helper functions we have on `Node`, but I don't think this will be so bad (luckily we dont have to touch CSS). I am adding underscores to the duplicated functions to help with type checking times.

mbrandonw:
> This looks good! One potential change is to use statics instead of `_`-prefixed functions. Should be the same compile-check difference but will be closer to the final form of the library.

good idea, done! will merge when green

PR #507, opened 2019-09-14
Html Upgrade Tweaks

stephencelis:
Removes the underscores, unwraps a few arrays, and cleans up the about page.

### DIFF
--- a/Sources/FunctionalCss/GridHelpersUpgrade.swift
+++ b/Sources/FunctionalCss/GridHelpersUpgrade.swift
@@ -3,33 +3,29 @@ import HtmlUpgrade
 
 // TODO: extract to grid helpers in design systems?
 extension Node {
-  public static func _gridRow(_ attribs: [Attribute<Tag.Div>], _ content: [Node]) -> Node {
-    return .div(attributes: _addClasses([Class.grid.row], to: attribs), .fragment(content))
+  public static func gridRow(attributes: [Attribute<Tag.Div>] = [], _ content: Node...) -> Node {
+    return .div(attributes: _addClasses([Class.grid.row], to: attributes), .fragment(content))
   }
 
   // TODO: extract to grid helpers in design systems?
-  public static func _gridRow(_ content: [Node]) -> Node {
-    return ._gridRow([], content)
-  }
-
-  // TODO: extract to grid helpers in design systems?
-  public static func _gridColumn(sizes: [Breakpoint: Int]) -> ([Node]) -> Node {
+  public static func gridColumn(sizes: [Breakpoint: Int]) -> (Node) -> Node {
     return { content in
-      ._gridColumn(sizes: sizes, [], content)
+      .gridColumn(sizes: sizes, attributes: [], content)
     }
   }
 
-  public static func _gridColumn(sizes: [Breakpoint: Int], _ content: [Node]) -> Node {
-    return ._gridColumn(sizes: sizes, [], content)
-  }
+  public static func gridColumn(
+    sizes: [Breakpoint: Int],
+    attributes: [Attribute<Tag.Div>] = [],
+    _ content: Node...
+    ) -> Node {
 
-  public static func _gridColumn(sizes: [Breakpoint: Int], _ attribs: [Attribute<Tag.Div>], _ content: [Node]) -> Node {
     let classes = [Class.grid.col(.mobile, nil)]
       + sizes
         .sorted(by: { $0.key.rawValue < $1.key.rawValue })
         .map(Class.grid.col(_:_:))
 
-    return .div(attributes: _addClasses(classes, to: attribs), .fragment(content))
+    return .div(attributes: _addClasses(classes, to: attributes), .fragment(content))
   }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/FunctionalCss/GridHelpersUpgrade.swift
```diff
@@ -3,33 +3,29 @@ import HtmlUpgrade
 
 // TODO: extract to grid helpers in design systems?
 extension Node {
-  public static func _gridRow(_ attribs: [Attribute<Tag.Div>], _ content: [Node]) -> Node {
-    return .div(attributes: _addClasses([Class.grid.row], to: attribs), .fragment(content))
+  public static func gridRow(attributes: [Attribute<Tag.Div>] = [], _ content: Node...) -> Node {
```
stephencelis: Updated the function to match the library a bit more by removing the overload method and give attributes a default.

Sources/FunctionalCss/GridHelpersUpgrade.swift
```diff
@@ -3,33 +3,22 @@ import HtmlUpgrade
...
+    return .div(attributes: _addClasses([Class.grid.row], to: attributes), .fragment(content))
   }
 
-  // TODO: extract to grid helpers in design systems?
-  public static func _gridRow(_ content: [Node]) -> Node {
-    return ._gridRow([], content)
-  }
-
-  // TODO: extract to grid helpers in design systems?
-  public static func _gridColumn(sizes: [Breakpoint: Int]) -> ([Node]) -> Node {
```
stephencelis: Removed this curried overload for now...assuming we don't want it.

Sources/PointFree/PageLayout.swift
```diff
@@ -107,13 +107,13 @@ struct SimplePageLayoutData<A> {
 }
 
 func _respond<A, B>(
-  view: @escaping (B) -> [HtmlUpgrade.Node],
+  view: @escaping (B) -> HtmlUpgrade.Node,
```
stephencelis: Now that `Node` has `fragment` I think we should rarely (if ever) work with `[Node]`.

Sources/Views/About.swift
```diff
@@ -29,38 +29,37 @@ public struct Host {
...
-    ._gridRow([
-      ._gridColumn(sizes: [.mobile: 12, .desktop: 7], [
-        .div(
-          attributes: [.class([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]])])],
-          .fragment(aboutSectionView + openSourceSection)
-        )
-        ]),
+public func aboutView(hosts: [Host]) -> Node {
+  return .gridRow(
+    .gridColumn(
```
stephencelis: Can flatten a lot of the array stuff now and work with `Node...` now.

Sources/Views/About.swift
```diff
@@ -29,38 +29,37 @@ public struct Host {
...
+        ],
+        hostsView(hosts: hosts)
+      )
+    )
+  )
 }
 
 private func hostsView(hosts: [Host]) -> Node {
-  return .fragment([
+  return [
```
stephencelis: `Node` is `ExpressibleByArrayLiteral`.

Sources/Views/About.swift
```diff
@@ -71,77 +70,79 @@ private func hostsView(hosts: [Host]) -> Node {
...
+         "open source"),
       " the ",
       .a(attributes: [.class([Class.pf.colors.link.green, Class.type.underline]), .href("https://www.kickstarter.com")], "Kickstarter"),
       " mobile apps."
-      ])
-    ]
-    + hosts.flatMap(hostView)
-  )
+    ),
+    .fragment(hosts.map(hostView))
```
stephencelis: I think generally we only need to use `fragment` directly is during `map`.

Sources/Views/About.swift
```diff
@@ -29,38 +29,37 @@ public struct Host {
...
-        )
-        ]),
+public func aboutView(hosts: [Host]) -> Node {
+  return .gridRow(
+    .gridColumn(
+      sizes: [.mobile: 12, .desktop: 7],
+      .div(
+        attributes: [.class([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]])])],
+        aboutSectionView,
+        openSourceSection
```
stephencelis: When both of these return `Node` instead they can be inlined, and no need to `.fragment(a + b)`.

Sources/FunctionalCss/GridHelpersUpgrade.swift
mbrandonw: yeahhhh there are maybe a few uses of it but we can just remove as we come across them

Sources/Views/About.swift
mbrandonw: amazing

Sources/FunctionalCss/GridHelpersUpgrade.swift
stephencelis: ha let's revisit it when we need it.

PR #508, opened 2019-09-15
Upgrade Subscribe Confirmation HTML

stephencelis:
Working with some of this HTML on the account page, so upgraded it.


Tests/PointFreeTests/__Snapshots__/SubscribeConfirmationTests/testPersonal_LoggedIn.1.Conn.txt
```diff
@@ -2140,8 +2140,6 @@ html {
                        fg-gray400"
                 id="pricing-preview">
           </span>
-          <span>
-          </span>
```
mbrandonw: nice that we can get rid of this hack

PR #512, opened 2019-09-15
Upgrade Blog to new HTML library.

mbrandonw:


Sources/PointFree/Blog/BlogAtomFeed.swift
```diff
@@ -28,7 +29,7 @@ private func feedView(posts: [BlogPost]) -> [Node] {
 
 private func atomEntry(for post: BlogPost) -> AtomEntry {
   return AtomEntry(
-    content: blogPostContentView(post),
+    content: downgrade(node: blogPostContentView(post)),
```
mbrandonw: need to do a bit of downgrade/upgrade dancing in this PR to accommodate for some shared views that I don't want to convert yet.

Sources/PointFree/PageLayout.swift
```diff
@@ -341,20 +341,3 @@ private let prismJsHead: [Html.ChildOf<Html.Tag.Head>] = [
     """
   )
 ]
-
-private func downgrade(node: HtmlUpgrade.Node) -> [Html.Node] {
```
mbrandonw: made this public and moved to `Styleguide` so that more modules could get access to it.

Sources/Styleguide/Components.swift
```diff
@@ -47,8 +48,30 @@ public func gitHubLink(text: String, type: GitHubLinkType, href: String?) -> Nod
   )
 }
 
-public func twitterShareLink(text: String, url: String, via: String? = nil) -> Node {
+extension HtmlUpgrade.Node {
+  public static func gitHubLink(text: String, type: GitHubLinkType, href: String?) -> HtmlUpgrade.Node {
```
mbrandonw: moved this from the enterprise file to here so it could be shared

PR #515, opened 2019-09-22
Upgrade home to new HTML

mbrandonw:
Upgrades the homepage to work off the new HTML library.


Sources/PointFree/Episode/Show.swift
```diff
@@ -468,7 +468,7 @@ private func episodeInfoView(permission: EpisodePermission, ep: Episode) -> [Nod
   return [
     div(
       [`class`([Class.padding([.mobile: [.all: 3], .desktop: [.all: 4]]), Class.pf.colors.bg.white])],
-      topLevelEpisodeInfoView(ep: ep)
+      downgrade(node: topLevelEpisodeInfoView(episode: ep, date: Current.date))
```
mbrandonw: the `topLevelEpisodeInfoView` view function was moved to the `Views` module to be used with the home stuff, but it also is used on the episode page. so we need to downgrade the nodes before they can be used.

also we have to pass in the current date dependency explicitly (at least until we get the `World` stuff in place).

PR #518, opened 2019-10-02
Bare minimum to support Xcode 11

mbrandonw:
I just wanted to do the bare minimum to get Xcode 11 support in. I'm not deleting the Xcode project or anything. It would be great to figure out how to continue using playgrounds. Even if SwiftUI previews can help us eventually, there's still some good use cases for playgrounds (like playing with the database or router).

### DIFF
--- a/Sources/PointFree/Subscribe.swift
+++ b/Sources/PointFree/Subscribe.swift
@@ -8,28 +8,14 @@ import Prelude
 import Stripe
 import Tuple
 
-let subscribeMiddleware =
-  filterMap(
-    require1 >>> pure,
-    or: redirect(
-      with: get1 >>> subscribeConfirmationWithSubscribeData,
-      headersMiddleware: flash(.error, "Error creating subscription!")
-    )
-    )
-    <<< filter(
-      get1 >>> ^\.pricing >>> validateQuantity,
-      or: redirect(
-        with: get1 >>> subscribeConfirmationWithSubscribeData,
-        headersMiddleware: flash(.error, "An invalid subscription quantity was used.")
-      )
-    )
-    <<< filter(
-      get1 >>> validateCoupon(forSubscribeData:),
-      or: redirect(
-        with: get1 >>> subscribeConfirmationWithSubscribeData,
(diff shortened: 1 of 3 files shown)


Sources/PointFree/Subscribe.swift
```diff
@@ -8,28 +8,14 @@ import Prelude
...
-        with: get1 >>> subscribeConfirmationWithSubscribeData,
-        headersMiddleware: flash(.error, "Coupons can only be used on individual subscription plans.")
-      )
-    )
+let subscribeMiddleware: Middleware<
+  StatusLineOpen,
+  ResponseEnded,
+  Tuple2<SubscribeData?, User?>,
+  Data
+  > = requireSubscribeData
```
mbrandonw: this chain of middlewares was segfaulting.

someday we should try what you did for reducers on middleware. middleware is probably a lot more similar to reducers than views.

stephencelis: Yeah that sounds good.

Sources/PointFree/Subscribe.swift
```diff
@@ -8,28 +8,14 @@ import Prelude
...
+    <<< validateQuanitity
```
stephencelis:
```suggestion
    <<< validateQuantity
```

Sources/PointFree/Subscribe.swift
```diff
@@ -136,3 +122,39 @@ private func subscribeConfirmationWithSubscribeData(_ subscribeData: SubscribeDa
...
+  return middleware |> filterMap(
+    require1 >>> pure,
+    or: redirect(
+      with: get1 >>> subscribeConfirmationWithSubscribeData,
+      headersMiddleware: flash(.error, "Error creating subscription!")
+    )
+  )
+}
+
+private func validateQuanitity(
```
stephencelis:
```suggestion
private func validateQuantity(
```

Sources/PointFree/Subscribe.swift
mbrandonw: oops didn't see this, will push to master...

PR #522, opened 2019-11-20
Fixes type annotation in Combine exercise

fonkadelic:


stephencelis:
:wave: @fonkadelic, thanks for the PR!

The type annotation should actually be okay! In this case `compactMap { $0 }` will filter out `nil`s and promote to a non-optional. If you want to be notified of `nil` failures, though, you can update the signature as you have and remove the `compactMap { $0 }`. It definitely might be nice to flesh out the exercise with this additional path, though, so thanks for thinking of it!

fonkadelic:
Thanks for clarifying @stephencelis!

PR #523, opened 2019-12-01
Fix playgrounds

stephencelis:
This gets all the playgrounds building again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,20 +16,20 @@ bootstrap-oss:
 	@$(MAKE) xcodeproj-oss
 	@$(MAKE) install-mm
 	@echo "  ✅ Bootstrapped! Opening Xcode..."
-	@sleep 1 && xed .
+	@open -g PointFree.xcworkspace
 
 bootstrap-oss-lite:
 	@echo "  ⚠️  Bootstrapping open-source Point-Free (lite)..."
 	@$(MAKE) xcodeproj-oss
 	@echo "  ✅ Bootstrapped! Opening Xcode..."
-	@sleep 1 && xed .
+	@open -g PointFree.xcworkspace
 
 bootstrap-private:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -43,7 +43,6 @@ install-mm:
 	@$(MAKE) install-mm-commoncrypto || (echo "$$MODULE_MAP_ERROR" && exit 1)
 	@$(MAKE) install-mm-cmark
 	@$(MAKE) install-mm-postgres
-	@$(MAKE) install-mm-xcodeproj
```
stephencelis: This step is no longer necessary (or possible) in Xcode 11.

PointFree.playground/Pages/Blog.xcplaygroundpage/Contents.swift
```diff
@@ -7,7 +7,7 @@ import WebKit
 Current = .mock
 let post = Current.blogPosts().first!
 
-let req = request(to: .blog(.show(post)), session: .loggedOut)
+let req = request(to: .blog(.show(.right(post.id))), session: .loggedOut)
```
stephencelis: This is the one playground that doesn't seem to run properly. It renders blanks...

Makefile
```diff
@@ -16,20 +16,20 @@ bootstrap-oss:
 	@$(MAKE) xcodeproj-oss
 	@$(MAKE) install-mm
 	@echo "  ✅ Bootstrapped! Opening Xcode..."
-	@sleep 1 && xed .
+	@open -g PointFree.xcworkspace
```
stephencelis: The `xed` command defaults to opening as an SPM package. Since we still rely on the project file for playgrounds, let's open it using `open` instead. I use the `-g` flag to open it in the background to avoid interrupting anything we might still wanna do in the terminal, but we could open in the foreground instead if you prefer!

PR #525, opened 2019-12-11
Ability to hide blog posts.

mbrandonw:
Blog posts that are temporal, like discounts, should be hidden.


Package.swift
```diff
@@ -155,7 +155,6 @@ let package = Package(
       name: "Models",
       dependencies: [
         "GitHub",
-        "HttpPipeline",
```
mbrandonw: just cleaning up an old dependency

mbrandonw:
I should mention, direct links will still work. it just won't be on the index.

Tests/PointFreeTests/BlogTests.swift
```diff
@@ -23,7 +23,7 @@ class BlogTests: TestCase {
   }
 
   func testBlogIndex() {
-    let conn = connection(from: request(to: .blog(.index), basicAuth: true))
```
mbrandonw: we used basic auth when first first building the blog, but it's no longer necessary.

PR #528, opened 2019-12-15
Coupon language fix

mbrandonw:
I wanted to make the coupon callout on the subscription confirmation page a little more pronounced, and while doing that I realized the correct copy wasn't showing, and so added a bunch of tests. And while added tests I found out that there was a bug, so fixed that.


Sources/Stripe/Model.swift
```diff
@@ -130,7 +130,7 @@ public struct Coupon: Equatable {
     public var formattedDescription: String {
       switch self {
       case let .amountOff(amountOff):
-        return "$\(amountOff) off"
+        return "$\(Int(amountOff.map(Double.init).dollars.rawValue)) off"
```
mbrandonw: we haven't used a $ amount coupon yet, but if we had it would have shown something like "$500 off" instead of "$5 off".

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -593,10 +593,12 @@ private func discountedTotalDisclaimer(coupon: Coupon?) -> Node {
         Class.pf.colors.fg.gray400
         ]),
     ],
+    .br,
+    .br,
```
mbrandonw: Just putting some space here:

image

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -593,10 +593,12 @@ private func discountedTotalDisclaimer(coupon: Coupon?) -> Node {
...
     coupon.name
       .map { .raw(" You are using the coupon <strong>\($0)</strong>") }
       ?? " You are using a coupon",
-    ", which gives you \(coupon.rate.percentOff.map { "\($0)%" } ?? "a fixed discount") off every billing period."
+    ", which gives you \(coupon.formattedDescription)."
```
mbrandonw: we had a helper for this copy that gives more information about the rate and duration.

Tests/PointFreeTests/SubscribeConfirmationTests.swift
```diff
@@ -16,7 +16,7 @@ import XCTest
 class SubscriptionConfirmationTests: TestCase {
   override func setUp() {
     super.setUp()
-//    record = true
+    record = true
```
stephencelis:
```suggestion
//    record = true
```

Tests/PointFreeTests/DiscountsTests.swift
```diff
@@ -29,15 +30,145 @@ class DiscountsTests: TestCase {
...
+      \.database.fetchSubscriptionByOwnerId .~ const(pure(nil)),
+      \.stripe.fetchCoupon .~ const(pure(fiftyPercentOffForever))
+    )
+
+    assertSnapshot(
+      matching: connection(
+        from: request(with: secureRequest("http://localhost:8080/discounts/blobfest"), session: .loggedIn)
+        )
+        |> siteMiddleware,
+      as: .ioConn
```
stephencelis: Worth adding image snaps as well? Or are you worried about repo size?

mbrandonw: it did seem like overkill, and we do have some coupon coverage in `SubscriptionConfirmationTests.swift`, just not on every permutation of rate and duration.

PR #530, opened 2019-12-18
Upgrade footer view to new HTML.

mbrandonw:
Had a little time to kill so thought I would get back into upgrading HTML...


Sources/Views/Footer.swift
```diff
@@ -21,8 +21,8 @@ private func footerInfoColumnsView(user: User?, year: Int) -> Node {
 
 private func linksColumnsView(currentUser: User?) -> Node {
   return [
-    .gridColumn(sizes: [.mobile: 12, .desktop: 6], contentColumnView(currentUser: currentUser)),
-    .gridColumn(sizes: [.mobile: 12, .desktop: 6], moreColumnView)
+    .gridColumn(sizes: [.mobile: 4, .desktop: 2], contentColumnView(currentUser: currentUser)),
+    .gridColumn(sizes: [.mobile: 4, .desktop: 2], moreColumnView)
```
mbrandonw: oof, copy paste bug that was caught by snapshot tests!

PR #531, opened 2019-12-20
Upgrade account to use new HTML

mbrandonw:
Moved the account view to the `Views` module and upgraded it to use the new HTML lib.


Sources/PointFree/Account/Invoices.swift
```diff
@@ -14,6 +14,7 @@ import Prelude
 import Stripe
 import Styleguide
 import Tuple
+import Views
```
mbrandonw: a few things moved to `Views` that we depend on in a few spots. we can find better modules to move that shared stuff to some day.

Sources/PointFree/StripeHtml.swift
```diff
@@ -124,255 +125,3 @@ private let stripeInputClass =
     | Class.flex.flex
     | Class.flex.justify.center
     | Class.size.width100pct
-
-private let countries: [(String, String)] = [
```
mbrandonw: not using this anymore so getting rid of it

Sources/PointFree/TODO.swift
```diff
@@ -16,19 +16,6 @@ import Tagged
 import Tuple
 import UrlFormEncoding
 
-extension Tagged where Tag == EncryptedTag, RawValue == String {
```
mbrandonw: moved to `Views`. seems weird to do that, it could be in its own module. we also need to get the `encrypted` and `decrypted` helpers out of `HttpPipeline`, doesn't make a lot of sense to be in there.

PR #534, opened 2019-12-22
More HTML upgrades

mbrandonw:
Whittling away at more upgrades. This started as converting emails, but then it kind of spiraled out of control.


Tests/PointFreeTests/EmailTests/__Snapshots__/InviteEmailTests/testInviteAcceptance.1.html
```diff
@@ -79,16 +79,17 @@
                            bold
                            ts-d-r2
                            ts-m-r1_5
-                           lh-2"
-                    class="m-pb2"
+                           lh-2
+                           m-pb2"
```
mbrandonw: this is a legit snapshot fix. in this view I was accidentally adding two `class` attributes, so while converting I merged it into a single `class`.

Tests/PointFreeTests/EmailTests/__Snapshots__/TeamEmailsTests/testTeammateRemovedEmailView.1.html
```diff
@@ -106,7 +106,9 @@
                             margin-left:0;
                             padding-top:1rem;
                             padding-bottom:1rem">
-                    You have removed Blob from your Point-Free team, which means they no longer
+                    You have removed 
+                    Blob
+                     from your Point-Free team, which means they no longer
```
mbrandonw: I don't know why the text is breaking this way now, but it doesn't affect the screenshot so I think it's ok.

PR #535, opened 2019-12-22
Even more HTML upgrades

mbrandonw:
Getting very close. Think we just have the RSS/Atom feeds to convert and then we'll be done.


Sources/PointFree/MetaLayout.swift
```diff
@@ -51,55 +51,58 @@ public struct Metadata<A> {
       }
   }
 
-  var metaNodes: [Node] {
-    return [
-
-      self.description.map { meta(name: "description", content: $0) },
```
mbrandonw: oops I guess I didn't finish this. will finish in a moment.

PR #538, opened 2019-12-28
Upgrade syndication module to new HTML

mbrandonw:
This brings us within shooting distance of finishing the HTML upgrade, but there are still a few more small things left. I may need help with the last bit.


stephencelis:
Could prob merge some of the upstream branches now, too!

PR #541, opened 2020-01-09
Swift 5.1

stephencelis:
🎉


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM swift:5.0.2
+FROM swift:5.1
```
mbrandonw: is it possible to use 5.1.3 for this too? I thought a bunch of foundation stuff was fixed on linux in releases after 5.1

Sources/Logger/Logger.swift
```diff
@@ -1,106 +0,0 @@
...
-    _ message: @autoclosure () -> A,
-    file: StaticString = #file,
-    line: UInt = #line) {
-
-    let file = String(String(describing: file).split(separator: "/").last!)
-
-    if level.rawValue >= self.level.rawValue {
-      switch level {
-      case .debug, .info, .warn:
-        print("[\(level)] \(file):\(line): \(message())", to: &self.output)
```
mbrandonw: we're losing the `file` and `line` stuff in the new logger. I think we should get that back cause it's been useful in the past.

Dockerfile
stephencelis: Per this page I believe 5.1 is the same as 5.1.3.

stephencelis: https://hub.docker.com/_/swift

Sources/Logger/Logger.swift
stephencelis: done!

PR #545, opened 2020-01-11
Linux Things

stephencelis:
Just a few quick things around Linux...


Dockerfile
```diff
@@ -1,23 +1,32 @@
...
 RUN make -C cmark install
 
-RUN swift package update
-RUN swift build --product Server --configuration release -Xswiftc -g && \
-    swift build --product Runner --configuration release -Xswiftc -g
-CMD .build/release/Server
+RUN swift build --configuration release --product Server \
+  && swift build --configuration release --product Runner
+
+FROM swift:5.1-slim
```
stephencelis: Trying the slim image...shouldn't hurt...

Dockerfile
```diff
@@ -1,23 +1,32 @@
-FROM swift:5.1
+FROM swift:5.1 as build
 
 RUN apt-get update
-RUN apt-get install -y cmake libpq-dev libssl-dev libz-dev openssl postgresql
```
stephencelis: We don't actually need PostgreSQL running on Linux.

Dockerfile
```diff
@@ -1,23 +1,32 @@
...
-	git checkout 1880e6535e335f143f9547494def01c13f2f331b
+RUN git clone https://github.com/commonmark/cmark \
+  && cd cmark \
+  && git checkout 1880e6535e335f143f9547494def01c13f2f331b
 RUN make -C cmark INSTALL_PREFIX=/usr
 RUN make -C cmark install
 
-RUN swift package update
-RUN swift build --product Server --configuration release -Xswiftc -g && \
-    swift build --product Runner --configuration release -Xswiftc -g
```
stephencelis: Do you know what `-g` was doing here?

Makefile
```diff
@@ -269,28 +269,6 @@ routes:
...
-	@testable import DatabaseTests; \
-	@testable import GitHubTests; \
-	@testable import ModelsTests; \
-	@testable import PointFreeRouterTests; \
-	@testable import PointFreeTests; \
-	@testable import StripeTests; \
-	@testable import StyleguideTests; \
-	@testable import SyndicationTests;
-
-sourcery-tests: check-sourcery
```
stephencelis: I think we can drop Sourcery completely soon...

Package.swift
```diff
@@ -188,7 +188,6 @@ let package = Package(
         "Html",
         "HtmlCssSupport",
         "HtmlPlainTextPrint",
-        "Html",
```
stephencelis: Noticed some warnings for dupe dependencies...

Dockerfile
mbrandonw: Yeah that is in order to include the stack trace symbols so that we get crash logs on linux:

https://github.com/pointfreeco/pointfreeco/pull/459/files#r312045748

Makefile
mbrandonw: 🙏

PR #546, opened 2020-01-12
Route Reflection

stephencelis:
This PR gets rid of code gen for routing isos in favor of some reflection-based code instead.


Sources/Stripe/Model.swift
```diff
@@ -249,7 +249,7 @@ public struct Invoice: Codable, Equatable {
   public var customer: Customer.Id
   public var discount: Discount?
   public var id: Id?
-  public var invoicePdf: String
+  public var invoicePdf: String?
```
stephencelis: While working on this branch I noticed that upcoming invoices weren't decoding because this field can be missing/null.

PR #549, opened 2020-01-26
Ignore new Stripe hook failures

stephencelis:
With the Stripe API came new hooks delivered for subscription creation failures. We can ignore them for now.

### DIFF
--- a/Sources/PointFree/StripeWebhooks.swift
+++ b/Sources/PointFree/StripeWebhooks.swift
@@ -21,9 +21,12 @@ let stripeWebhookMiddleware: Middleware<
   Data
   >
   = validateStripeSignature
+    <<< filter(
+      { $0.data.object.either({ $0.number != nil }, const(true)) },
+      or: writeStatus(.ok) >=> respond(text: "OK")
+    )
     <<< filterMap(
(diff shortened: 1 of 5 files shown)


Sources/Stripe/Model.swift
```diff
@@ -251,7 +251,7 @@ public struct Invoice: Codable, Equatable {
   public var id: Id?
   public var invoicePdf: String?
   public var lines: ListEnvelope<LineItem>
-  public var number: Number
+  public var number: Number?
```
stephencelis: Temporary invoices can be delivered via hook and they do not have a number.

Sources/PointFree/StripeWebhooks.swift
```diff
@@ -21,9 +21,12 @@ let stripeWebhookMiddleware: Middleware<
   Data
   >
   = validateStripeSignature
+    <<< filter(
+      { $0.data.object.either({ $0.number != nil }, const(true)) },
```
mbrandonw: is it possible to write a middleware test for this?

stephencelis: Done.

PR #550, opened 2020-01-28
Coupla build speed-ups

stephencelis:
A couple low-hanging optimizations. Middleware chains are the primary offender for our build times, but I'm not sure if there's any easy way of cleaning those up without untangling a lot of types.

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -17,8 +17,8 @@ RUN git clone https://github.com/commonmark/cmark \
 RUN make -C cmark INSTALL_PREFIX=/usr
 RUN make -C cmark install
 
-RUN swift build --configuration release --product Server -Xswiftc -g \
-  && swift build --configuration release --product Runner -Xswiftc -g
+RUN swift build --configuration release --product Server -Xswiftc -g -Xswiftc -Xfrontend -Xswiftc -warn-long-function-bodies=200 -Xswiftc -Xfrontend -Xswiftc -warn-long-expression-type-checking=200 \
+  && swift build --configuration release --product Runner -Xswiftc -g -Xswiftc -Xfrontend -Xswiftc -warn-long-function-bodies=200 -Xswiftc -Xfrontend -Xswiftc -warn-long-expression-type-checking=200
 
 FROM swift:5.1-slim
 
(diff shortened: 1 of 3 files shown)


Dockerfile.local
```diff
@@ -17,8 +17,8 @@ RUN git clone https://github.com/commonmark/cmark \
 RUN make -C cmark INSTALL_PREFIX=/usr
 RUN make -C cmark install
 
-RUN swift build --configuration release --product Server -Xswiftc -g \
-  && swift build --configuration release --product Runner -Xswiftc -g
+RUN swift build --configuration release --product Server -Xswiftc -g -Xswiftc -Xfrontend -Xswiftc -warn-long-function-bodies=200 -Xswiftc -Xfrontend -Xswiftc -warn-long-expression-type-checking=200 \
+  && swift build --configuration release --product Runner -Xswiftc -g -Xswiftc -Xfrontend -Xswiftc -warn-long-function-bodies=200 -Xswiftc -Xfrontend -Xswiftc -warn-long-expression-type-checking=200
```
stephencelis: Nice way of checking this stuff in Swift. Is there a way of getting build settings like these into SPM projects without an Xcode project file?

Sources/Models/BlogPosts/BlogPost0011_SolutionsToZipExercisesPt1.swift
```diff
@@ -418,12 +425,7 @@ introductory series to `zip`! If you thought those were too easy, be sure to che
 [part 2](/episodes/ep24-the-many-faces-of-zip-part-2) and
 [part 3](/episodes/ep25-the-many-faces-of-zip-part-3) too. Until next time!
 """,
-      timestamp: nil,
-      type: .paragraph
-    ),
-  ],
-  coverImage: "https://d1iqsrac68iyd8.cloudfront.net/posts/0011-solutions-to-zip-pt1/poster.jpg",
-  id: 11,
-  publishedAt: Date(timeIntervalSince1970: 1_532_930_223 + 604_800*2 + 60*60*24),
```
stephencelis: This was taking 12+ seconds to compile on Linux.

Sources/Stripe/Model.swift
```diff
@@ -100,7 +100,10 @@ public struct Coupon: Equatable {
     case let .amountOff(amountOff):
       return cents - amountOff
     case let .percentOff(percentOff):
-      return cents.map { Int((Double($0) * (1 - (Double(percentOff) / 100))).rounded()) }
```
stephencelis: Also taking 12+ seconds.

Dockerfile.local
mbrandonw: Huh not sure!

PR #551, opened 2020-01-29
Allow trialing subscriptions to have access to episodes.

mbrandonw:


Sources/Models/SubscriberState.swift
```diff
@@ -55,17 +56,25 @@ public enum SubscriberState {
   public var isNonSubscriber: Bool {
     switch self {
     case .teammate(status: .active, _),
-         .owner(hasSeat: _, status: .active, _):
+         .teammate(status: .trialing, _),
+         .owner(hasSeat: _, status: .active, _),
+         .owner(hasSeat: _, status: .trialing, _):
       return false
```
stephencelis: Could also stop matching against `status` here and below and `return self.isActive`, but can do that later.

Tests/PointFreeTests/EpisodePageTests.swift
```diff
@@ -436,4 +436,18 @@ class EpisodePageTests: TestCase {
...
+
+    var subscription = Subscription.mock
+    subscription.stripeSubscriptionStatus = .trialing
+    Current.database.fetchSubscriptionById = { _ in pure(subscription) }
+
+    let episode = request(to: .episode(.left(Current.episodes().first!.slug)), session: .loggedIn(as: .mock))
+
+    let conn = connection(from: episode)
+
+    assertSnapshot(matching: conn |> siteMiddleware, as: .ioConn)
```
stephencelis: Fun thing to experiment later with is to test active/trialing both in the same test with the same snapshot name.

PR #554, opened 2020-02-12
GitHub Actions + Swift 5.2 Prep

stephencelis:
Needed a non-creative thing to work on at the end of the day so did a few infra-related things. This PR:

- Moves to GitHub Actions
- Upgrades the Package.swift file in anticipation of Swift 5.2
- Cleans up a few things I found along the way (like a `.bin` directory I found with a _ton_ of Sourcery stuff)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+on: [push]
+
+jobs:
+  build:
+    name: MacOS
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Homebrew Update
+      run: brew update
```
stephencelis: This adds a minute to builds (and was something we avoided on Circle), but without it we got the same failure as Circle 🤷‍♂

Dockerfile.local
```diff
@@ -1,12 +1,10 @@
-FROM swift:5.1 as build
+FROM swift:5.1
```
stephencelis: CI builds shouldn't be slim (or build release) so that they can run tests.

Dockerfile.local
```diff
@@ -17,18 +15,6 @@ RUN git clone https://github.com/commonmark/cmark \
 RUN make -C cmark INSTALL_PREFIX=/usr
 RUN make -C cmark install
 
-RUN swift build --configuration release --product Server -Xswiftc -g -Xswiftc -Xfrontend -Xswiftc -warn-long-function-bodies=200 -Xswiftc -Xfrontend -Xswiftc -warn-long-expression-type-checking=200 \
-  && swift build --configuration release --product Runner -Xswiftc -g -Xswiftc -Xfrontend -Xswiftc -warn-long-function-bodies=200 -Xswiftc -Xfrontend -Xswiftc -warn-long-expression-type-checking=200
+RUN swift build --build-tests --jobs 1 -Xswiftc -D -Xswiftc OSS
```
stephencelis: We should investigate this or submit a bug, but switching to `--jobs 1` is what prevented the weird Docker exits I was seeing.

Dockerfile.local
```diff
@@ -17,18 +15,6 @@ RUN git clone https://github.com/commonmark/cmark \
...
-RUN apt-get install -y libpq-dev libssl-dev libz-dev
-
-WORKDIR /app
-
-COPY --from=build /usr/include/cmark* /usr/include/
-COPY --from=build /usr/lib/libcmark* /usr/lib/
-COPY --from=build /build/.build/release/Server /usr/bin
-COPY --from=build /build/.build/release/Runner /usr/bin
-ENTRYPOINT ["Server"]
+ENTRYPOINT [".build/debug/Server"]
```
stephencelis: This still lets us run things as a server locally by default.

Makefile
```diff
@@ -101,10 +101,9 @@ db-drop:
 	dropdb --username pointfreeco pointfreeco_test || true
 	dropuser pointfreeco || true
 
-
 xcodeproj-oss: check-dependencies
```
stephencelis: We can prob delete the Xcode project and module map commands entirely.

Package.swift
```diff
@@ -11,7 +11,6 @@ let package = Package(
     .executable(name: "Runner", targets: ["Runner"]),
     .executable(name: "Server", targets: ["Server"]),
     .library(name: "Database", targets: ["Database"]),
-    .library(name: "DatabaseTestSupport", targets: ["DatabaseTestSupport"]),
```
stephencelis: I accidentally deleted this line, but it's making me think that unless other packages depend on this one, maybe we don't need to declare anything here but public products?

Package.swift
```diff
@@ -44,80 +43,68 @@ let package = Package(
...
       dependencies: [
-        "Either",
         "GitHub",
-        "Logging",
         "Models",
-        "PostgreSQL",
-        "Prelude",
         "Stripe",
-        "Tagged",
+        .product(name: "Either", package: "swift-prelude"),
```
mbrandonw: Whoa, what’s this fanciness??

stephencelis: Swift 5.2 will require describing external packages now. I think it improves resolution, and it's nice as documentation.

Sources/PointFreeTestSupport/TestCase.swift
```diff
@@ -45,7 +45,7 @@ open class TestCase: XCTestCase {
   }
 
   public var isScreenshotTestingAvailable: Bool {
-    if ProcessInfo.processInfo.environment["CIRCLECI"] == nil {
+    if ProcessInfo.processInfo.environment["CI"] == nil {
```
stephencelis: This gets set by GitHub actions now.

Tests/DatabaseTests/DatabaseTestCase.swift
```diff
@@ -1,41 +0,0 @@
-import Database
-import GitHub
-import GitHubTestSupport
-import Logging
-import ModelsTestSupport
-import Prelude
-import SnapshotTesting
-import XCTest
-
-class DatabaseTestCase: XCTestCase {
```
stephencelis: I moved this back to `PointFreeTests`. We need to rely on `Current` for the env vars.

Tests/PointFreeTests/DatabaseTests.swift
```diff
@@ -1,26 +1,30 @@
...
 import ModelsTestSupport
 import GitHubTestSupport
 import Logging
 import SnapshotTesting
 import XCTest
+@testable import PointFree
 
-final class DatabaseTests: DatabaseTestCase {
+
+final class DatabaseTests: TestCase {
```
mbrandonw: Is there no more DatabaseTestCase?

Tests/PointFreeTests/XCTestManifests.swift
```diff
@@ -148,6 +148,18 @@ extension ChangeTests {
     ]
 }
 
+extension DatabaseTests {
+    // DO NOT MODIFY: This is autogenerated, use:
+    //   `swift test --generate-linuxmain`
```
stephencelis: I think we can get rid of this with auto-discovery, but can save for another PR later...

.github/workflows/ci.yml
mbrandonw: huh dang. wonder if we could speed it up by saving some cache artifacts or something. something to look into later

Makefile
mbrandonw: yeah, even for playgrounds we found a way to get the packages and playgrounds to play nicely together. so down to delete whenever we want

Package.swift
mbrandonw: nice!

Tests/DatabaseTests/DatabaseTestCase.swift
mbrandonw: why was this necessary? are env vars handled differently for GH actions?

stephencelis: We weren't previously running CI under docker, so we were relying on a hard-coded localhost-based PostgreSQL server running.

Tests/PointFreeTests/DatabaseTests.swift
stephencelis: Not needed since moving back to PointFreeTests...

stephencelis:
> there are more things to do before we are fully on GH actions?

I don't think so. This is it!

PR #558, opened 2020-02-13
Add slow type checking warnings to SPM package

stephencelis:
Just wanted to wrap up a quick exploration from this morning to restore warnings when editing Point-Free as a Swift package and not an Xcode project.

image

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,6 +2,16 @@
 
 import PackageDescription
 
+extension SwiftSetting {
+  static let warnLongExpressionTypeChecking = unsafeFlags(
+    [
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -345,13 +377,27 @@ let package = Package(
...
+      swiftSettings: [.warnLongExpressionTypeChecking]
+    ),
 
   ]
 )
+
+extension SwiftSetting {
+  static let warnLongExpressionTypeChecking = unsafeFlags(
+    [
+      "-Xfrontend", "-warn-long-expression-type-checking=200",
```
stephencelis: We can do something higher than `200` if this is too noisy.

PR #559, opened 2020-02-15
Build Time Speed-Ups

stephencelis:
I'll be tacking on some more stuff on this branch, but curious about CI speedups along the way.


Package.swift
```diff
@@ -20,22 +20,6 @@ let package = Package(
   products: [
     .executable(name: "Runner", targets: ["Runner"]),
     .executable(name: "Server", targets: ["Server"]),
-    .library(name: "Database", targets: ["Database"]),
```
stephencelis: This is only if other libraries wanna pull PointFree in and import frameworks...for now simpler to not maintain this list.

Sources/PointFree/About.swift
```diff
@@ -10,7 +10,7 @@ import Views
 let aboutResponse: Middleware<StatusLineOpen, ResponseEnded, Tuple3<User?, SubscriberState, Route?>, Data> =
   writeStatus(.ok)
     >=> map(lower)
-    >>> _respond(
+    >>> respond(
```
stephencelis: You mentioned the underscored version can go away...

Sources/PointFree/Account/Change.swift
```diff
@@ -31,10 +22,22 @@ let subscriptionChangeMiddleware: Middleware<
       )
 )
 
+private let requireUserAndPricingAndSeats
+  : MT<Tuple2<User?, Pricing?>, Tuple3<User, Int, Pricing>>
```
stephencelis: Using an `MT` type as a shorthand for describing middleware transformers...makes it slightly easier to break this stuff up.

Sources/PointFree/Account/Account.swift
```diff
@@ -9,11 +9,11 @@ import Stripe
 import Tuple
 import Views
 
-let accountResponse: Middleware<StatusLineOpen, ResponseEnded, Tuple2<User?, SubscriberState>, Data> =
-  filterMap(require1 >>> pure, or: loginAndRedirect)
+let accountResponse//: M<Tuple2<User?, SubscriberState>>
```
mbrandonw: did you mean to drop the type and have this comment here?

Sources/PointFree/Account/Change.swift
mbrandonw: I've actually got typealiases for this I've been using in other parts:

https://github.com/pointfreeco/pointfreeco/blob/master/Sources/PointFree/TODO.swift#L20-L21

Sources/PointFree/Admin/EpisodeCredit.swift
```diff
@@ -21,15 +21,10 @@ let showEpisodeCreditsMiddleware: Middleware<
...
-let redeemEpisodeCreditMiddleware: Middleware<
-  StatusLineOpen,
-  ResponseEnded,
-  Tuple3<User, User.Id?, Int?>,
-  Data
-  > =
-  filterMap(
-      over2(fetchUser(id:)) >>> sequence2 >>> map(require2),
-      or: redirect(to: .admin(.episodeCredits(.show)), headersMiddleware: flash(.error, "Could not find that user."))
+let redeemEpisodeCreditMiddleware
```
mbrandonw: is it not worth keeping the type here? does it not help at all?

Sources/PointFree/Admin/NewEpisodes.swift
```diff
@@ -13,7 +13,7 @@ import Prelude
 import Styleguide
 import Tuple
 
-let showNewEpisodeEmailMiddleware: AppMiddleware<Prelude.Unit> =
+let showNewEpisodeEmailMiddleware: M<Prelude.Unit> =
```
mbrandonw: do we want to use `M` instead of `AppMiddleware`?

Package.swift
mbrandonw: actually is this necessary to get targets in Xcode too? when I dont have these decelerations im not able to build these targets in isolation anymore

stephencelis: They all show up to me like this:

image

I'll address comments/merge if nothing's blocking, but we can revert this part if it's behaving weirdly.

Sources/PointFree/Account/Account.swift
stephencelis: Most of the time, typing these middleware led to _slower_ builds (by 2x). I user comments to extract/simplify some of these. I'll delete this comment...

Sources/PointFree/Admin/EpisodeCredit.swift
stephencelis: Makes it worse somehow :/

Sources/PointFree/Admin/NewEpisodes.swift
stephencelis: I didn't notice the overlap till I moved `M` and `MT` into TODO.swift. I figure this is all a temp situation till we rethink HttpPipeline, so cool with any names. Preference before merging?

mbrandonw: cool, let's keep as is then!

PR #560, opened 2020-02-15
Add LiveDatabaseTestCase

stephencelis:
This audits our tests for ones that actually need the live database and avoids the IO of creating/migrating/dropping for a bunch of test cases.


stephencelis:
Dunno why but Ubuntu CI started failing out of the blue :/ Can't repro locally

Tests/PointFreeTests/AccountTests/UpdateProfileTests.swift
```diff
@@ -96,6 +96,13 @@ class UpdateProfileTests: TestCase {
     assertSnapshot(matching: output, as: .conn)
     #endif
   }
+}
+
+class UpdateProfileTests: LiveDatabaseTestCase {
```
mbrandonw: should this inherit from `TestCase` instead?

Tests/PointFreeTests/AuthTests.swift
```diff
@@ -10,7 +10,7 @@ import Optics
 import SnapshotTesting
 import XCTest
 
-class AuthTests: TestCase {
+class AuthIntegrationTests: LiveDatabaseTestCase {
```
mbrandonw: small potential gotcha for having two test classes in a single file is that snapshots may overwrite each other if the test function has the same name. only the #file and #function is baked into the snapshot file name.

Tests/PointFreeTests/AccountTests/UpdateProfileTests.swift
stephencelis: Ha yeah.

Tests/PointFreeTests/AuthTests.swift
stephencelis: Yeah...though I think it's OK since collisions will get caught in test runs?

stephencelis:
Other than the update profile mismatch I believe everything's audited!

PR #562, opened 2020-02-16
Break out 3 simple dependencies

mbrandonw:
Right now there a few very simple things inside `PointFreePrelude` that could be modules on their own, and everyone that wants access to those things is forced to compile a lot more than is necessary. This breaks a few of those things as a small step to unraveling some unnecessary dependencies.

In particular, it adds a module to hold the `EmailAddress` and `DecodableRequest` types (both of which could prob just be in swift-tagged), as well as some simple foundation helpers for building url requests and perform URL data tasks.


.github/workflows/ci.yml
```diff
@@ -25,6 +25,6 @@ jobs:
     name: Ubuntu
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v1
```
mbrandonw: saw you do this in your other PR...

Package.swift
```diff
@@ -21,6 +21,9 @@ let package = Package(
     .executable(name: "Runner", targets: ["Runner"]),
     .executable(name: "Server", targets: ["Server"]),
     .library(name: "Database", targets: ["Database"]),
+    .library(name: "DecodableRequest", targets: ["DecodableRequest"]),
+    .library(name: "EmailAddress", targets: ["EmailAddress"]),
+    .library(name: "FoundationPrelude", targets: ["FoundationPrelude"]),
```
mbrandonw: I'm adding these libraries because it's what seems to give us targets that we can build in isolation in Xcode...

Package.swift
```diff
@@ -80,6 +84,32 @@ let package = Package(
...
+      ],
+      swiftSettings: [.warnLongExpressionTypeChecking]
+    ),
+
+    .target(
+      name: "FoundationPrelude",
+      dependencies: [
+        .product(name: "Either", package: "swift-prelude"),
+        .product(name: "Logging", package: "swift-log"),
+        .product(name: "UrlFormEncoding", package: "swift-web"),
```
mbrandonw: ideally we could ween this package off of `Logging` and `UrlFormEncoding` too

Package.swift
```diff
@@ -104,11 +134,11 @@ let package = Package(
...
       name: "GitHub",
       dependencies: [
-        "PointFreePrelude",
+        "DecodableRequest",
+        "EmailAddress",
         .product(name: "Either", package: "swift-prelude"),
+        "FoundationPrelude",
         .product(name: "Logging", package: "swift-log"),
-        .product(name: "Optics", package: "swift-prelude"),
-        .product(name: "Prelude", package: "swift-prelude"),
```
mbrandonw: by using `DecodableRequest`, `EmailAddress` and `FoundationPrelude` we can stop depending on heavier modules like prelude and optics.

Sources/FoundationPrelude/URLRequest.swift
```diff
@@ -0,0 +1,123 @@
+import Either
+import Foundation
+import Logging
+import UrlFormEncoding
+
+extension URLRequest {
+  public mutating func guaranteeHeaders() {
```
mbrandonw: switched some optics stuff to just be mutating methods on `URLRequest`

Sources/FoundationPrelude/URLRequest.swift
```diff
@@ -0,0 +1,123 @@
...
+
+  public mutating func attach(formData: [String: Any]) {
+    self.httpBody = Data(urlFormEncode(value: formData).utf8)
+  }
+
+}
+
+public func dataTask(
+  with request: URLRequest,
+  logger: Logger?
```
mbrandonw: instead of passing in logger we could just provide a function to do the logging

Sources/GitHub/Client.swift
```diff
@@ -36,9 +35,11 @@ extension Client {
 
   public init(clientId: Id, clientSecret: Secret, logger: Logger?) {
     self.init(
-      fetchAuthToken: fetchGitHubAuthToken(clientId: clientId, clientSecret: clientSecret) >>> runGitHub(logger),
-      fetchEmails: fetchGitHubEmails >>> runGitHub(logger),
-      fetchUser: fetchGitHubUser >>> runGitHub(logger)
```
mbrandonw: removing dependence on `>>>`, we could prob uncurry `runGitHub` and use trailing closures to make this a lil nicer

Package.swift
```diff
@@ -137,10 +167,12 @@ let package = Package(
     .target(
       name: "Mailgun",
       dependencies: [
+        "DecodableRequest",
+        "EmailAddress",
         .product(name: "HttpPipeline", package: "swift-web"),
         "Models",
-        "PointFreePrelude",
         .product(name: "Either", package: "swift-prelude"),
+        "FoundationPrelude",
```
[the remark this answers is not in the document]
mbrandonw: Ah down for that. Meant to ask

Package.swift
mbrandonw: I only see this:

image

maybe it's fixed in a beta, but not in 11.3

stephencelis:
Gonna merge this in case I have a lil time today to knock away at cleanup work

PR #563, opened 2020-02-16
Speed up a bunch of test compilation times.

mbrandonw:
(NB: This is based off of #559)

Just like middlewares, some tests take awhile to build do to the way we configure the environment. This flattens things to just use plain mutation. Some things are messier this way. It would be worth seeing if we can regain the ergonomics by using overture while still keeping compile times down.

### DIFF
--- a/Sources/PointFree/Session.swift
+++ b/Sources/PointFree/Session.swift
@@ -18,7 +18,7 @@ private let cookieExpirationDuration: TimeInterval = 315_360_000 // 60 * 60 * 24
 public func writeSessionCookieMiddleware<A>(_ update: @escaping (Session) -> Session)
   -> (Conn<HeadersOpen, A>)
   -> IO<Conn<HeadersOpen, A>> {
-    
+
     return { conn in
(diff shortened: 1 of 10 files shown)


Sources/PointFree/Session.swift
```diff
@@ -13,23 +13,24 @@ public enum CookieTransform: String, Codable {
...
+
 public func writeSessionCookieMiddleware<A>(_ update: @escaping (Session) -> Session)
   -> (Conn<HeadersOpen, A>)
   -> IO<Conn<HeadersOpen, A>> {
-
+    
     return { conn in
       let value = update(conn.request.session)
       guard value != conn.request.session else { return pure(conn) }
-      return setCookie(
```
mbrandonw: this was taking about 2.5 seconds to compile, and at first I thought it was due to the `map` and `??` stuff, but most likely it was due to the arithmetic being done.

stephencelis: Ha jeez, yeah.

PR #564, opened 2020-02-17
Route and JS events for video progress.

mbrandonw:
We'll be gearing up for collections soon, and so wanted to start getting video progress in place. This sets up the route necessary to receiving progress from the client, and uses Vimeo's JS API to ping new route every time the track changes by a percent. This doesn't have any of the database stuff, that will come next.


Sources/PointFree/Account/PrivateRss.swift
```diff
@@ -266,7 +266,7 @@ private func item(forUser user: User, episode: Episode) -> RssItem {
       type: "video/mp4",
       url: episode.fullVideo.downloadUrl
     ),
-    guid: url(to: .episode(.left(episode.slug))),
+    guid: url(to: .episode(.show(.left(episode.slug)))),
```
mbrandonw: I bundled up the show and index episode routes into a route enum of its own, so that will cause a bit of churn where we link to episodes.

mbrandonw: I also think we could prob get rid of the `Either<String, Episode.Id>` and just do a `String`. I do think think we ever take advantage of the fact that you can link to an episode with just the id...

Sources/PointFree/Episode/Show.swift
```diff
@@ -59,6 +59,22 @@ let useCreditResponse =
     <<< validateCreditRequest
     <| applyCreditMiddleware
 
+let progressResponse: AppMiddleware<
```
mbrandonw: this is just a stub of a middleware that will do the work for progress.

Sources/PointFreeRouter/Routes.swift
```diff
@@ -64,6 +63,12 @@ public enum Route: Equatable {
     case requestInvite(EnterpriseAccount.Domain, EnterpriseRequestFormData)
   }
 
+  public enum EpisodeRoute: Equatable {
+    case index
+    case progress(param: Either<String, Episode.Id>, percent: Int)
+    case show(Either<String, Episode.Id>)
```
mbrandonw: the new route enum

Sources/Views/EpisodeVideoView.swift
```diff
@@ -44,6 +44,21 @@ window.addEventListener("load", function (event) {
     player.setCurrentTime(time);
     if (play) { player.play(); }
   }
+
+  var lastSeenPercent = 0
+  player.on('timeupdate', function(data) {
```
mbrandonw: this notifies us every ~250ms of the current video play times. I use this to determine when the video progress percentage increases by 1 or more, and if it does I fire off an API request

Sources/Views/EpisodeVideoView.swift
```diff
@@ -44,6 +44,21 @@ window.addEventListener("load", function (event) {
...
+    console.log(data.percent - lastSeenPercent)
+    if (data.percent - lastSeenPercent >= 0.01) {
+      lastSeenPercent = data.percent;
+
+      var httpRequest = new XMLHttpRequest();
+      httpRequest.open(
+        "POST",
+        window.location.pathname + "/progress?percent=" + Math.round(data.percent * 100)
```
mbrandonw: I didn't want to have to think about somehow reconstructing the URL the route, so I think this works. let me know if you have a better idea.

Sources/PointFreeRouter/Routes.swift
stephencelis: Could also change `EpisodeRoute` to `EpisodesRoute` and that way we could start grouping more specific routing for particular episodes:

```swift
case episode(Either<String, Episode.Id>, EpisodeRoute)
…
.episode(.left(slug), .show)
.episode(.left(slug), .progress(0.5)
```

Sources/Views/EpisodeVideoView.swift
stephencelis: Yeah, templatizing seems more complicated...

Sources/PointFreeRouter/Routes.swift
```diff
@@ -151,12 +156,17 @@ let routers: [Router<Route>] = [
...
     <¢> get %> "blog" %> "posts" %> pathParam(.blogPostIdOrString) <% end,
 
-  .case(Route.episode)
-    <¢> get %> "episodes" %> pathParam(.episodeIdOrString) <% end,
-
-  .case(.episodes)
+  .case(.episode(.index))
     <¢> get %> "episodes" <% end,
 
+  parenthesize(.case { Route.episode(Route.EpisodeRoute.progress(param: $0, percent: $1)) })
```
stephencelis: Worth simplifying to?
```suggestion
  parenthesize(.case { .episode(.progress(param: $0, percent: $1)) })
```
Or does type checking start to suffer?

mbrandonw: I think I only had that while trying to get things to compile.

Sources/PointFreeRouter/Routes.swift
mbrandonw: def down to do that, but how do you envision support `index` too? just a `case episodes` next to `case episode`?

mbrandonw:
gonna merge now, but let's chat about https://github.com/pointfreeco/pointfreeco/pull/564#pullrequestreview-359476058 tomorrow!

PR #566, opened 2020-02-18
DB endpoint for video progress

mbrandonw:
This adds the DB support necessary for the video progress feature.


Package.swift
```diff
@@ -386,14 +386,6 @@ let package = Package(
       swiftSettings: [.warnLongExpressionTypeChecking]
     ),
 
-    .testTarget(
-      name: "SyndicationTests",
```
mbrandonw: there weren't any tests in this target, so just deleting for now

Sources/Database/Database.swift
```diff
@@ -38,10 +38,11 @@ public struct Client {
   public var incrementEpisodeCredits: ([Models.User.Id]) -> EitherIO<Error, [Models.User]>
   public var insertTeamInvite: (EmailAddress, Models.User.Id) -> EitherIO<Error, TeamInvite>
   public var migrate: () -> EitherIO<Error, Prelude.Unit>
-  public var redeemEpisodeCredit: (Int, Models.User.Id) -> EitherIO<Error, Prelude.Unit>
+  public var redeemEpisodeCredit: (Episode.Sequence, Models.User.Id) -> EitherIO<Error, Prelude.Unit>
```
mbrandonw: I upgraded the `sequence: Int` field on `Episode` to a tagged type to keep things clear, and it helped clean up some things too.

Sources/Database/Database.swift
```diff
@@ -391,6 +395,23 @@ private struct _Client {
...
+    episodeSequence: Episode.Sequence,
+    percent: Int,
+    userId: Models.User.Id
+  ) -> EitherIO<Error, Prelude.Unit> {
+    return self.execute(
+      #"""
+      INSERT INTO "episode_progresses" ("episode_sequence", "percent", "user_id")
+      VALUES ($1, $2, $3)
+      ON CONFLICT ("episode_sequence", "user_id") DO UPDATE
+      SET "percent" = GREATEST(episode_progresses.percent, $2)
```
mbrandonw: this makes sure that if a lower percentage comes in we won't count it towards their progress.

Sources/Database/Database.swift
```diff
@@ -1031,6 +1052,24 @@ private struct _Client {
       ADD FOREIGN KEY ("subscription_id") REFERENCES "subscriptions" ("id")
       """
       )))
+      .flatMap(const(execute(
+        #"""
+        CREATE TABLE IF NOT EXISTS "episode_progresses" (
+          "id" uuid DEFAULT uuid_generate_v1mc() PRIMARY KEY NOT NULL,
+          "episode_sequence" smallint NOT NULL,
+          "percent" smallint NOT NULL,
```
mbrandonw: cool to use `smallint` for these fields?

Tests/PointFreeTests/DatabaseTests.swift
```diff
@@ -81,4 +80,69 @@ final class DatabaseTests: LiveDatabaseTestCase {
 
     XCTAssertEqual(subscription.id, freshUser.subscriptionId)
   }
+
+  func testUpdateEpisodeProgress() {
```
mbrandonw: this test runs against the live DB to make sure that the query logic around taking the max of percentages works correctly.

Package.swift
```diff
@@ -381,19 +381,12 @@ let package = Package(
     .target(
       name: "Syndication",
       dependencies: [
+        "Models",
```
mbrandonw: this now depends on `Models` to get access to `Episode.Sequence`, but really a bunch of stuff in `Syndication` should prob be split into its own module (prob in swift-html), and then we could have a light layer for transforming episode specific models into to RSS content.

mbrandonw:
Oh I should mention, this is _only_ the backend support for the feature. We are not yet invoking the DB endpoint when the client sends info to the backend.

PR #567, opened 2020-02-18
Invoke DB from backend endpoint.

mbrandonw:
This ties together #564 and #566 by invoking the DB endpoint when the backend endpoint is called.

### DIFF
--- a/Sources/PointFree/Episode/Show.swift
+++ b/Sources/PointFree/Episode/Show.swift
@@ -71,20 +71,38 @@ private let validateUserEpisodePermission
     <<< validateCreditRequest
 
 let progressResponse: M<
-  Tuple5<
+  Tuple4<
   Either<String, Episode.Id>,
   Int,
   Models.User?,
-  SubscriberState,
-  Route?>
+  SubscriberState>
   > =
   filterMap(
     over1(episode(forParam:)) >>> require1 >>> pure,
     or: writeStatus(.notFound) >=> end
     )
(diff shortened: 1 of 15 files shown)


Sources/PointFree/Episode/Show.swift
```diff
@@ -71,20 +71,41 @@ private let validateUserEpisodePermission
     <<< validateCreditRequest
 
 let progressResponse: M<
-  Tuple5<
+  Tuple4<
   Either<String, Episode.Id>,
-  Int,
   Models.User?,
   SubscriberState,
-  Route?>
```
mbrandonw: route wasn't needed so got rid of it, and moved the `percent: Int` parameter to the end because I was able to make use of the `userEpisodePermission` middleware by having the first 3 arguments be episode, user and subscriber state.

Sources/PointFree/Episode/Show.swift
```diff
@@ -71,20 +71,41 @@ private let validateUserEpisodePermission
...
+  Int>
   > =
   filterMap(
     over1(episode(forParam:)) >>> require1 >>> pure,
     or: writeStatus(.notFound) >=> end
     )
-    <<< filterMap(require3 >>> pure, or: loginAndRedirect)
-    <| writeStatus(.ok)
-    >=> end
+    <| userEpisodePermission
```
mbrandonw: pretty cool that I could reuse this middleware.

PR #569, opened 2020-02-18
Add Migrations for Referral Code

stephencelis:


Sources/Database/Database.swift
```diff
@@ -1070,6 +1070,45 @@ private struct _Client {
         ON "episode_progresses" ("episode_sequence", "user_id")
         """
       )))
+      .flatMap(const(execute(
+        """
+      CREATE SEQUENCE IF NOT EXISTS users_referral_code_sequence
```
stephencelis: This is to create a small auto-incrementing sequence in the rare case that two people create their accounts at the exact same second...

Sources/Database/Database.swift
```diff
@@ -1070,6 +1070,45 @@ private struct _Client {
...
+      """
+      )))
+      .flatMap(const(execute(
+        """
+      CREATE OR REPLACE FUNCTION next_referral_code (OUT result bigint)
```
stephencelis: This function uses the "snowflake" approach to unique id generation for shorter identifiers. We can take the resulting integer and render it in, e.g., base-36: `"c1vvvvjox"`

stephencelis: (It's probably overkill, lol, but I didn't have any other approaches that came to mind.)

PR #570, opened 2020-02-18
Referral routing

stephencelis:
This updates the user model and adds routing for referrals. This does _not_ introduce any logic around referrals yet.


Sources/Database/Database.swift
```diff
@@ -363,15 +363,7 @@ private struct _Client {
...
-    SELECT "users"."email",
-           "users"."episode_credit_count",
-           "users"."github_user_id",
-           "users"."github_access_token",
-           "users"."id",
-           "users"."is_admin",
-           "users"."name",
-           "users"."subscription_id",
-           "users"."rss_salt"
+    SELECT "users".*
```
stephencelis: Database decoding errors are super opaque and we get them whenever we add new fields to the user model so I think `*` is prob the way to go for these queries...

Sources/Models/SubscribeConfirmationData.swift
```diff
@@ -4,15 +4,18 @@ import Stripe
 public struct SubscribeConfirmationData: Equatable {
   public var billing: Pricing.Billing
   public var isOwnerTakingSeat: Bool
+  public var referralCode: User.ReferralCode?
```
stephencelis: Referral code now comes along for the ride here...

Sources/PointFreeRouter/Routes.swift
```diff
@@ -308,20 +310,23 @@ private let subscriberDataIso = PartialIso<String, SubscribeData?>(
...
     let isOwnerTakingSeat = keyValues
-      .first { key, value in key == SubscribeData.CodingKeys.isOwnerTakingSeat.rawValue }?.1
+      .first { k, _ in k == SubscribeData.CodingKeys.isOwnerTakingSeat.rawValue }?.1
       .flatMap(Bool.init)
       ?? false
 
     let rawCouponValue = keyValues.first(where: { key, value in key == "coupon" })?.1
     let coupon = rawCouponValue == "" ? nil : rawCouponValue.flatMap(Coupon.Id.init(rawValue:))
+    let referralCode = keyValues
+      .first(where: { key, value in key == SubscribeData.CodingKeys.referralCode.rawValue })?.1
```
mbrandonw: I guess this value is also not needed right?

```suggestion
      .first(where: { key, _ in key == SubscribeData.CodingKeys.referralCode.rawValue })?.1
```

Tests/PointFreeTests/SubscribeConfirmationTests.swift
```diff
@@ -355,7 +364,8 @@ class SubscriptionConfirmationTests: TestCase {
           lane: .team,
           billing: nil,
           isOwnerTakingSeat: nil,
-          teammates: nil
+          teammates: nil,
+          referralCode: nil
```
mbrandonw: is it possible to make a test for when the referral code is present?

stephencelis: I'm gonna add tests when the logic is actually in place. Not sure what is possible to test at this point...

mbrandonw: it's testing the gnarly logic in the `subscriberDataIso` for the route: 

https://github.com/pointfreeco/pointfreeco/pull/570/files#diff-549405fc9f778f602011a17f333a61c5R308

that stuff is still just a hacky workaround to our lack of properly decoding that info from the URL.

PR #573, opened 2020-02-20
Basic Referrals UI

stephencelis:
Some basic UI, more tests, and another edge case.


Sources/PointFree/Subscribe.swift
```diff
@@ -215,6 +215,7 @@ private func validateReferrer(
     }
 
     let isSubscribeDataValidForReferral = subscribeData.pricing.lane == .personal
+      && user.referrerId == nil
```
stephencelis: Missed this edge case.

Sources/PointFree/SubscribeConfirmation.swift
```diff
@@ -78,6 +78,19 @@ private func validateReferralCode(
       )
     }
 
+    guard currentUser?.referrerId == nil else {
```
stephencelis: And this one.

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -767,17 +767,25 @@ function updateSeats() {
...
+        + (monthly ? "" : " times <strong>12 months</strong>")
+        + "."
+    )
+  } else {
+    document.getElementById("pricing-preview").innerHTML = (
+      monthly
+        ? "You and your referrer will receive an <strong>$18 credit</strong> when you subscribe. It will apply to future invoices."
+        : "You and your referrer will <strong>save $18</strong> when you subscribe. You will be charged $150 today and $168 on renewal."
+    )
+  }
```
stephencelis: More JS in order to dynamically display different text :/ I'd like to clean some of this up but maybe we decide on how we want this to look first.

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -767,17 +767,25 @@ function updateSeats() {
...
+        + " per month</strong>"
+        + (seats > 1 ? " times <strong>" + seats + " seats</strong>" : "")
+        + (monthly ? "" : " times <strong>12 months</strong>")
+        + "."
+    )
+  } else {
+    document.getElementById("pricing-preview").innerHTML = (
+      monthly
+        ? "You and your referrer will receive an <strong>$18 credit</strong> when you subscribe. It will apply to future invoices."
+        : "You and your referrer will <strong>save $18</strong> when you subscribe. You will be charged $150 today and $168 on renewal."
```
mbrandonw: is it possible to do this without hard coding the values?

PR #574, opened 2020-02-20
Referral Account UI

stephencelis:
This renders a referral link for users on their account page.


Sources/Views/Account/Account.swift
```diff
@@ -263,12 +264,13 @@ private func privateRssFeed(
   let rssLink: Node = rssUrl
     .map { rssUrl in
       [
-        .p(
-          .a(
-            attributes: [.class([Class.pf.type.underlineLink]), .href(rssUrl)],
-            .text(String(rssUrl.prefix(40)) + "...")
+        .ul(
+          .li(
```
stephencelis: Decided to render both RSS and referral links in a list item bullet to make them stand out.

PR #575, opened 2020-02-20
Team invite: DB and route

mbrandonw:
Adds a column to `users` for the team invite code, and adds some routes for the landing page and the action to actually join a team.


Sources/PointFree/Account/Team.swift
```diff
@@ -9,24 +9,36 @@ import Prelude
...
+let joinTeamLandingMiddleware: M<Tuple3<User?, SubscriberState, User.TeamInviteCode>>
+  = requireOwner
+    <| writeStatus(.ok)
+    >=> end
+
+let joinTeamMiddleware: M<Tuple3<User?, SubscriberState, User.TeamInviteCode>>
+  = writeStatus(.ok)
+    >=> end
+
+private func requireOwner<Z>(
```
mbrandonw: I needed to turn this into a function so that it could operate on `T3`s which allow me to reuse for the new team middlewares.

Sources/Models/User.swift
```diff
@@ -28,7 +29,8 @@ public struct User: Decodable, Equatable {
     referralCode: ReferralCode,
     referrerId: Id?,
     rssSalt: RssSalt,
-    subscriptionId: Subscription.Id?
+    subscriptionId: Subscription.Id?,
+    teamInviteCode: TeamInviteCode
```
stephencelis: Should the invite code be on the user or on the sub?

mbrandonw: On the user right? Why would it be on the sub?

PR #584, opened 2020-02-27
Update README.md

alexito4:
Hi guys,
I found the steps on the readme a little misleading. To run `make db` you need to be in your project folder, which of course makes sense, but following the steps linearly doesn't get you there 😅


stephencelis:
Looks good, thanks!

Issue #561, opened 2020-02-15
cmark.h not found


ZkHaider:
<img width="273" alt="image" src="https://user-images.githubusercontent.com/7844550/74595002-66e1ad00-4ff1-11ea-9e5b-31780fc8d133.png">

Steps to reproduce:

1.) Make sure `brew install cmark` is installed
2.) Run `brew services start postgresql`
3.) Clone pointfreeco via `git clone`
4.) Run `make db`
5.) Run `make`
6.) Run `swift package generate-xcodeproj`
7.) Hit CMD+B on PointfreeCo package to to get it to build. 

I'm assuming the PR to remove the `xcodeproj` building in the `Makefile` might have caused this?

stephencelis:
@ZkHaider Are you running things in Xcode 11.3 or higher? You should be able to open the project as a SwiftPM project now and avoid the Xcode project (you can delete it) entirely.

ZkHaider:
> @ZkHaider Are you running things in Xcode 11.3 or higher? You should be able to open the project as a SwiftPM project now and avoid the Xcode project (you can delete it) entirely.

Hey Stephen thanks for the direction :) love the work you guys have done at Pointfree and will try that and see if it works out ok.

alexito4:
Hi @stephencelis I found the same issue and you're right is fine if you open the Package directly. But I've been having issues with that (that's why I tried the project generation path). Just as an FYI. 

For example the `OSS` compiler directive is not there and it tries to compile https://github.com/pointfreeco/pointfreeco/blob/09514fe24e3879679bc830da218db7b839c4bd87/Sources/PointFree/Bootstrap.swift#L97
I commented the private episodes part and the Server compiled but crashed at runtime. Now it seems to work, I think that executing the Runner target made it work? 

Now that the server is running I tired the tests, and there are a lot that fail. Some screenshoots seem to have a scrollbar on my machine. Other tests are because the environment is different ($ vs. £) but I guess that's expected.

<details><summary>screenshoots</summary>
<p>

![public png_2_7DDDFD62-36F2-4EFF-B836-5C486A57BDA1](https://user-images.githubusercontent.com/750912/75477184-2fcdad00-599c-11ea-989c-849c2a3c9ea7.png)
![public png_1_7DDDFD62-36F2-4EFF-B836-5C486A57BDA1](https://user-images.githubusercontent.com/750912/75477188-30feda00-599c-11ea-89e7-75cd13c73b0e.png)
[...]

PR #585, opened 2020-02-27
Removed ".playground" suffix from the Downloads

alexito4:
Your downloads link always have a ".playground" suffix:

Screenshot 2020-02-27 at 20 34 03

I think this made sense in the beginning when your downloads were always Playgrounds. But recently, with the composable architecture series, that's not longer the case. 

Maybe is just me, but I was a little confused when clicking on a link that says "playground" and found a GitHub page with no playground in it :P

Maybe you can consider this small change 😉 

Screenshot 2020-02-27 at 20 36 56

(as described here https://github.com/pointfreeco/pointfreeco/issues/561#issuecomment-592128341 I was having issues with the test, but I think the snapshots are updated properly now.)


stephencelis:
Ha that's true! We'll chat about it

Sources/Views/EpisodePage.swift
```diff
@@ -174,7 +174,7 @@ private func downloadsView(codeSampleDirectory: String) -> Node {
         .href(gitHubUrl(to: GitHubRoute.episodeCodeSample(directory: codeSampleDirectory))),
         .class([Class.pf.colors.link.yellow, Class.margin([.mobile: [.left: 1]]), Class.align.middle])
       ],
-      .text("\(codeSampleDirectory).playground")
+      .text("\(codeSampleDirectory)")
```
stephencelis: I think we can even pass this directly now!
```suggestion
      .text(codeSampleDirectory)
```

stephencelis:
Thanks for taking the time to get things running and making this PR 👍 Welcome to the PF team 😂

Sources/Views/EpisodePage.swift
alexito4: You're right 😅 sorry, it was a long day ^^

alexito4:
Thanks, I'm glad I could help :)

PR #586, opened 2020-02-27
Fix Package.swift to work in Xcode OSS

stephencelis:
Fixes an issue @alexito4 brought up in #561.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,17 @@
 // swift-tools-version:5.1
 
+import Foundation
 import PackageDescription
+
+let isOss = !FileManager.default.fileExists(
+  atPath: URL(fileURLWithPath: #file)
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -12,6 +23,14 @@ extension SwiftSetting {
   )
 }
 
+extension Array where Element == SwiftSetting {
+  static let pointFreeSettings: Array = isOss
+    ? [.define("OSS"), .warnLongExpressionTypeChecking]
+    : [.warnLongExpressionTypeChecking]
+}
+
+let oss = FileManager.default.fileExists(atPath: "Sources/Models/Transcripts/.git")
```
mbrandonw: is this being used?

stephencelis: Oh woops

PR #587, opened 2020-02-27
Run GitHub Actions on PR

stephencelis:
Hopefully this is all that's needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - master
   pull_request:
     branches:
-      - master
+      - '*'
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
```
stephencelis: This will only run CI when code is pushed to master.

.github/workflows/ci.yml
```diff
@@ -1,6 +1,12 @@
...
+  pull_request:
+    branches:
+      - '*'
```
stephencelis: And this will run CI against all PRs...

PR #596, opened 2020-03-05
Collections Section

stephencelis:
Still a WIP:

- [ ] Top/bottom navigation
- [ ] Episode linking
- [ ] Pull in the play SVG from the new ep page


Sources/PointFree/Collections/CollectionSection.swift
```diff
@@ -0,0 +1,49 @@
...
+    <<< fetchCollectionSectionMiddleware
+    <| map(lower)
+    >>> writeStatus(.ok)
+    >=> respond(
+      view: collectionSection,
+      layoutData: { currentUser, _, collection, section in
+        SimplePageLayoutData(
+          currentUser: currentUser,
+          data: (collection, section),
+          extraStyles: collectionIndexStylesheet,
```
mbrandonw: we should add a trello card or a todo somewhere to track that we need to add meta tags to all of our new pages

PR #597, opened 2020-03-06
Start of new episode page

mbrandonw:
Still lots to do, but this is the basics. I haven't written tests because there is a lot more to do, and the test suite for the episode page is pretty extensive, so want to capture all of that later so I can compare them side-by-side.

image


Sources/PointFree/Episode/NewShow.swift
```diff
@@ -0,0 +1,45 @@
...
+    )
+    <<< fetchEpisodeForParam
+    <| writeStatus(.ok)
+    >=> userEpisodePermission
+    >=> map(lower)
+    >>> respond(
+      view: Views.newEpisodePageView(episodePageData:),
+      layoutData: { permission, episode, currentUser, subscriberState, currentRoute in
+        let previousEpisode = Current.episodes().first(where: { $0.sequence == episode.sequence - 1 })
+        let nextEpisode = Current.episodes().first(where: { $0.sequence == episode.sequence + 1 })
```
mbrandonw: I'm using the side bar to show only the previous/next episodes. I'm not integrating the collections stuff at all yet.

Sources/PointFree/Episode/Show.swift
```diff
@@ -50,7 +50,7 @@ let useCreditResponse =
     <<< validateUserEpisodePermission
     <| applyCreditMiddleware
 
-private let fetchEpisodeForParam
+let fetchEpisodeForParam
```
mbrandonw: made a few things internals so that I could reuse them

Sources/Styleguide/PointFreeBaseStyles.swift
```diff
@@ -30,6 +30,7 @@ extension Class {
         public static let dark = CssSelector.class("bg-dark")
         public static let gray150 = CssSelector.class("bg-gray150")
         public static let gray650 = CssSelector.class("bg-gray650")
+        public static let gray850 = CssSelector.class("bg-gray-850")
```
mbrandonw: this causes a lot of snapshot churn

stephencelis: Should it also be?
```suggestion
        public static let gray850 = CssSelector.class("bg-gray850")
```

PR #599, opened 2020-03-06
Update Episode Route to Use Episode

stephencelis:


stephencelis:
This isn't ready...some of the mock stuff made it more complicated, so let's chat about it.

stephencelis:
One failing test around episode credits that I haven't quite figured out yet but my brain's a bit zonked...

Sources/PointFree/Episode/Show.swift
```diff
@@ -13,15 +13,18 @@ import Styleguide
...
+  = writeStatus(.ok)
     >=> userEpisodePermission
     >=> map(lower)
     >>> respond(
       view: Views.episodeView(episodePageData:),
       layoutData: { permission, episode, currentUser, subscriberState, currentRoute in
         let navStyle: NavStyle = currentUser == nil ? .mountains(.main) : .minimal(.light)
+        #if DEBUG
+        var episode = Current.episodes().first(where: { $0.sequence == episode.sequence }) ?? episode
+        #endif
```
stephencelis: These lines are necessary for routing to go through tests. You might have a better idea...

mbrandonw: one thought comes to mind, we could inject the episodes into the router here: https://github.com/pointfreeco/pointfreeco/blob/master/Sources/PointFree/Bootstrap.swift#L88

PR #600, opened 2020-03-07
A few small collection fixes

mbrandonw:
* Fixed the bottom navigation
* Fixed hover color for lessons
* Made the back-to-collection link a bit whiter, was too dark
* Cleaned up some padding


Sources/Views/Collections/CollectionSection.swift
```diff
@@ -71,7 +71,7 @@ private func coreLessons(_ lessons: [Episode.Collection.Section.Lesson]) -> Node
           .class([
             Class.padding([
               .desktop: [.leftRight: 5, .top: 3, .bottom: 4],
-              .mobile: [.leftRight: 3, .topBottom: 2],
+              .mobile: [.leftRight: 3, .top: 2, .bottom: 3],
```
mbrandonw: it looks more symmetric to give a bit more bottom padding than top, because the top gets a little extra since headers get padding.

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -153,7 +153,7 @@ private func relatedItems(_ relatedItems: [Episode.Collection.Section.Related])
     .gridRow(
       attributes: [
         .class([
-          Class.grid.between(.desktop),
+          Class.padding([.mobile: [.topBottom: 2], .desktop: [.topBottom: 3]])
```
mbrandonw: `between` shouldn't be needed since there's only one column taking up the full width. and this padding is from me moving it out of a child and into this parent.

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -187,7 +187,9 @@ private func relatedItem(_ relatedItem: Episode.Collection.Section.Related) -> N
   return .gridColumn(
     sizes: [.mobile: 12],
     attributes: [
-      .style(margin(top: .px(4))),
+    .class([
+      Class.padding([.mobile: [.bottom: 2]])
```
mbrandonw: it didn't see the `4px` margin was necessary, but I could have missed something. I added some extra bottom padding so that the blurb+episode for each related item was a little separated from the next one

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -197,21 +199,20 @@ private func relatedItem(_ relatedItem: Episode.Collection.Section.Related) -> N
...
+            .mobile: [.top: 1],
           ]),
           Class.padding([.mobile: [.leftRight: 2]]),
           Class.pf.collections.hoverBackground,
           Class.pf.collections.hoverLink,
           Class.pf.colors.border.gray800,
           Class.pf.colors.bg.gray900,
         ]),
-        .href(url(to: .episode(.show(.right(episode.id))))),
+        .href(url(to: .episode(.show(.left(episode.slug))))),
```
mbrandonw: I think our current route doesn't even work with `.right(Episode.Id)`s. should probably just remove that version of the route, doesn't seem we even use it

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -197,21 +199,20 @@ private func relatedItem(_ relatedItem: Episode.Collection.Section.Related) -> N
...
         .style(
-          backgroundColor(.other("#fafafa"))
-          <> borderColor(all: .other("#e8e8e8"))
-          <> borderWidth(left: .px(4))
-          <> height(.px(48))
+          borderColor(all: .other("#e8e8e8"))
```
mbrandonw: removed the background color inline style so that it didn't override the hover.

I think we should move off of inline styles as much as possible. it's hard to do when you need to give a little bump to something (like the 4px margin between lessons), but other times we could either pick something as close as possible to what is already in our style guide, or we could introduce new utility classes.

mbrandonw: I also removed the explicit `height: 48px` and instead relied on the internal padding of the row to give it its height

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -377,10 +376,13 @@ private func sectionNavigation(
         .class([
           Class.flex.items.center,
           Class.grid.center(.mobile),
+          Class.padding([
+            .desktop: [.leftRight: 5],
+            .mobile: [.leftRight: 3],
+          ])
         ]),
         .style(
-          height(.px(88))
```
mbrandonw: I removed the explicit height on the bottom navigation and instead rely on the padding to define its height. this allows us to make the border extend from the top to the bottom of the row

Sources/Views/Svg.swift
```diff
@@ -9,7 +9,6 @@ public func playIconSvgBase64(fill: String = "121212") -> String {
   base64EncodedString("""
 <svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
   <g fill="none" fill-rule="evenodd">
-    <path fill="#FFF" d="M-73-1123H951v3583H-73z"/>
```
mbrandonw: this was my bad on master, there shouldn't be a white background

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -102,17 +102,16 @@ private func coreLesson(_ lesson: Episode.Collection.Section.Lesson) -> Node {
...
           Class.grid.row,
-          Class.padding([.mobile: [.leftRight: 2]]),
+          Class.padding([.mobile: [.leftRight: 2, .topBottom: 2]]),
           Class.pf.collections.hoverBackground,
           Class.pf.collections.hoverLink,
           Class.pf.colors.border.gray800,
           Class.pf.colors.bg.white,
         ]),
-        .href(url(to: .episode(.show(.right(lesson.episode.id))))),
+        .href(url(to: .episode(.show(.left(lesson.episode.slug))))),
```
stephencelis: Ah yeah. This was part of what made me go down the rabbit hole.

PR #601, opened 2020-03-08
Episode page prep before final cut over

mbrandonw:
This gets the new episode page very close. The next PR will convert the tests, and cut over everyone to use the new page. A few things to note:

* The side bar now shows all the collection episodes when in that context, and otherwise just shows previous/next episodes.
* I dropped the extra/secret route and instead just check if the current user is an admin to see if the page should be served or not. It was becoming too much of a pain to maintain both middlewares at once
* I think we need to resolve the optional collection slug soon. I had to force unwrap in a few spots, and it introduced a subtle bug where I was searching for a collection via a slug, and so if the slug was `nil` it just matched the first collection that had a `nil` slug. It's possible to fix that case, but makes me think we have more subtle bugs lying in wake. If we decide optional is the way to go we could maybe just use an enum instead of an optional to make it very clear why it is that way.
* I generalized some middlewares and transformers so that I could reuse them in a few spots, but that introduced some churn in the PR. It may be easier to read some parts as a split-diff


.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - '*'
```
mbrandonw: we chatted about this, but doing this so that we get CI running on all of our branches, not just PR's. it introduces a double run on PR open, but seems worth it. Travis did the same thing, not sure if there's a way around it

Sources/FunctionalCss/FlexGrid.swift
```diff
@@ -1,6 +1,7 @@
 import Css
 import Prelude
 
+// http://flexboxgrid.com
```
mbrandonw: this is the grid that I modeled our grid after, and I keep forgetting where i got it from, so just putting it here for documentation.

Sources/Models/Collections/AlgebraicDataTypes.swift
```diff
@@ -17,6 +17,7 @@ TODO
       whereToGoFromHere: #"""
 TODO
 """#
-    )
+    ),
+    slug: "algebraic-data-types"
```
mbrandonw: added this slug so that it wouldn't be `nil`. if we really do want to make it `nil` we can revert this and I can do extra `nil` checking elsewhere.

Sources/PointFree/Episode/NewShow.swift
```diff
@@ -1,45 +0,0 @@
-import Either
-import HttpPipeline
-import Models
-import PointFreeRouter
-import Prelude
-import Tuple
-import Views
-
-let newEpisodeResponse: M<Tuple4<Either<String, Episode.Id>, User?, SubscriberState, Route?>>
```
mbrandonw: getting rid of this temporary middleware, and now have it consolidated in the original `episodeResponse` middleware

Sources/PointFree/Episode/Show.swift
```diff
@@ -7,33 +7,36 @@ import HtmlCssSupport
...
 import Models
+import PointFreePrelude
 import PointFreeRouter
 import Prelude
 import Styleguide
 import Tuple
 import Views
 
-let episodeResponse =
+let episodeResponse: M<Tuple5<Either<String, Episode.Id>, User?, SubscriberState, Route?, Episode.Collection.Slug?>> =
```
mbrandonw: middleware now takes an optional collection slug for the case that you are viewing this episode in the context of a collection.

Sources/PointFree/Episode/Show.swift
```diff
@@ -7,33 +7,36 @@ import HtmlCssSupport
...
   fetchEpisodeForParam
     <| writeStatus(.ok)
     >=> userEpisodePermission
     >=> map(lower)
     >>> respond(
-      view: Views.episodeView(episodePageData:),
-      layoutData: { permission, episode, currentUser, subscriberState, currentRoute in
-        let navStyle: NavStyle = currentUser == nil ? .mountains(.main) : .minimal(.light)
+      view: newOrLegacyEpisodeView(data:),
```
mbrandonw: this helper view allows us to check if the current user is an admin to determine which episode to server

Sources/PointFree/Episode/Show.swift
```diff
@@ -7,33 +7,36 @@ import HtmlCssSupport
...
+      layoutData: { permission, episode, currentUser, subscriberState, currentRoute, collectionSlug in
+        let navStyle: NavStyle =
+          currentUser?.isAdmin == .some(true) ? .minimal(.black)
+            : currentUser == nil ? .mountains(.main)
+            : .minimal(.light)
```
mbrandonw: some logic around which navigation to show based on new/old episode page

Sources/PointFree/Episode/Show.swift
```diff
@@ -45,29 +48,86 @@ let episodeResponse =
     }
 )
 
-let useCreditResponse =
-  fetchEpisodeForParam
+private func episodePageData(
```
mbrandonw: this function produces the data necessary to render either the old or new episode page. We already had `EpisodePageData`, which I duplicated to `NewEpisodePageData` that holds data specific to the new page, and then I map that data to an `Either<EpisodePageData, NewEpisodePageData>`. Once we launch we will be able to delete and clean up this stuff.

Sources/PointFree/Episode/Show.swift
```diff
@@ -45,29 +48,86 @@ let episodeResponse =
...
+  currentUser: User?,
+  collectionSlug: Episode.Collection.Slug?,
+  episode: Episode,
+  permission: EpisodePermission,
+  subscriberState: SubscriberState
+) -> Either<EpisodePageData, NewEpisodePageData> {
+  if currentUser?.isAdmin == .some(true) {
```
mbrandonw: admins get the new episode page, and in order to show the new episode page we have to determine if we are in the context of a collection or the context of a direct episode visit.

Sources/PointFree/Episode/Show.swift
```diff
@@ -45,29 +48,86 @@ let episodeResponse =
...
+}
+
+private func newOrLegacyEpisodeView(data: Either<EpisodePageData, NewEpisodePageData>) -> Node {
+  switch data {
+  case let .left(data): return episodeView(episodePageData: data)
+  case let .right(data): return newEpisodePageView(episodePageData: data)
+  }
+}
+
+func useCreditResponse<Z>(
```
mbrandonw: this is an upgrade of `useCreditResponse` so that it doesn't care about what data is tacked onto the end of the tuple.

Sources/PointFree/Episode/Show.swift
```diff
@@ -45,29 +48,86 @@ let episodeResponse =
...
+    <| applyCreditMiddleware)
+}
 
-private let validateUserEpisodePermission
-  : MT<
-  Tuple4<Episode, User?, SubscriberState, Route?>,
-  Tuple5<EpisodePermission, Episode, User, SubscriberState, Route?>
-  >
-  = { userEpisodePermission >=> $0 }
+private func fetchEpisodeForParam<Z>(
```
mbrandonw: same here

Sources/PointFree/Episode/Show.swift
```diff
@@ -45,29 +48,86 @@ let episodeResponse =
...
+  middleware: @escaping M<T5<Episode, User?, SubscriberState, Route?, Z>>
+) -> M<T5<Either<String, Episode.Id>, User?, SubscriberState, Route?, Z>> {
+  middleware
+    |> filterMap(
+      over1(episode(forParam:)) >>> require1 >>> pure,
+      or: episodeNotFoundResponse
+  )
+}
+
+private func episodeNotFoundResponse<Z>(
```
mbrandonw: and here

Sources/PointFree/Episode/Show.swift
```diff
@@ -45,29 +48,86 @@ let episodeResponse =
...
+  conn: Conn<StatusLineOpen, T5<Either<String, Episode.Id>, User?, SubscriberState, Route?, Z>>
+) -> IO<Conn<ResponseEnded, Data>> {
+  conn
+    |> writeStatus(.notFound)
+    >=> respond { episodeNotFoundView(user: get2($0), subscriberState: get3($0), route: get4($0)) }
+}
+
+private func validateUserEpisodePermission<Z>(
```
mbrandonw: and here

all of that allows us to reuse those middlewares in a few situations.

for example, the middleware that uses a credit was being run off of some of these, which required us to pass along a collection slug even though it doesn't care about it. by generalizing it a bit I was able to make it take only the data it needed.

Sources/PointFree/Episode/Show.swift
```diff
@@ -210,39 +270,41 @@ func userEpisodePermission<I, Z>(
       .map { conn.map(const($0 .*. conn.data)) }
 }
 
-
-private let episodeNotFoundView = { param, user, subscriberState, route in
+private func episodeNotFoundView(
```
mbrandonw: the `episodeNotFoundView` was previously super super messy, taking a whole bunch of params that weren't necessary and then were just ignored with `_`s. I finally cleaned it up a bit, though I think the ergonomics can be improved a lot more.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -55,6 +55,10 @@ private func render(conn: Conn<StatusLineOpen, T3<(Models.Subscription, Enterpri
       return conn.map(const(user .*. subscriberState .*. route .*. subRoute .*. unit))
         |> blogMiddleware
 
+    case let .collections(.episode(collectionSlug, sectionSlug, episodeParam)):
+      return conn.map(const(episodeParam .*. user .*. subscriberState .*. route .*. collectionSlug .*. unit))
+        |> episodeResponse
```
mbrandonw: you can now visit an episode page in the context of a collection

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -86,16 +90,12 @@ private func render(conn: Conn<StatusLineOpen, T3<(Models.Subscription, Enterpri
       return conn
         |> redirect(to: path(to: .home))
 
-    case let .episode(.newShow(param)):
-      return conn.map(const(param .*. user .*. subscriberState .*. route .*. unit))
-        |> newEpisodeResponse
```
mbrandonw: the new route was adding a lot of complexity so dropped it

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -86,16 +90,12 @@ private func render(conn: Conn<StatusLineOpen, T3<(Models.Subscription, Enterpri
...
-
     case let .episode(.progress(param: param, percent: percent)):
       return conn.map(const(param .*. user .*. subscriberState .*. percent .*. unit))
         |> progressResponse
 
     case let .episode(.show(param)):
-      return conn.map(const(param .*. user .*. subscriberState .*. route .*. unit))
+      return conn.map(const(param .*. user .*. subscriberState .*. route .*. nil .*. unit))
```
mbrandonw: existing episode route passes `nil` for collection since it's a direct visit

Sources/PointFreeRouter/Routes.swift
```diff
@@ -61,6 +61,7 @@ public enum Route: Equatable {
   }
 
   public enum Collections: Equatable {
+    case episode(Episode.Collection.Slug, Episode.Collection.Section.Slug, Either<String, Episode.Id>)
```
mbrandonw: new route for visiting an episode in the context of a collection.

Sources/PointFreeRouter/Routes.swift
```diff
@@ -174,12 +174,16 @@ let routers: [Router<Route>] = [
   .case { .collections(.section($0, $1)) }
     <¢> get %> "collections" %> pathParam(.tagged(.string)) <%> pathParam(.tagged(.string)) <% end,
 
+  parenthesize(.case { .collections(.episode($0, $1, $2)) })
+    <¢> get %> "collections"
+    %> pathParam(.tagged(.string))
+    <%> pathParam(.tagged(.string))
+    <%> pathParam(.episodeIdOrString)
+    <% end,
```
mbrandonw: this route produces things like `/collections/map-zip-flatmap/map/e13-many-faces-of-map`. 

down to workshop this url a bit more if you have ideas.

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -42,7 +42,7 @@ public func collectionSection(
       length: section.length,
       blurb: section.blurb
     ),
-    coreLessons(section.coreLessons),
+    coreLessons(collection: collection, section: section),
```
mbrandonw: there's a few spots we need to thread collection and section info deeper into the UI so that we can link to episode pages properly

Sources/Views/Collections/CollectionSection.swift
```diff
@@ -108,7 +117,8 @@ private func coreLesson(_ lesson: Episode.Collection.Section.Lesson) -> Node {
           Class.pf.colors.border.gray800,
           Class.pf.colors.bg.white,
         ]),
-        .href(url(to: .episode(.show(.right(lesson.episode.id))))),
+        // TODO: figure out force unwrap
+        .href(url(to: .collections(.episode(collection.slug!, section.slug, .left(lesson.episode.slug))))),
```
mbrandonw: here's an example of where I need to force unwrap

Sources/Views/EpisodePage.swift
```diff
@@ -452,13 +452,20 @@ private func sectionsMenu(episode: Episode, permission: EpisodePermission?) -> N
 
 let divider = Node.hr(attributes: [.class([Class.pf.components.divider])])
 
-func transcriptView(blocks: [Episode.TranscriptBlock], isEpisodeViewable: Bool) -> Node {
+func transcriptView(
+  blocks: [Episode.TranscriptBlock],
+  isEpisodeViewable: Bool,
+  needsExtraPadding: Bool = true
```
mbrandonw: this is a hack that will be cleaned up soon. the new episode page needs less padding than the old and so I need a way to customizing this.

Sources/PointFreeRouter/Routes.swift
stephencelis: I think it works well. Easy to manually chop off a path component to get to another valid page. Could be down to prepend the section with its sequence, though.

PR #609, opened 2020-03-09
Collections design

mbrandonw:
This brings in some design for collections. It also puts the array of collections in the environment so that we can control it during testing.


Sources/Models/Collections/AlgebraicDataTypes.swift
```diff
@@ -2,15 +2,20 @@ extension Episode.Collection {
   public static let algebraicDataTypes = Self(
     section: .init(
       blurb: #"""
-TODO
+There is a wonderful correspondence between Swift's type system and algebra that we are all familiar with from grade school. By understanding this correspondence we can understand our data structures at a much higher level, and this allows us to remove invalid states from our types, thus making things we want to be impossible, actually impossible.
```
mbrandonw: started filling in some blurbs, more work needs to be done on these

Sources/ModelsTestSupport/Mocks.swift
```diff
@@ -99,6 +99,98 @@ extension Models.User {
   }
 }
 
+extension Episode.Collection {
+  public static let mock = Episode.Collection(
```
mbrandonw: added a mock for collections

Sources/PointFree/Collections/CollectionIndex.swift
```diff
@@ -14,13 +14,14 @@ let collectionsIndexMiddleware: M<Tuple3<User?, SubscriberState, Route?>>
     <| map(lower)
     >>> writeStatus(.ok)
     >=> respond(
-      view: { collectionIndex },
+      view: collectionIndex(collections:),
```
mbrandonw: this view function now takes the array of collections so that we can test it

Sources/PointFree/Collections/CollectionIndex.swift
```diff
@@ -14,13 +14,14 @@ let collectionsIndexMiddleware: M<Tuple3<User?, SubscriberState, Route?>>
...
       layoutData: { currentUser, subscriberState, route in
         SimplePageLayoutData(
           currentRoute: route,
           currentSubscriberState: subscriberState,
           currentUser: currentUser,
-          data: (),
+          data: Current.collections,
+          extraStyles: collectionIndexStyles,
```
mbrandonw: a few extra styles we have to inject into this view, but as I'll explain in a bit I think we could even get rid of this later...

Sources/PointFree/Collections/CollectionSection.swift
```diff
@@ -38,7 +38,7 @@ private let fetchCollectionSectionMiddleware
   = filterMap(
     {
       let (user, subscriberState, route, collectionSlug, sectionSlug) = lower($0)
-      return pure(Episode.Collection.all.first(where: { $0.slug == collectionSlug }).flatMap { collection in
+      return pure(Current.collections.first(where: { $0.slug == collectionSlug }).flatMap { collection in
```
mbrandonw: environment now holds collections

Sources/Views/Collections/CollectionIndex.swift
```diff
@@ -7,41 +7,184 @@ import PointFreeRouter
...
+      )
+    ),
+    .ul(
+      attributes: [
+        .class([
+          Class.margin([.mobile: [.all: 0]]),
+          Class.padding([.mobile: [.all: 0], .desktop: [.leftRight: 2, .topBottom: 0]]),
+          Class.type.list.styleNone,
+          Class.flex.wrap,
+          Class.flex.flex
```
mbrandonw: the tricky part about the design of the collections index is that in each row if one item is bigger than the other they should both grow to be the same size. I did this for the [pricing](http://www.pointfree.co/pricing) page, and it took some time to get rid, and so I just copied it for this one.

with a bit of time I could clean this up, but let's just roll with it for now

Sources/Views/Collections/CollectionIndex.swift
```diff
@@ -7,41 +7,184 @@ import PointFreeRouter
...
+  ]
+}
+
+private let colors = ["#4cccff", "#79f2b0", "#fff080", "#974dff"]
+private let combos = colors
+  .flatMap { color in colors.map { (color, $0) } }
+  .filter { $0 != $1 }
+
+private func collectionItem(collection: Episode.Collection, index: Int) -> ChildOf<Tag.Ul> {
+  let (lower, upper) = combos[index % combos.count]
```
mbrandonw: this cycles through all the color combos for the gradient

Sources/Views/Collections/CollectionIndex.swift
```diff
@@ -7,41 +7,184 @@ import PointFreeRouter
...
+//            // TODO: figure out force unwrap
+//            .href(url(to: .collections(.show(collection.slug!)))),
+//            .class([
+//              Class.align.middle,
+//              Class.pf.colors.link.purple,
+//              Class.pf.type.body.regular,
+//              Class.margin([.mobile: [.top: 4]])
+//            ])
+//          ],
+//          .text("See collection (\(10) episodes)"),
```
mbrandonw: I wanna have a CTA in the collection card, but I was having trouble getting it to anchor to the bottom (even though I was able to do it for the pricing page :/). I come back to this later

Tests/PointFreeTests/CollectionsTests.swift
```diff
@@ -19,9 +19,36 @@ class CollectionsTests: TestCase {
...
+        ]
+      )
+    }
+    #endif
+  }
+
   func testCollectionShow() {
     let conn = connection(
-      from: request(to: .collections(.show("map-zip-flat-map")), basicAuth: true)
+      from: request(to: .collections(.show(Current.collections[0].slug)), basicAuth: true)
```
mbrandonw: update tests to use mocks so that we can control their output

Sources/Views/NewEpisodePage.swift
```diff
@@ -792,7 +792,7 @@ private func episodeHeader(
         .div(
           attributes: [
             .class([
-              Class.padding([.mobile: [.top: 1, .leftRight: 4]]),
+              Class.padding([.mobile: [.top: 1, .leftRight: 2], .desktop: [.leftRight: 4]]),
```
mbrandonw: unrelated to the rest of the PR, just found some padding that was off in mobile

PR #612, opened 2020-03-09
Show collections in footer.

mbrandonw:
Hidden behind admin flag, will remove once we launch.


Sources/PointFree/ApiMiddleware.swift
```diff
@@ -28,7 +28,7 @@ extension Api {
       self.publishedAt = episode.publishedAt
       self.sequence = episode.sequence
       self.subscriberOnly = episode.isSubscriberOnly(currentDate: currentDate)
-      self.title = episode.title
+      self.title = episode.fullTitle
```
mbrandonw: unrelated to this pr, but just thought of this other spot we should show the full title

Tests/PointFreeTests/PricingLandingTests.swift
```diff
@@ -19,7 +19,7 @@ class PricingLandingIntegrationTests: LiveDatabaseTestCase {
   }
 
   func testLanding_LoggedIn_InactiveSubscriber() {
-    var user = User.admin
+    var user = User.mock
```
mbrandonw: some old tests were using admin mocks because these views used to be guarded for admins only. changing that now

PR #613, opened 2020-03-10
Collections Finesse

stephencelis:


Sources/Views/NewEpisodePage.swift
```diff
@@ -450,10 +443,9 @@ private func chaptersRow(episode: Episode) -> Node {
             sizes: [.mobile: 9],
             attributes: [
               .class([
+                Class.padding([.mobile: [.leftRight: 1]]),
                 Class.pf.type.body.small,
-                Class.padding([.mobile: [.leftRight: 1]])
-              ]),
-              .style(padding(topBottom: .px(2)))
```
stephencelis: I noticed that timestamps weren't aligned with chapter titles...

PR #614, opened 2020-03-10
Lots of little nits from @amotion

mbrandonw:
Chatted with @amotion about a bunch of small little improvements we could do to the episode page.


Sources/Views/NewEpisodePage.swift
```diff
@@ -228,21 +228,10 @@ private func collectionHeaderRow(
...
         attributes: [
           .class([
-            Class.pf.colors.fg.gray650,
-            Class.typeScale([.mobile: .r0_75]),
-            Class.type.lineHeight(1),
-            Class.type.caps,
-            Class.type.normal,
-            Class.padding([.mobile: [.all: 0]]),
-            Class.margin([.mobile: [.leftRight: 0, .top: 0, .bottom: 1]])
+            sidebarShoutTitleClass
```
mbrandonw: these classes were repeated for two titles, so extracted to something that could be shared

Sources/Views/NewEpisodePage.swift
```diff
@@ -200,7 +201,6 @@ private func sequentialEpisodes(
             attributes: [
               .class([
                 Class.pf.type.body.regular,
-                Class.pf.colors.link.gray650,
```
mbrandonw: rows in the side bar are intended to be gray only when they are "completed"

PR #616, opened 2020-03-10
Some Collections Blurbs

stephencelis:


Sources/Models/BlogPosts/BlogPost0025_yearInReview.swift
```diff
@@ -178,8 +178,8 @@ powers the HTML rendering of this very site!
 
 The design of this library was covered in 4 episodes:
 
-* [Domain Specific Languages: Part 1](/episodes/ep26-domain-specific-languages-part-1)
-* [Domain Specific Languages: Part 2](/episodes/ep27-domain-specific-languages-part-2)
+* [Domain-Specific Languages: Part 1](/episodes/ep26-domain-specific-languages-part-1)
+* [Domain-Specific Languages: Part 2](/episodes/ep27-domain-specific-languages-part-2)
```
stephencelis: Realizing these should be hyphenated...

Sources/Models/Episode.swift
```diff
@@ -202,12 +202,16 @@ public struct Episode: Equatable {
 
         public enum Content: Equatable {
           case episodes([Episode])
-          case collection(Collection)
+          case collections([Collection])
```
stephencelis: Added the ability to add a related set of collections...

PR #625, opened 2020-03-10
Fix some mobile prblems on episode page.

mbrandonw:


Sources/Views/NewEpisodePage.swift
```diff
@@ -116,7 +116,7 @@ private func sideBar(
           Class.border.all,
           Class.border.rounded.all,
           Class.pf.colors.border.gray850,
-          Class.margin([.mobile: [.left: 3]])
+          Class.margin([.mobile: [.leftRight: 2], .desktop: [.left: 3, .right: 0]])
```
mbrandonw: the margin that I had before was only appropriate for desktop, not mobile. on mobile we need both left and right margins for the side bar

Sources/Views/NewEpisodePage.swift
```diff
@@ -789,7 +789,7 @@ private func episodeHeader(
         .div(
           attributes: [
             .class([
-              Class.padding([.mobile: [.top: 1, .leftRight: 2], .desktop: [.leftRight: 4]]),
+              Class.padding([.mobile: [.top: 1, .leftRight: 0], .desktop: [.leftRight: 4]]),
```
mbrandonw: there was too much padding on the blurb

mbrandonw:
> no new snaps?

yeah just pushed, hopefully it goes green after that

PR #631, opened 2020-03-11
Make collection references lazy so that we can do cyclical references.

mbrandonw:
I was able to figure out the cyclical references in our collections. We needed to make the related content lazy _and_ we need to make the parsing collection a computed `var` instead of a `let`.

That causes a bunch of whitespace differences so you may want to view this diff without whitespace.

### DIFF
--- a/Sources/Models/Collections/Parsing.swift
+++ b/Sources/Models/Collections/Parsing.swift
@@ -1,73 +1,75 @@
 extension Episode.Collection {
-  public static let parsing = Self(
-    blurb: #"""
+  public static var parsing: Self {
+    Self(
+      blurb: #"""
 Parsing is a surprisingly ubiquitous problem in programming. Every time we construct an integer or a URL from a string, we are technically doing parsing. After demonstrating the many types of parsing that Apple gives us access to, we will take a step back and define the essence of parsing in a single type. That type supports many wonderful types of compositions, and allows us to break large, complex parsing problems into small, understandable units.
 """#,
-    sections: [
-      .init(
-        blurb: #"""
+      sections: [
+        .init(
+          blurb: #"""
 Parsing is a difficult, but surprisingly ubiquitous programming problem, and functional programming has a lot to say about it. Let's take a moment to understand the problem space of parsing, and see what tools Swift and Apple gives us to parse complex text formats.
 """#,
-        coreLessons: [
-          .init(episode: .ep56_whatIsAParser_pt1),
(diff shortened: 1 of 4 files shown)


Sources/Models/Collections/Parsing.swift
```diff
@@ -1,73 +1,75 @@
...
+          ],
+          related: [
+            // TODO: bring back when we figure out recursive references
+            .init(
+              blurb: #"""
+Parsing is just one of many problems functional programming solves by defining a core, composable, transformable unit. We apply these exact same techniques to randomness and even architecture!
+"""#,
+              content: .collections([
+                .randomness,
+                .composableArchitecture,
```
mbrandonw: can now bring back these references

Sources/Models/Episode.swift
```diff
@@ -201,17 +201,19 @@ public struct Episode: Equatable {
         }
 
         public enum Content: Equatable {
-          case episodes([Episode])
-          case collections([Collection])
-          case section(Collection, index: Int)
+          case episodes(@autoclosure () -> [Episode])
+          case collections(@autoclosure () -> [Collection])
+          case section(@autoclosure () -> Collection, index: Int)
```
mbrandonw: just gotta make everything lazy and make parsing a computed var

Sources/Models/Episode.swift
```diff
@@ -201,17 +201,19 @@ public struct Episode: Equatable {
...
+            .episodes([episode()])
           }
 
-          public static func collection(_ collection: Collection) -> Content {
-            .collections([collection])
+          public static func collection(_ collection: @escaping @autoclosure() -> Collection) -> Content {
+            .collections([collection()])
           }
+
+          public static func == (lhs: Content, rhs: Content) -> Bool { return true }
```
mbrandonw: oops will fix this 😬

Tests/ModelsTests/CollectionTests.swift
```diff
@@ -0,0 +1,8 @@
+import XCTest
+@testable import Models
+
+final class CollectionTests: XCTestCase {
+  func testAllCollections() {
+    XCTAssertEqual(Episode.Collection.all, Episode.Collection.all)
```
mbrandonw: I put this silly test in because without the laziness it will compile (using just computed vars) but it will crash at runtime. so just wanted a test to load up the collections to make sure all is fine

PR #632, opened 2020-03-11
Fix rounded corner in side bar.

mbrandonw:
Fixes a tiny rounded corner bug when the last row of the side bar is the current episode:

image


Sources/Views/EpisodePage.swift
```diff
@@ -111,42 +111,32 @@ private func sideBar(
     let nextEpisodes = section.coreLessons[(currentEpisodeIndex+1)...].map { $0.episode }
 
     return .div(
-      attributes: [
-        .class([
-          Class.border.all,
```
mbrandonw: all these classes were duplicated so I extracted them into a helper

Sources/Views/EpisodePage.swift
```diff
@@ -111,42 +111,32 @@ private func sideBar(
...
       sequentialEpisodeRow(episode: nextEpisode, type: .next)
     )
   }
 }
 
+private let sideBarClasses
+  = Class.border.all
+    | Class.border.rounded.all
+    | Class.pf.colors.border.gray850
+    | Class.layout.overflowHidden
```
mbrandonw: this is the key tho, making overflow hidden

stephencelis: nice

PR #640, opened 2020-03-27
Swift 5.2

stephencelis:
Went GM recently...


Package.swift
```diff
@@ -58,11 +58,11 @@ let package = Package(
         "GitHub",
         "Models",
         "Stripe",
-        .product(name: "Either", package: "swift-prelude"),
+        .product(name: "Either", package: "Prelude"),
```
stephencelis: There seems to be a lot of weirdness with how SPM package naming works, to the point where we can't even seem to upgrade to SPM tools version 5.2 without possibly renaming all of our upstream packages...

Package.swift
```diff
@@ -58,11 +58,11 @@ let package = Package(
...
         .product(name: "Logging", package: "swift-log"),
-        .product(name: "PostgreSQL", package: "postgresql"),
-        .product(name: "Prelude", package: "swift-prelude"),
-        .product(name: "Tagged", package: "swift-tagged"),
+        .product(name: "PostgreSQL", package: "PostgreSQL"),
```
stephencelis: (And even if that _does_ fix it, we don't control this dependency, so we'd have to fork it just to get the naming right.)

stephencelis:
Mac should start building again next week: https://github.com/actions/virtual-environments/issues/620

.github/workflows/ci.yml
```diff
@@ -1,27 +1,24 @@
...
+        brew uninstall --ignore-dependencies libpq
+        brew install cmark postgresql
+        initdb /usr/local/var/postgres
     - name: Start Postgres
-      run: brew services start postgresql || brew update && brew services start postgresql
-    - name: Select Xcode 11.3
-      run: sudo xcode-select -s /Applications/Xcode_11.3.app
+      run: brew services start postgresql
+    - name: Select Xcode 11.4
+      run: sudo xcode-select -s /Applications/Xcode_11.4.app
```
freak4pc, line 21: 🤯 where did you even bump into this trick lol 
is it documented somewhere they have all of these versions installed in that specific path?

stephencelis: There's a helpful doc here: https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners

Specifically if you click through to the Mac environment you get that info here: https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md#xcode

freak4pc: Thanks,
Good to know :) Applied this here with a matrix job and works well  https://github.com/CombineCommunity/CombineExt/blob/master/.github/workflows/tests.yml

PR #641, opened 2020-03-27
Use Stripe invoice behavior

stephencelis:


Sources/PointFree/Account/Change.swift
```diff
@@ -42,31 +42,8 @@ func changeSubscription(
     return { conn in
       let (currentSubscription, newPricing) = conn.data
 
-      let newPrice = defaultPricing(for: newPricing).map { $0 * newPricing.quantity }
```
stephencelis: We don't need to rely on this dance anymore. Stripe handles it for us.

PR #656, opened 2020-05-07
Regional discounts

mbrandonw:
This implements the core of regional discounts. There are just a few things for us to consider around edge cases and persisting coupons on subscription updates, but we can do that in another PR.


.pf-env.example
```diff
@@ -8,6 +8,7 @@
   "GITHUB_CLIENT_SECRET": "deadbeef-client-secret",
   "MAILGUN_DOMAIN": "mg.domain.com",
   "MAILGUN_PRIVATE_API_KEY": "deadbeef-mg-api-key",
+  "REGIONAL_DISCOUNT_COUPON_ID": "regional-discount",
```
mbrandonw, line 11: this is a coupon I created on our sandbox stripe

Sources/Models/SubscribeConfirmationData.swift
```diff
@@ -6,16 +6,19 @@ public struct SubscribeConfirmationData: Equatable {
   public var isOwnerTakingSeat: Bool
   public var referralCode: User.ReferralCode?
   public var teammates: [EmailAddress]
+  public var useRegionalDiscount: Bool
```
mbrandonw, line 9: this is the data used for the subscription confirmation page at `/subscribe`

Sources/Models/SubscribeData.swift
```diff
@@ -8,21 +8,24 @@ public struct SubscribeData: Equatable {
   public var referralCode: User.ReferralCode?
   public var teammates: [EmailAddress]
   public var token: Stripe.Token.Id
+  public var useRegionalDiscount: Bool
```
mbrandonw, line 11: this is the data that is submitted in the POST request to `/subscribe`

Sources/PointFree/Subscribe.swift
```diff
@@ -27,94 +28,139 @@ private let validateSubscribeData
     <<< validateCoupon
     <<< validateReferrer
 
-private func subscribe(_ conn: Conn<StatusLineOpen, Tuple3<User, SubscribeData, Referrer?>>)
-  -> IO<Conn<ResponseEnded, Data>> {
-
-    let referralDiscount: Cents<Int> = -18_00
+private func subscribe(
```
mbrandonw, line 31: the core the discounting logic is in here...

Sources/PointFree/Subscribe.swift
```diff
@@ -27,94 +28,139 @@ private let validateSubscribeData
...
+) -> IO<Conn<ResponseEnded, Data>> {
+
+  let (user, subscribeData, referrer) = lower(conn.data)
+  let referrerDiscount: Cents<Int> =
+    referrer?.stripeSubscription.discount?.coupon.id == Current.envVars.regionalDiscountCouponId
+      ? -9_00
+      : -18_00
+  let referredDiscount: Cents<Int> = subscribeData.useRegionalDiscount
+    ? -9_00
+    : -18_00
```
mbrandonw, line 42: this computes the referral discount we give out based on if the regional discount is applied.

this means the referred person only gets a $9 credit if they use the regional discount, and the referrer only gets $9 if they used the regional discount.

Sources/PointFree/Subscribe.swift
```diff
@@ -27,94 +28,139 @@ private let validateSubscribeData
...
+  )
+    .flatMap { customer -> EitherIO<Error, Stripe.Subscription> in
+      let country = customer.sources.data.first?.left?.country
+
+      guard country != nil || !subscribeData.useRegionalDiscount else {
+        return throwE(
+          StripeErrorEnvelope(
+            error: .init(
+              message: """
+                Couldn't verify issue country on credit card. Please try another credit card.
```
mbrandonw, line 59: this just guards against us not having a card in the `sources` array, which I don't think will ever actually happen.

Sources/PointFree/Subscribe.swift
```diff
@@ -27,94 +28,139 @@ private let validateSubscribeData
...
+                """
+            )
+          )
+        )
+      }
 
-    let (user, subscribeData, referrer) = lower(conn.data)
+      guard !subscribeData.useRegionalDiscount
+        || DiscountCountry.all.contains(where: { $0.countryCode == country })
```
mbrandonw, line 67: this guards against using the regional discount when the CC country is not in our list of allowed ones.

Sources/PointFree/Subscribe.swift
```diff
@@ -27,94 +28,139 @@ private let validateSubscribeData
...
-    let customer = Current.stripe.createCustomer(
-      subscribeData.token,
-      user.id.rawValue.uuidString,
-      user.email,
-      nil,
-      subscribeData.pricing.interval == .year ? referrer.map(const(referralDiscount)) : nil
-    )
+      let regionalDiscountCouponId = subscribeData.useRegionalDiscount
+        ? Current.envVars.regionalDiscountCouponId
+        : nil
```
mbrandonw, line 84: the regional discount is really just a 50% coupon under the hood.

Sources/PointFree/Subscribe.swift
```diff
@@ -150,15 +196,26 @@ private func validateQuantity(_ pricing: Pricing) -> Bool {
...
+      |> redirect(to: .login(redirect: url(to: .pricingLanding)))
 }
 
 private func validateCoupon(forSubscribeData subscribeData: SubscribeData) -> Bool {
-  return subscribeData.coupon == nil || subscribeData.pricing.quantity == 1
+  subscribeData.coupon == nil
+    // Do not allow using coupons on team subscriptions
+    || subscribeData.pricing.quantity == 1
+    // Do not allow using regional discount coupon id directly
+    && subscribeData.coupon != Current.envVars.regionalDiscountCouponId
```
mbrandonw, line 211: making sure that if the regional discount coupon id leaks that people can't use it except through that dedicated flow.

Sources/PointFree/Subscribe.swift
```diff
@@ -150,15 +196,26 @@ private func validateQuantity(_ pricing: Pricing) -> Bool {
...
+}
+
+private func validateCouponAndRegionalDiscount(
+  forSubscribeData subscribeData: SubscribeData
+) -> Bool {
+  // Don't allow using coupon and regional discount at once
+  subscribeData.coupon == nil || !subscribeData.useRegionalDiscount
```
mbrandonw, line 218: and people shouldn't be allowed to use coupons with a regional discount.

Sources/Stripe/Model.swift
```diff
@@ -98,7 +103,7 @@ public struct Coupon: Equatable {
     case let .percentOff(percentOff):
       return cents.map { cents -> Int in
         let discountPercent = Double(100 - percentOff) / 100
-        return Int(Double(cents) * discountPercent)
+        return Int(round(Double(cents) * discountPercent))
```
mbrandonw, line 106: unrelated to regional discounts, but fixes a floating point error

Sources/Stripe/Model.swift
```diff
@@ -5,6 +5,7 @@ import TaggedMoney
 
 public struct Card: Codable, Equatable {
   public var brand: Brand
+  public var country: Country
```
mbrandonw, line 8: as far as I know this is a non-optional field from Stripe

Sources/Views/PricingLanding.swift
```diff
@@ -357,7 +358,10 @@ func pricingPlan(
           plan.features.map { feature in
             .li(
               attributes: [.class([Class.padding([.mobile: [.top: 1]])])],
-              [.text(feature)]
+              .div(
+                attributes: [.class([pricingPlanFeatureClass])],
+                .raw(unsafeMark(from: feature))
```
mbrandonw, line 363: started using some markdown in the plan features content

Sources/Views/PricingLanding.swift
```diff
@@ -560,14 +565,15 @@ private let faqItems: [Node] = Faq.allFaqs.flatMap { faq in
       ],
       .text(faq.question)
     ),
-    .p(
+    .markdownBlock(
```
mbrandonw, line 568: and in some FAQ answers

Sources/Views/SubscriptionConfirmation.swift
```diff
@@ -37,14 +37,155 @@ public func subscriptionConfirmation(
...
+      isLoggedIn: currentUser != nil,
+      lane: lane,
+      coupon: coupon,
+      referrer: referrer,
+      useRegionalDiscount: subscribeData.useRegionalDiscount
+    )
+  )
+}
+
+private func additionalDiscountInfo(
```
mbrandonw, line 59: this function determines what to show in the yellow info box. the logic is pretty gnarly because there are 4 possibilities:

image

image

image

image

Sources/PointFreeRouter/Routes.swift
```diff
@@ -389,6 +396,7 @@ private let subscriberDataIso = PartialIso<String, SubscribeData?>(
     if let referralCode = data.referralCode?.rawValue {
       parts.append("\(SubscribeData.CodingKeys.referralCode.rawValue)=\(referralCode)")
     }
+    parts.append("\(SubscribeData.CodingKeys.useRegionalDiscount.rawValue)=\(data.useRegionalDiscount)")
```
stephencelis, line 399: Should we only append this if it's true?

Sources/Stripe/Model.swift
stephencelis: 🙏

Sources/PointFreeRouter/Routes.swift
mbrandonw: yep can do...

stephencelis: After reviewing all the snaps it should reduce some churn there, too.

PR #667, opened 2020-05-27
Subscription Deactivation

stephencelis:


Sources/Models/SubscriberState.swift
```diff
@@ -27,16 +32,26 @@ public enum SubscriberState {
...
   }
 
+  public var deactivated: Bool? {
+    switch self {
+    case .nonSubscriber:
+      return nil
+    case let .owner(_, _, _, deactivated):
+      return deactivated
+    case let .teammate(_, _, deactivated):
+      return deactivated
```
mbrandonw, line 49: could be this too right?

```suggestion
    case let .owner(_, _, _, deactivated),
             .teammate(_, _, deactivated):
      return deactivated
```

PR #674, opened 2020-07-21
Add SwiftUI Collection

stephencelis:


Sources/Models/Collections/EnumsAndStructs.swift
```diff
@@ -53,6 +53,24 @@ Let's apply case paths to a real-world use case: application architecture! We'll
         ],
         title: "Case Paths",
         whereToGoFromHere: nil
+//        whereToGoFromHere: #"""
```
stephencelis, line 56: We have a circular reference to this collection in the Composable SwiftUI Bindings series...wonder if we can cook up a better fix for this issue...

PR #675, opened 2020-07-27
Properly track feed request events.

mbrandonw:
Just a quick PR to make sure we are tracking feed request events, even with an invalid user agent.

### DIFF
--- a/Sources/PointFree/Account/PrivateRss.swift
+++ b/Sources/PointFree/Account/PrivateRss.swift
@@ -12,6 +12,7 @@ import Syndication
 import Tuple
 
 let accountRssMiddleware = decryptUrlAndFetchUser
+  <<< { trackFeedRequest(userId: \.first.id) >=> $0 }
   <<< validateUserAndSaltAndUserAgent
   <<< fetchActiveStripeSubscription
   <| map(lower)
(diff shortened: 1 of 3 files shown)


Sources/PointFree/Account/PrivateRss.swift
```diff
@@ -12,6 +12,7 @@ import Syndication
 import Tuple
 
 let accountRssMiddleware = decryptUrlAndFetchUser
+  <<< { trackFeedRequest(userId: \.first.id) >=> $0 }
```
mbrandonw, line 15: This is the core fix. We were tracking the feed request too late in the chain of middleware, and so we would detect the invalid UA and then prevent executing any more of the middleware. This moves it up higher. I should prob refactor this to be a middleware transformer instead of middleware, but can do that later.

Tests/PointFreeTests/PrivateRssTests.swift
```diff
@@ -176,6 +181,36 @@ class PrivateRssTests: TestCase {
 
     let conn = connection(from: req)
 
+    assertSnapshot(matching: conn |> siteMiddleware, as: .ioConn)
+    XCTAssertTrue(feedRequestEventCreated)
```
mbrandonw, line 185: this test confirms that the `createFeedRequestEvent` database endpoint is called even in the invalid UA flow.

Sources/PointFree/Account/PrivateRss.swift
```diff
@@ -84,7 +85,6 @@ private let requireActiveSubscription: (
 private let accountRssResponse
   : Middleware<StatusLineOpen, ResponseEnded, (Stripe.Subscription?, User), Data>
   = writeStatus(.ok)
-    >=> trackFeedRequest
```
stephencelis, line 87: Is this not necessary here?

stephencelis: Oh I blew past the shuffle above.

PR #676, opened 2020-07-29
Create deploy.yml

stephencelis:
This prob won't work as-is but needs to be merged to start kicking off manual builds for debugging, which I'll do in a followup PR.


stephencelis:
K this is ready to go!

PR #690, opened 2021-02-08
Limit tokens to established GitHub accounts

stephencelis:


Sources/GitHub/Model.swift
```diff
@@ -56,6 +59,12 @@ public struct GitHubUser: Codable {
   }
 
   public typealias Id = Tagged<GitHubUser, Int>
+
+  private enum CodingKeys: String, CodingKey {
+    case createdAt = "created_at"
```
stephencelis, line 64: Be nice to figure out if we can finally do the snake case stuff...

Sources/Database/Database.swift
```diff
@@ -498,21 +499,24 @@ private struct _Client {
...
     ON CONFLICT ("github_user_id") DO UPDATE
     SET "github_access_token" = $3, "name" = $4
     """,
       [
         email.rawValue,
         envelope.gitHubUser.id.rawValue,
         envelope.accessToken.accessToken,
-        envelope.gitHubUser.name
+        envelope.gitHubUser.name,
+        envelope.gitHubUser.createdAt.distance(to: now()) < 60*60*24*7 ? 0 : 1
```
stephencelis, line 519: Limiting to accounts older than a week for now...could tweak this value, though.

mbrandonw (approved):
very nice, thanks!

Sources/Database/Database.swift
fernandadepaula: Jj

PR #697, opened 2021-03-05
Fix regional discount

stephencelis:
We were inaccurately displaying the subscription price for regional discounts with referrals.


mbrandonw (approved):
nice thanks!

PR #699, opened 2021-03-22
Small typo in #139’s solutions.

jasdev:
Needed the `minimumTolerance` parameter and to box `now` up into a closure. ✅


jasdev:
Ah, I guess we could rework the unconstrained `Scheduler.failing` to default to a `.zero` minimum tolerance, but then that’d diverge from the episode’s code.

stephencelis:
The episode code was wrong, sorry about that! Fixed now.

PR #703, opened 2021-04-05
Process new episode email blurb Markdown

stephencelis:


Tests/PointFreeTests/EmailTests/__Snapshots__/NewEpisodeEmailTests/testNewEpisodeEmail_Announcement_NonSubscriber.2.txt
```diff
@@ -8,8 +8,8 @@
 
 ### #2: Proof in Functions
 
- <http://localhost:8080/episodes/ep2-proof-in-functions>This is a short blurb to give a high-level overview of what the episode is about. It can only be plain
-text, no markdown allowed. Here is some more text just to have some filler.
+ <http://localhost:8080/episodes/ep2-proof-in-functions><p>This is a short blurb to give a high-level overview of what the episode is about. It can only be plain
```
stephencelis, line 11: Preprocessing makes the plain text emails a lil noisy, but we are already doing this for announcements anyway (see above fold). I think plain text views are pretty rare, though, so this should be fine.

mbrandonw (approved):

PR #704, opened 2021-04-26
Swift 5.3

stephencelis:
Now that 5.4 is officially out it's about time we get on...5.3? This PR was mainly motivated by Xcode 12.5 getting installed and compile time being just awful. If this merges/deploys cleanly can then try 5.4.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -48,8 +48,8 @@ var package = Package(
       name: "Ccmark",
       pkgConfig: "libcmark",
       providers: [
+        .apt(["cmark"]),
         .brew(["cmark"]),
-        .apt(["cmark"])
       ]
     ),
 
--- a/Package.swift
+++ b/Package.swift
@@ -1,53 +1,40 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.3
 
 import Foundation
 import PackageDescription
 
-#if swift(>=5.3)
-let filePath = #filePath
-#else
-let filePath = #file
-#endif
-
-let isOss = !FileManager.default.fileExists(
-  atPath: URL(fileURLWithPath: filePath)
-    .deletingLastPathComponent()
-    .appendingPathComponent("Sources")
-    .appendingPathComponent("Models")
-    .appendingPathComponent("Transcripts")
-    .appendingPathComponent(".git")
-    .path
-)
-
-extension SwiftSetting {
-  static let warnLongExpressionTypeChecking = unsafeFlags(
-    [
-      "-Xfrontend", "-warn-long-expression-type-checking=200",
-      "-Xfrontend", "-warn-long-function-bodies=200",
-    ],
-    .when(configuration: .debug)
-  )
-}
-
-extension Array where Element == SwiftSetting {
-  static let pointFreeSettings: Array = isOss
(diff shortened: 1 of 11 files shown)


Package.swift
```diff
@@ -1,53 +1,40 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.3
 
 import Foundation
 import PackageDescription
 
-#if swift(>=5.3)
```
stephencelis, line 6: Moved this work to the end.

Package.swift
```diff
@@ -1,53 +1,40 @@
...
+    .library(name: "ModelsTestSupport", targets: ["ModelsTestSupport"]),
+    .library(name: "PointFree", targets: ["PointFree"]),
+    .library(name: "PointFreePrelude", targets: ["PointFreePrelude"]),
+    .library(name: "PointFreeRouter", targets: ["PointFreeRouter"]),
+    .library(name: "PointFreeTestSupport", targets: ["PointFreeTestSupport"]),
+    .library(name: "Stripe", targets: ["Stripe"]),
+    .library(name: "StripeTestSupport", targets: ["StripeTestSupport"]),
+    .library(name: "Styleguide", targets: ["Styleguide"]),
+    .library(name: "Syndication", targets: ["Syndication"]),
+    .library(name: "Views", targets: ["Views"]),
```
stephencelis, line 33: Wanted to restore these schemes.

Package.swift
```diff
@@ -1,53 +1,40 @@
...
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
     .package(url: "https://github.com/ianpartridge/swift-backtrace.git", .exact("1.1.0")),
-    .package(url: "https://github.com/pointfreeco/Ccmark.git", .branch("main")),
```
stephencelis, line 50: System packages are deprecated because they can be defined as targets now, as it now is below.

Package.swift
```diff
@@ -57,6 +44,15 @@ let package = Package(
...
   targets: [
 
+    .systemLibrary(
+      name: "Ccmark",
+      pkgConfig: "libcmark",
+      providers: [
+        .apt(["cmark"]),
+        .brew(["cmark"]),
+      ]
+    ),
```
stephencelis, line 54: How we're supposed to define system libraries these days.

Package.swift
```diff
@@ -70,8 +66,7 @@ let package = Package(
         .product(name: "PostgreSQL", package: "PostgreSQL"),
         .product(name: "Prelude", package: "Prelude"),
         .product(name: "Tagged", package: "Tagged"),
-      ],
-      swiftSettings: .pointFreeSettings
```
stephencelis, line 74: Mutating these in at the very end now.

Sources/Views/Account/Account.swift
```diff
@@ -548,9 +548,9 @@ private func subscriptionTeammateOverview(_ data: AccountData) -> Node {
     enterpriseShareLink = []
   }
 
-  return .gridRow(
+  return Node.gridRow(
```
stephencelis, line 551: Without making these `Node`s explicit these expressions took over a minute to compile.

PR #706, opened 2021-05-04
Move layout code to Views module

stephencelis:
This makes it possible to render an entire page from the `Views` module, which means we can move or start writing some snapshot tests there, as well as render Xcode previews there.


Package.swift
```diff
@@ -31,6 +31,7 @@ var package = Package(
     .library(name: "Styleguide", targets: ["Styleguide"]),
     .library(name: "Syndication", targets: ["Syndication"]),
     .library(name: "Views", targets: ["Views"]),
+    .library(name: "WebPreview", targets: ["WebPreview"])
```
stephencelis, line 34: Added a new web preview module to render HTML in an Xcode preview.

mbrandonw (approved):

PR #708, opened 2021-06-14
Update 0149-DerivedBehavior.swift to remove repeated word in description

provBarryTolnas:
small typo "we will will build" should be "we will build"


mbrandonw (approved):
Ah, thanks @provBarryTolnas! Will deploy soon!

PR #720, opened 2021-09-08
Fix small typo in Safer Conciser Forms

AndreiVidrasco:


stephencelis (approved):
Ha good catch, thanks!

PR #722, opened 2021-09-20
Vimeo fixes

stephencelis:
Should further reduce bandwidth.

### DIFF
--- a/Sources/Models/Episode.swift
+++ b/Sources/Models/Episode.swift
@@ -364,13 +364,12 @@ public struct Episode: Equatable {
   }
   
   public struct Video: Codable, Equatable {
-    // TODO: Tagged<Bytes, Int>?
     public var bytesLength: Int
     public var vimeoId: Int
     public var vimeoSecret: String
 
-    public var downloadUrl: String {
-      "https://player.vimeo.com/external/\(self.vimeoId).hd.mp4?s=\(self.vimeoSecret)&profile_id=174&download=1"
+    public func downloadUrl(_ quality: Quality) -> String {
+      "https://player.vimeo.com/external/\(self.vimeoId).hd.mp4?s=\(self.vimeoSecret)&profile_id=\(quality.rawValue)&download=1"
     }
     public var streamingSource: String {
       "https://player.vimeo.com/video/\(self.vimeoId)?pip=1"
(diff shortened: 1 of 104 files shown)


Sources/PointFree/Account/PrivateRss.swift
```diff
@@ -288,7 +288,7 @@ private func item(forUser user: User, episode: Episode) -> RssItem {
     enclosure: .init(
       length: episode.fullVideo.bytesLength,
       type: "video/mp4",
-      url: episode.fullVideo.downloadUrl
+      url: episode.fullVideo.downloadUrl(.hd720)
```
stephencelis, line 291: We could potentially do 1080p for the latest episode...but maybe we wait till we figure out if our bandwidth gets in the acceptable range.

Sources/PointFree/Account/PrivateRss.swift
```diff
@@ -374,7 +374,7 @@ private func invalidatedItem(errorMessage: String) -> RssItem {
     enclosure: .init(
       length: Episode.ep0_introduction.fullVideo.bytesLength,
       type: "video/mp4",
-      url: Episode.ep0_introduction.fullVideo.downloadUrl
+      url: Episode.ep0_introduction.fullVideo.downloadUrl(.sd540)
```
stephencelis, line 377: For the invalid feed I'm just doing standard def.

Sources/PointFree/AtomFeed.swift
```diff
@@ -74,6 +74,7 @@ private func items() -> [RssItem] {
   return Current
     .episodes()
     .sorted(by: their({ $0.freeSince ?? $0.publishedAt }, >))
+    .prefix(4)
```
stephencelis, line 77: This will only put the latest 4 episodes in the public feed.

Sources/PointFree/AtomFeed.swift
```diff
@@ -130,7 +131,7 @@ can access your private podcast feed by visiting \(url(to: .account(.index))).
     return .init(
       length: video.bytesLength,
       type: "video/mp4",
-      url: video.downloadUrl
+      url: video.downloadUrl(.sd540)
```
stephencelis, line 134: Standard def for public feed.

Sources/Models/Episode.swift
```diff
@@ -364,30 +364,17 @@ public struct Episode: Equatable {
...
     }
     public var streamingSource: String {
       "https://player.vimeo.com/video/\(self.vimeoId)?pip=1"
     }
 
-    public init(
-      bytesLength: Int,
-      downloadUrl: String,
-      streamingSource: String
-    ) {
```
stephencelis, line 383: Got rid of this with a regex find and replace.

mbrandonw (approved):

PR #724, opened 2021-09-21
Prep for Swift 5.5

stephencelis:
Let's hold off till the official docker image is released. But I think otherwise we're good to go.


stephencelis:
5.5 just landed so should be good to merge! https://hub.docker.com/_/swift/

mbrandonw (approved):

PR #726, opened 2021-09-23
Update Poster Images

stephencelis:
Updates to use our own host as default.


Tests/PointFreeTests/PrivateRssTests.swift
```diff
@@ -11,24 +11,26 @@ import SnapshotTesting
...
 import XCTest
 
+private let episodes: [Episode] = [
+  .ep0_introduction,
+  .ep1_functions,
+  .ep2_sideEffects,
+  .ep3_uikitStylingWithFunctions,
+  .ep10_aTaleOfTwoFlatMaps,
+  .ep22_aTourOfPointFree,
+].map { update($0) { $0.image = "" } }
```
mbrandonw, line 21: instead of empty string would it be better to put a hard coded url just so that the snapshots render something more reasonable?

mbrandonw (approved):

PR #753, opened 2021-11-21
Email users when invalid user agent is detected.

mbrandonw:
We now email the user when we detect an invalid user agent to the private RSS feed.


Sources/PointFree/Account/Cancel.swift
```diff
@@ -221,7 +221,7 @@ let cancelEmailView = simpleEmailLayout(cancelEmailBodyView) <<< { owner, subscr
     Your \(subscription.plan.nickname) subscription has been canceled and will remain active through
     \(dateFormatter.string(from: subscription.currentPeriodEnd)).
     """,
-    template: .default,
+    template: .default(),
```
mbrandonw, line 224: The `.default` email template now takes a `Bool` parameter to determine if we want to show the PF logo or not. It seems kind of heavy handed in some types of emails, such as the one this PR adds. In fact, we may want to audit our existing emails and decide to omit the logo on some of them.

Sources/PointFree/Emails/EmailLayouts.swift
```diff
@@ -34,6 +36,8 @@ struct SimpleEmailLayoutData<A> {
 
 let emailStylesheet = styleguide
   <> a % key("border-bottom", "1px solid black")
+  <> p % lineHeight(1.5)
+  <> p % padding(bottom: .rem(0.75))
```
mbrandonw, line 40: this makes our emails read a bit nicer. will have to update a bunch of snaps.

Tests/PointFreeTests/AccountTests/__Snapshots__/UpdateProfileTests/testUpdateEmailSettings.email_settings_after_update.txt
```diff
@@ -2,4 +2,5 @@
   ▿ EmailSetting
     - newsletter: Newsletter.newEpisode
     ▿ userId: 00000000-0000-0000-0000-000000000001
-      - rawValue: 00000000-0000-0000-0000-000000000001
+      ▿ rawValue: 00000000-0000-0000-0000-000000000001
+        - uuid: "00000000-0000-0000-0000-000000000001"
```
mbrandonw, line 6: I'm not sure why these snapshots changed when I re-recorded everything. They are also the only failures on this branch. 🤔

PR #761, opened 2022-01-03
Runtime warnings blog post

mbrandonw:


Sources/Views/TranscriptBlockView.swift
```diff
@@ -52,7 +52,7 @@ public func transcriptBlockView(
         .target(.blank),
         .rel(.init(rawValue: "noopener noreferrer")),
       ],
-      .img(src: src, alt: "", attributes: [.class(imageClasses)])
+      .img(src: src, alt: block.content, attributes: [.class(imageClasses)])
```
mbrandonw, line 55: Seems like a good idea to use the block's content as the alt text of the image.

Sources/Views/TranscriptBlockView.swift
```diff
@@ -141,12 +141,10 @@ let innerVideoContainerClass: CssSelector =
   Class.size.height100pct
     | Class.size.width100pct
     | Class.position.absolute
-    | Class.pf.colors.bg.gray650
```
mbrandonw, line 144: dropping these gray backgrounds so that we can use transparent PNGs

stephencelis (approved):

PR #762, opened 2022-01-08
Confusing in the sample code of `SomeNewsAboutContramap`

ytyubox:
Thank you for share the idea. I think I am confused by the example.


stephencelis (approved):
Aha! Good catch, thanks!

PR #793, opened 2022-12-03
Update subscription plan (individual → personal) of Student Discounts page

Jager-yoo:
On [Announcing Student Discounts](https://www.pointfree.co/blog/posts/10-announcing-student-discounts) page, it is written that a discount code for `individual` subscription plan will be given, but on [Pricing](https://www.pointfree.co/pricing) page, there's no `individual` plan.

I don't know when, but the name of the plan seems to be changed from `individual` to `personal`.
To be precise, I update the words.


mbrandonw (approved):
Good catch! We did change the naming from "individual" to "personal" a while ago.

PR #802, opened 2022-12-22
Start converting database endpoints

stephencelis:
A bit of a rabbit hole in the final commit due to `async` and XCTest not playing nicely, especially around snapshots, but was able to get an `async` branch of SnapshotTesting in the works and things look a lot more stable 🤞


Sources/Database/Database.swift
```diff
@@ -198,6 +187,14 @@ public struct Client {
     self.upsertUser = upsertUser
   }
 
+  public func fetchSubscription(user: Models.User) async throws -> Models.Subscription {
+    do {
+      return try await self.fetchSubscriptionById(user.subscriptionId.unwrap())
+    } catch {
+      return try await self.fetchSubscriptionByOwnerId(user.id)
+    }
+  }
```
stephencelis, line 196: New helper to replace instances of `subscription <|> ownerSubscription`.

Sources/Database/Live.swift
```diff
@@ -30,10 +31,13 @@ extension Client {
           RETURNING *
           """
         )
-        .first(decoding: EnterpriseAccount.self)
+        .first()
+        .get()
+        .unwrap()
+        .decode(model: EnterpriseAccount.self, keyDecodingStrategy: .convertFromSnakeCase)
```
stephencelis, line 37: For some reason NIO Postgres doesn't have a way to configure this for the entire pool, so a lot of these lines are verbose.

We had `EitherIO` helpers that hid away the work and we could bring some similar helpers back in the future, but wanted to finish the conversion first.

Sources/PointFree/Account/Change.swift
```diff
@@ -133,16 +134,17 @@ private func fetchSeatsTaken<A>(
...
-        .map { $0.right?.count ?? 0 },
-    ])
+    let invitesAndTeammates = IO {
+      async let teamInvites = Current.database.fetchTeamInvites(user.id)
+        .performAsync()
+        .count
+      async let teammates = Current.database.fetchSubscriptionTeammatesByOwnerId(user.id)
+        .count
+      return ((try? await teamInvites) ?? 0) + ((try? await teammates) ?? 0)
+    }
```
stephencelis, line 144: Fun lil conversion.

Sources/PointFree/SubscribeConfirmation.swift
```diff
@@ -220,39 +219,27 @@ func redirectActiveSubscribers<A>(
...
-        ?? throwE(unit)
-
-      let ownerSubscription =
-        (user?.id)
-        .map { Current.database.fetchSubscriptionByOwnerId($0).mapExcept(requireSome) }
-        ?? throwE(unit)
-
-      let race = (userSubscription.run.parallel <|> ownerSubscription.run.parallel).sequential
-
-      return EitherIO(run: race)
```
stephencelis, line 235: I removed this race for a simpler cascade, which is more deterministic. We actually have a test that fails about 7% of the time because of this non-determinism. Prob don't need to race our own DB anyway.

Sources/PointFreePrelude/Either.swift
```diff
@@ -19,6 +19,15 @@ public func requireSome<A>(_ e: Either<Error, A?>) -> Either<Error, A> {
...
 }
 
+extension Optional {
+  struct Nil: Error {}
+
+  public func unwrap() throws -> Wrapped {
+    guard let unwrapped = self else { throw Nil() }
+    return unwrapped
+  }
+}
```
stephencelis, line 29: Helper for unwrapping optionals in throwing contexts. Feels missing from the standard library.

Tests/FunctionalCssTests/FunctionalCssTests.swift
```diff
@@ -5,14 +5,15 @@ import PointFreeTestSupport
 import SnapshotTesting
 import XCTest
 
+@MainActor
 class FunctionalCssTests: TestCase {
-  override func setUp() {
-    super.setUp()
-    //    SnapshotTesting.record = true
+  override func setUp() async throws {
+    try await super.setUp()
```
stephencelis, line 11: Updated all tests to be `@MainActor` and `async throws`:

- XCTest basically requires things to be on the main thread, but async tests throw a wrench in this
- XCTest on Linux fails to build tests when async tests are mixed with non-async tests.

stephencelis:
A consequence of async snapshots is that it waits long enough for images to load, so some image snapshots are more consistent now.

mbrandonw (approved):

PR #805, opened 2022-12-23
Update Mailgun.Client to use async/await.

mbrandonw:


Sources/PointFree/Account/Team.swift
```diff
@@ -101,27 +99,22 @@ private let requireTeammate: MT<Tuple2<User.ID, User>, Tuple2<User, User>> = fil
...
-  }
+  async let response1 = sendEmail(
+    to: [teammate.email],
+    subject: "You have been removed from \(owner.displayName)’s Point-Free team",
+    content: inj2(youHaveBeenRemovedEmailView(.teamOwner(owner)))
+  )
 
-  return zip2(
-    parallel(
-      sendEmail(
```
mbrandonw, line 112: replacing a fire-and-forgot-zip-parallel with a few `async let`s.

Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -60,15 +60,13 @@ private func updateProfileMiddlewareHandler(
   let updateFlash: Middleware<HeadersOpen, HeadersOpen, Prelude.Unit, Prelude.Unit>
   if data.email.rawValue.lowercased() != user.email.rawValue.lowercased() {
     updateFlash = flash(.warning, "We've sent an email to \(user.email) to confirm this change.")
-    parallel(
-      sendEmail(
+    Task {
+      _ = try await sendEmail(
```
mbrandonw, line 64: using `Task { ... }` for fire-and-forget in a few spots.

Sources/PointFree/Account/Team.swift
```diff
@@ -83,9 +83,7 @@ let removeTeammateMiddleware =
       try await Current.database
         .removeTeammateUserIdFromSubscriptionId(teammate.id, teammateSubscriptionId)
 
-      // Fire-and-forget emails to owner and teammate
-      sendEmailsForTeammateRemoval(owner: currentUser, teammate: teammate)
-        .run({ _ in })
+      try await sendEmailsForTeammateRemoval(owner: currentUser, teammate: teammate)
```
stephencelis, line 86: Should this be wrapped in a `Task`?

Sources/PointFree/Account/Team.swift
stephencelis: I think the one difference here is that `async let`s will still throw and cause the request to fail. That might be fine, but if we wanna stay on the happy path for the request maybe we should use `Task` here as well?

stephencelis (approved):
Nice! Just a coupla small questions.

Sources/PointFree/Account/Team.swift
mbrandonw: yep good catch. updating now.

Sources/PointFree/Account/Team.swift
mbrandonw: yeah, I decided to just inline it all: https://github.com/pointfreeco/pointfreeco/pull/805/commits/95319eba064fdc4abe106583e1b5bbcb9ccff968

PR #809, opened 2023-01-02
Use dependencies

mbrandonw:
Basics of converting Environment to the new dependencies management system.


Sources/Models/BlogPost.swift
```diff
@@ -45,3 +46,67 @@ public struct BlogPost: Equatable, Identifiable {
     case stephen
   }
 }
+
+extension BlogPost: TestDependencyKey {
+  public static var testValue = {
+    [
```
mbrandonw, line 52: I'm using the `BlogPost` type as a place to express the `blogPosts` dependency, but if that feels weird I can just define a little private enum key.

Sources/Models/BlogPosts/BlogPost0020_PodcastRSS.swift
```diff
@@ -42,7 +42,7 @@ public let post0020_PodcastRSS = BlogPost(
         Since the time we launched this feature we have changed this slightly. Only yearly subscribers get access to the entire catalogue of episodes in their podcast RSS feed, and monthly subscribers get access to the last 4 episodes.
         """,
       timestamp: nil,
-      type: .correction
+      type: .box(.correction)
```
mbrandonw, line 45: I generalized how we can do little informational boxes in transcripts/blog posts. can do now do things like correction, tip, info, warn, etc...

Sources/Models/BlogPosts/BlogPost0091_EOYSaleLastDay.swift
```diff
@@ -59,6 +59,6 @@ public let post0091_2022EOYSaleLastDay = BlogPost(
   coverImage: nil,
   hidden: true,
   id: 91,
-  publishedAt: referenceDateFormatter.date(from: "2022-12-31")!,
+  publishedAt: yearMonthDayFormatter.date(from: "2022-12-31")!,
```
mbrandonw, line 62: also started using the formatter in more places so I renamed it.

mbrandonw:
I have a branch going on the private transcripts repo that we will need to merge when we merge this.

Sources/PointFree/Account/Cancel.swift
```diff
@@ -34,7 +37,9 @@ private func cancelResponse(
     }
     _ = try await Current.stripe
       .cancelSubscription(stripeSubscription.id, stripeSubscription.status == .pastDue)
-    Task { try await sendCancelEmail(to: user, for: stripeSubscription) }
+    await fireAndForget {
+      try await sendCancelEmail(to: user, for: stripeSubscription)
+    }
```
mbrandonw, line 42: Here's an example of using the `fireAndForget` dependency. This makes it easy to test sending the cancellation email, which I added test coverage for.

Sources/PointFree/Account/UpdateProfile.swift
```diff
@@ -58,13 +62,6 @@ private func updateProfileMiddlewareHandler(
   let updateFlash: Middleware<HeadersOpen, HeadersOpen, Prelude.Unit, Prelude.Unit>
   if data.email.rawValue.lowercased() != user.email.rawValue.lowercased() {
     updateFlash = flash(.warning, "We've sent an email to \(user.email) to confirm this change.")
-    Task {
-      _ = try await sendEmail(
-        to: [user.email],
-        subject: "Email change confirmation",
-        content: inj2(confirmEmailChangeEmailView((user, data.email, emailChangePayload)))
-      )
-    }
```
mbrandonw, line 67: I moved this down to the async IO below, _and_ made it so that it's sent only if the `Current.database.updateUser` succeeds, which is probably more correct.

Sources/PointFree/Bootstrap.swift
```diff
@@ -9,119 +9,22 @@ import PointFreePrelude
 import PointFreeRouter
 import PostgresKit
 import Prelude
-
-public func bootstrap(eventLoopGroup: EventLoopGroup) -> EitherIO<Error, Prelude.Unit> {
+ 
+public func bootstrap() -> EitherIO<Error, Prelude.Unit> {
```
mbrandonw, line 13: pretty much all of bootstrap goes away since most of this work is in preparing dependencies, and that work just happens in each respective `liveValue`.

mbrandonw: maybe we don't even need this anymore...

Sources/PointFree/Environment.swift
```diff
@@ -1,69 +1,154 @@
...
-    self.renderHtml = renderHtml
-    self.renderXml = renderXml
-    self.stripe = stripe
-    self.uuid = uuid
+import Stripe 
+
+// NB: Deprecate remove soon: @available(*, deprecated)
+public var Current: DependencyValues {
+  DependencyValues._current
+}
```
mbrandonw, line 17: Can deprecate/delete this in a later PR to minimize churn.

Sources/PointFree/Environment.swift
```diff
@@ -1,69 +1,168 @@
...
+  public var mainEventLoopGroup: MultiThreadedEventLoopGroup {
+    get { self[MainEventLoopGroupKey.self] }
+    set { self[MainEventLoopGroupKey.self] = newValue }
+  }
+}
+
+extension Database.Client: DependencyKey {
+  public static var liveValue: Self {
+    @Dependency(\.envVars) var envVars
+    @Dependency(\.mainEventLoopGroup) var mainEventLoopGroup
```
mbrandonw, line 102: example of using `@Dependency` from within the `liveValue` of another dependency. Seems to work well AFAICT.

Sources/PointFree/Session.swift
```diff
@@ -13,6 +14,18 @@ public enum CookieTransform: String, Codable {
   case encrypted
 }
 
+extension CookieTransform: DependencyKey {
+  public static let liveValue = Self.encrypted
+  public static let testValue = Self.plaintext
```
mbrandonw, line 19: fun example of a dependency that doesn't need an "unimplemented" test value. best to just use plaintext cookies.

Sources/PointFreeTestSupport/TestCase.swift
```diff
@@ -16,13 +17,24 @@ open class TestCase: XCTestCase {
...
+      $0.database = .mock
+      $0.date.now = .mock
+      $0.envVars = $0.envVars.assigningValuesFrom(ProcessInfo.processInfo.environment)
+      $0.gitHub = .mock
+      $0.mailgun = .mock
+      $0.stripe = .mock
+      $0.uuid = .incrementing
+    } operation: {
+      super.invokeTest()
+    }
```
mbrandonw, line 31: This is essentially what we are doing on main, but I think eventually we would like to remove all these and be more explicit in tests that are actually using one of these dependencies.

Tests/PointFreeTests/AccountTests/AccountTests.swift
```diff
@@ -73,13 +74,13 @@ final class AccountTests: TestCase {
   }
 
   func testAccount() async throws {
-    Current = .teamYearly
+    await withDependencyValues {
+      $0.teamYearly()
+    } operation: {
```
mbrandonw, line 79: Pretty much all test changes should fall into this simple pattern: opening up `withDependencyValues`, moving `Current` mutation inside the first closure, changing to `$0`, and then moving assertions to the `operation` closure.

Tests/PointFreeTests/AccountTests/CancelTests.swift
```diff
@@ -25,34 +27,43 @@ final class CancelTests: TestCase {
...
+      let cancelSubscription = Current.stripe.cancelSubscription
+      $0.stripe.cancelSubscription = {
+        immediately = $1
+        return try await cancelSubscription($0, $1)
+      }
+      $0.mailgun.sendEmail = { email in
+        expectation.fulfill()
+        XCTAssertEqual(email.subject, "[testing] Your subscription has been canceled")
+        return SendEmailResponse(id: "mail-id", message: "")
+      }
```
mbrandonw, line 42: Here's an example of how I was able to test that the cancellation email is sent when cancelling a subscription.

Tests/PointFreeTests/__Snapshots__/ApiTests/testEpisode_NotFound.1.Conn.txt
```diff
@@ -11,7 +11,18 @@ X-Frame-Options: SAMEORIGIN
...
...5em}img{border-style:none}button,input,optgroup,select,textarea{font-family:inherit;font-size:100%;line-height:1.15;margin:0}button,input{overflow:visible}button,select{text-transform:none}[type=button],[type=reset],[type=submit],button{-webkit-appearance:button}[type=button]::-moz-focus-inner,[type=reset]::-moz-focus-inner,[type=submit]::-moz-focus-inner,button::-moz-focus-inner{border-style:none;padding:0}[type=button]:-moz-focusring,[type=reset]:-moz-focusring,[type=submit]:-moz-focusring,button:-moz-focusring{outline:1px dotted ButtonText}fieldset{padding:.35em .75em .625em}legend{box-sizing:border-box;color:inherit;display:table;max-width:100%;padding:0;white-space:normal}progress{vertical-align:baseline}textarea{overflow:auto}[type=checkbox],[type=radio]{box-sizing:border-box;padding:0}[type=number]::-webkit-inner-spin-button,[type=number]::-webkit-outer-spin-button{height:auto}[type=search]{-webkit-appearance:textfield;outline-offset:-2px}[type=search]::-webkit-search-decoration{-webkit-appearance:none}::-webkit-file-upload-button{-webkit-appearance:button;font:inherit}details{display:block}summary{display:list-item}template{display:none}[hidden]{display:none}</style><style>
```
mbrandonw, line 14: this test was previously accidentally rendering in minified form, but we now have a better way of making sure that tests always render in pretty format.

Tests/PointFreeTests/__Snapshots__/BlogTests/testBlogIndex.1.Conn.txt
```diff
@@ -2192,9 +2192,7 @@ Prism.languages.insertBefore('swift', 'other-directive', {
               </div>
               <div>
                 <div class="md-ctn">
-                  <p>This is the main content of the blog post. Each paragraph can use markdown, but titles code snippets
-should be broken out into separate content blocks so that we can use the JS syntax highlighting
-library. For example, here is some code:</p>
+                  <p>This is the main content of the blog post. Each paragraph can use markdown, but titles code snippets should be broken out into separate content blocks so that we can use the JS syntax highlighting library. For example, here is some code:</p>
```
mbrandonw, line 2195: this is just me cleaning up some incorrect new lines (and same for some changes below)

Sources/PointFree/Environment.swift
john-flanagan: What happens if you create a circular dependency between `liveValue`s?

john-flanagan: Makes sense. That's about what I figured. I'm excited to see the implementation of `swift-dependencies`. It looks like a great tool from what I see in the various public `swift-deps` branches 😄

stephencelis (approved):
Looks good!

PR #810, opened 2023-01-07
Add NIO HTTP client

stephencelis:
Since moving to async we've noticed some flakiness with URLSession on Linux. Let's swap out an endpoint for NIO HTTP and see if it improves.


Package.swift
```diff
@@ -171,12 +179,21 @@ var package = Package(
       ]
     ),
 
+    .target(
+      name: "LoggingDependencies",
```
stephencelis, line 183: Moving some core dependencies, like `Logger`, higher up so they can be used outside the `PointFree` module.

Package.swift
```diff
@@ -226,6 +243,15 @@ var package = Package(
       ]
     ),
 
+    .target(
+      name: "NIODependencies",
```
stephencelis, line 247: Same goes for `EventLoopGroup`.

Package.swift
```diff
@@ -279,6 +306,20 @@ var package = Package(
       ]
     ),
 
+    .target(
+      name: "PointFreePrelude",
```
stephencelis, line 310: (Alphabetized.)

Sources/FoundationPrelude/AsyncHTTPClient.swift
```diff
@@ -0,0 +1,57 @@
+import AsyncHTTPClient
```
stephencelis, line 1: Technically this dependency is pretty heavyweight, since it compiles all of NIO. Could experiment extracting it to a "live" dependency module in the future...

Sources/FoundationPrelude/AsyncHTTPClient.swift
```diff
@@ -0,0 +1,57 @@
...
+
+public func dataTask(
+  with request: HTTPClientRequest
+) async throws -> (ByteBuffer, HTTPClientResponse) {
+  @Dependency(\.httpClient) var client
+  @Dependency(\.logger) var logger
+  let response = try await client
+    .execute(request, timeout: .seconds(Int64(timeoutInterval)), logger: logger)
+  let contentLength = response.headers.first(name: "content-length").flatMap(Int.init)
+    ?? 1024 * 1024  // 1MB
```
stephencelis, line 54: The library is kinda funny...it seems you must specify a limit to the response size. 1MB seems plenty, though...

Sources/NIODependencies/NIO.swift
```diff
@@ -0,0 +1,16 @@
...
+extension DependencyValues {
+  public var mainEventLoopGroup: any EventLoopGroup {
+    get { self[MainEventLoopGroupKey.self] }
+    set { self[MainEventLoopGroupKey.self] = newValue }
+  }
+}
+
+public enum MainEventLoopGroupKey: TestDependencyKey {
+  public static var testValue: any EventLoopGroup {
+    EmbeddedEventLoop()
```
stephencelis, line 14: NIOCore does not contain the live event loop implementations. To get those we need to compile all of NIO. There's a lightweight `NIOEmbedded` module, though, that we can use in tests, instead.

If we're worried about losing the "integration" of using a live loop, we either need to depend on all of NIO here, or we need to make sure all our tests `withDependencies` a live loop in...

Tests/GitHubTests/GitHubTests.swift
```diff
@@ -1,4 +1,6 @@
+@testable import AsyncHTTPClient
```
stephencelis, line 1: Apparently HTTPClient requests are completely opaque, so in order to snapshot test them I'm doing a testable import to get access to the underlying value.

We might just want to fully abstract around the idea of an HTTP client in the future to avoid this. More work and kinda annoying, but it seems like this library isn't currently built with testability/modularity in mind.

Sources/FoundationPrelude/AsyncHTTPClient.swift
mbrandonw: Maybe we should up it just a bit. Some Stripe responses are pretty big.

mbrandonw (approved):

Sources/FoundationPrelude/AsyncHTTPClient.swift
```diff
@@ -0,0 +1,58 @@
...
```
stephencelis, line 55:
```suggestion
    ?? 5 * 1024 * 1024  // 5MB
```

PR #812, opened 2023-01-08
Use @Dependency for current user, route and subscription

mbrandonw:
The current user, route and subscription are ubiquitous in the site and we go through a lot of trouble to pass that data around everywhere. So, let's try moving it to `@Dependency` to more easily propagate it. Doing that actually fixed a number of small _quirks_ in the site due to us being too lazy to explicitly pass everything (e.g. some pages act like your logged out even if you are logged in because we pass `user = nil` to the middleware).


Package.swift
```diff
@@ -261,6 +263,15 @@ var package = Package(
...
     ),
 
+    .target(
+      name: "PointFreeDependencies",
+      dependencies: [
+        "Models",
+        "PointFreeRouter",
+        .product(name: "Dependencies", package: "swift-dependencies"),
+      ]
+    ),
```
mbrandonw, line 273: I'm putting all the dependency keys in here for now, but we could split into more modules later if we want. Unfortunately we have to depend on `PointFreeRouter`, which is pretty heavy, because the site's routes _and_ router all live in there. We should probably break the routes out into its own module.

Sources/Models/SubscriberState.swift
```diff
@@ -9,10 +9,12 @@ public enum SubscriberState {
...
   public init(
-    user: User?, subscriptionAndEnterpriseAccount: (Models.Subscription, EnterpriseAccount?)?
+    user: User?,
+    subscription: Models.Subscription?,
+    enterpriseAccount: EnterpriseAccount?
   ) {
-    switch (user, subscriptionAndEnterpriseAccount) {
-    case let (.some(user), .some((subscription, enterpriseAccount))):
+    switch (user, subscription) {
+    case let (.some(user), .some(subscription)):
```
mbrandonw, line 17: We were using a weird optional tuple for an argument to the init due to how the middleware transformers aggregated all the data together. I've greatly simplified that and so this can now be simpler.

Sources/PointFree/About.swift
```diff
@@ -8,15 +8,12 @@ import Tuple
 import Views
 
 func aboutResponse(
-  _ conn: Conn<StatusLineOpen, (User?, SubscriberState, SiteRoute?)>
+  _ conn: Conn<StatusLineOpen, Void>
```
mbrandonw, line 11: lots of middlewares were able to drop most/all of their data.

Sources/PointFree/Account/AccountMiddleware.swift
```diff
@@ -1,43 +1,47 @@
...
+  let account = conn.data
 
   switch account {
   case let .confirmEmailChange(payload):
     return conn.map(const(payload))
       |> confirmEmailChangeMiddleware
 
   case .index:
-    return conn.map(const(user .*. subscriberState .*. unit))
+    return conn.map(const(currentUser .*. subscriberState .*. unit))
```
mbrandonw, line 28: There are still a few spots I thread the current user through explicitly, such as when need to require a present `User` and redirect when `nil`, although I did update some spots.

As we push more to get rid of `IO` and `EitherIO` I'm wondering if there could be a throwing computed property `loggedInUser: User`, and depending on what error is throw we could possibly catch the error higher up and perform the login+redirect there. Something to discuss soon.

Sources/PointFree/Admin/AdminMiddleware.swift
```diff
@@ -1,65 +1,66 @@
...
-{
+  guard let currentUser = currentUser
+  else { return loginAndRedirect(conn) }
 
-  let (user, route) = lower(conn.data)
+  guard currentUser.isAdmin
+  else {
+    return conn
+      |> redirect(to: .home, headersMiddleware: flash(.error, "You don't have access to that."))
+  }
```
mbrandonw, line 21: Here's an example where I unrolled some middleware transformers and just performed the logic inline. Got rid of `requireAdmin` and instead just unwrap the user and make sure their an admin.

mbrandonw: I guess `admin: User` could also be a throwing property on `DependencyValues` that throws a different kind of error, which is caught higher up and does the redirect+flash.

Sources/PointFree/Admin/FreeEpisodes.swift
```diff
@@ -24,12 +24,15 @@ let sendFreeEpisodeEmailMiddleware:
...
-    Tuple2<User, Episode.ID>,
+    Episode.ID,
     Data
-  > =
-    filterMap(get2 >>> fetchEpisode >>> pure, or: redirect(to: .admin(.freeEpisodeEmail())))
-    <| sendFreeEpisodeEmails
-    >=> redirect(to: .admin())
+> = { conn in
+  guard let episode = fetchEpisode(conn.data)
+  else { return conn |> redirect(to: .admin(.freeEpisodeEmail())) }
```
mbrandonw, line 31: More manual unrolling of middleware transformers. No more `filterMap`.

Sources/PointFree/Auth.swift
```diff
@@ -56,47 +57,19 @@ public func loginAndRedirect<A>(_ conn: Conn<StatusLineOpen, A>) -> IO<Conn<Resp
   conn |> redirect(to: .login(redirect: conn.request.url?.absoluteString))
 }
 
-public func currentUserMiddleware<A>(_ conn: Conn<StatusLineOpen, A>)
```
mbrandonw, line 59: I inlined this in `siteMiddleware`, so no more need for it.

Sources/PointFree/Auth.swift
```diff
@@ -56,47 +57,19 @@ public func loginAndRedirect<A>(_ conn: Conn<StatusLineOpen, A>) -> IO<Conn<Resp
...
 
   return { conn in
-    return conn.map(const(conn.data.second))
-      |> (get1(conn.data) == nil
-        ? middleware
-        : redirect(to: .account(), headersMiddleware: flash(.warning, "You’re already logged in.")))
-  }
-}
-
-public func currentSubscriptionMiddleware<A, I>(
```
mbrandonw, line 83: Same for this.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -11,103 +12,142 @@ import Styleguide
...
 import Views
 
-public let siteMiddleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
-  requestLogger(logger: { Current.logger.log(.info, "\($0)") }, uuid: UUID.init)
-  <<< requireHerokuHttps(allowedInsecureHosts: allowedInsecureHosts)
-  <<< redirectUnrelatedHosts(isAllowedHost: { isAllowed(host: $0) }, canonicalHost: canonicalHost)
-  <<< router(notFound: routeNotFoundMiddleware)
-  <| currentUserMiddleware
-  >=> currentSubscriptionMiddleware
-  >=> render(conn:)
```
mbrandonw, line 21: I had a hard time wrapping the entire app stack in a `withDependency` for setting up the current user, route, etc. because it needed to be around both `render(conn:)` _and_ `route(notFound:)` (in order for 404's to show correct logged in state).

So, I unrolled all of this into a single async function. In some ways it's pretty nice, but once we get back to our server work I do think having a stack of middlewares to compose is nice.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -11,103 +12,142 @@ import Styleguide
...
-{
+public let siteMiddleware = { conn in
+  IO { await _siteMiddleware(conn) }
+}
 
-  return { middleware in
-    @Dependency(\.siteRouter) var siteRouter
+private func _siteMiddleware(
+  _ conn: Conn<StatusLineOpen, Prelude.Unit>
+) async -> Conn<ResponseEnded, Data> {
```
mbrandonw, line 21: And here's the unrolled middleware.

Tests/PointFreeTests/AccountTests/__Snapshots__/AccountTests/testAccount.1.Conn.txt
```diff
@@ -1896,6 +1896,10 @@ html {
           content="summary_large_image">
     <meta name="twitter:site"
           content="@pointfreeco">
+    <meta property="og:url"
+          content="http://localhost:8080/account">
+    <meta name="twitter:url"
+          content="http://localhost:8080/account">
```
mbrandonw, line 1902: Now that we are more easily threading current route throughout the app, there are a bunch of snapshots that update because they can now set meta tags.

Tests/PointFreeTests/InviteTests.swift
```diff
@@ -362,9 +362,14 @@ class InviteTests: TestCase {
       UUID(uuidString: "deadbeef-dead-beef-dead-beefdead0001")!)
 
     await withDependencies {
+      $0 = .test
```
mbrandonw, line 365: for a few tests I've started being more strict by reseting everything back to test values. Currently we mock quite a few things in the base test case, but I think eventually we want to get rid of all of that and start being very explicit in tests. By doing this I was able to track down quite a few subtle things in our tests.

PR #815, opened 2023-01-10
Async siteMiddleware

mbrandonw:
This makes `siteMiddleware` fully async instead of returning an `IO`. This is a big change because all tests need to be updated to await directly instead of going through `IO`, but it opens us up to writing fully async site endpoints with no mention of IO.

### DIFF
--- a/Tests/PointFreeTests/AccountTests/AccountTests.swift
+++ b/Tests/PointFreeTests/AccountTests/AccountTests.swift
@@ -58,7 +58,7 @@ final class AccountIntegrationTests: LiveDatabaseTestCase {
 
     let conn = connection(from: request(to: .team(.leave), session: .loggedIn(as: currentUser)))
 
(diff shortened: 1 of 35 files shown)


Tests/PointFreeTests/AccountTests/AccountTests.swift
```diff
@@ -58,7 +58,7 @@ final class AccountIntegrationTests: LiveDatabaseTestCase {
 
     let conn = connection(from: request(to: .team(.leave), session: .loggedIn(as: currentUser)))
 
-    await assertSnapshot(matching: conn |> siteMiddleware, as: .ioConn)
+    await assertSnapshot(matching: await siteMiddleware(conn), as: .conn)
```
stephencelis, line 61: Wanna find/replace these extra `await`s?

```suggestion
    await assertSnapshot(matching: siteMiddleware(conn), as: .conn)
```

stephencelis (approved):

PR #816, opened 2023-01-11
Revert the reverts

mbrandonw:
We reverted #815 due to a deadlock, but we've now found a workaround (PR forthcoming).


mbrandonw:
Fix here #817.

PR #817, opened 2023-01-11
Trick to properly start up server

mbrandonw:
We were seeing some very strange behavior with starting up our server once it started making fuller use of async in server endpoints. It seemed like possibly a bug in the Swift language. But then we found [this](https://github.com/apple/swift-nio/blob/1ce136b4c392bd7427cf408d65305a05d162fb46/Sources/NIOAsyncAwaitDemo/main.swift#L66-L76) trick in a NIO sample, and it worked for us. 🤨


Sources/Server/main.swift
```diff
@@ -0,0 +1,36 @@
...
+import PointFree
+import Prelude
+import Dispatch
+
+// NB: this DispachGroup/Task is necessary to work around strange deadlock issues in Swift async.
+//     For more info, see:
+//     https://github.com/apple/swift-nio/blob/1ce136b4c392bd7427cf408d65305a05d162fb46/Sources/NIOAsyncAwaitDemo/main.swift#L66-L76.
+let group = DispatchGroup()
+group.enter()
+Task {
```
mbrandonw, line 13: @weissi Sorry to ping you out of the blue, but we found this trick from your PR https://github.com/apple/swift-nio/pull/1701, but there wasn't much info about why it's needed or why it works.

Without this we were seeing very strange behavior in processing requests. A request would come into the server, and our `siteMiddleware` would finish processing it (even the `defer` would execute inside), yet somehow the function was never fully exited. It would just hang forever. It honestly seemed like a Swift language bug.

This `DispatchGroup`/`Task` trick fixed it for us. Is it safe to use in production?

weissi: These days you can use async main functions and that's best.

weissi: If you can't do that, then the best workaround is probably something along the lines of

```
Task {
    try await stuff()
    DispatchQueue.main.async {
        exit(0)
    }
}

RunLoop.main.run()
```

mbrandonw: > These days you can use async main functions and that's best.

Yeah, that's exactly how we started, but we had very weird issues with it. This task group trick was the thing that fixed it.

But now I've just found another workaround to fix it, which is to pass `siteMiddleware` directly to our `run` helper instead of expanding it into a closure:

```diff
// Server
 run(
-  { await siteMiddleware($0) },
+  siteMiddleware,
   on: envVars.port,
```

This has got to be a Swift concurrency bug.

mbrandonw: Thanks for the info @weissi. We're going to try to get a minimal repro to submit a bug report.

PR #818, opened 2023-01-12
Show when episode is finished

mbrandonw:
This is an MVP of showing when an episode has been finished on the homepage. We can do more in the future, but this covers the biggest use case.

Once this is deployed we can run the following SQL to backfill the new `is_finished` field in the DB:

```sql
UPDATE
  episode_progresses
SET
  is_finished = "percent" >= 90
```


Sources/Database/Live.swift
```diff
@@ -778,6 +785,24 @@ extension Client {
...
+        try await database.run(
+          """
+          DROP INDEX IF EXISTS "index_episode_progresses_on_episode_sequence_user_id"
+          """
+        )
+        try await database.run(
+          """
+          CREATE UNIQUE INDEX IF NOT EXISTS "index_episode_progresses_on_user_id_episode_sequence"
+          ON "episode_progresses" ("user_id", "episode_sequence")
+          """
```
mbrandonw, line 804: Re-indexing `episode_progresses` so that `user_id` is first to be more efficient.

Sources/Views/Home.swift
```diff
@@ -167,39 +167,103 @@ private func episodesListView(episodes: ArraySlice<Episode>, emergencyMode: Bool
...
+  @Dependency(\.siteRouter) var siteRouter;
+
+  let brightness: Stylesheet
+  let watched: Node
+  if episodeProgresses[episode.sequence]?.isFinished == true {
+    if episode.sequence.rawValue.quotientAndRemainder(dividingBy: 4).remainder == 3 {
+      brightness = key("filter", "grayscale(1) brightness(175%)")
+    } else {
+      brightness = key("filter", "grayscale(1)")
+    }
```
mbrandonw, line 198: I have to do this funny thing to bump the brightness of every 4th image because the purple posters tend to be a little darker than the others.

Ideally we will separate the tint from all of our posters and apply the tints in CSS, but that's a lot of work.

Sources/PointFree/SiteMiddleware.swift
```diff
@@ -82,8 +84,12 @@ public func siteMiddleware(
 
   return await withDependencies {
     $0.currentUser = currentUser
-    $0.requestID = requestID
     $0.currentRoute = siteRoute ?? .home
+    $0.episodeProgresses = .init(
+      progresses.map { ($0.episodeSequence, $0) },
+      uniquingKeysWith: { $1 }
+    )
+    $0.requestID = requestID
```
mbrandonw, line 92: Since this actually only ended up being used on the homepage, I could be convinced to pull this out of dependencies and just thread it through to the view. But, in the future I'm sure we will want to show this info in more places, and at that time we will probably want to do this. Just a thought.

Sources/Database/Live.swift
stephencelis: Would be cool to have a schema version sometime to scope these migrations, but prob would wanna revisit databases from scratch first 😅

Sources/Models/EpisodeProgress.swift
```diff
@@ -0,0 +1,22 @@
...
+    percent: Int,
+    userID: User.ID
+  ) {
+    self.episodeSequence = episodeSequence
+    self.isFinished = isFinished
+    self.percent = percent
+    self.userID = userID
+  }
+
+  public var id: Episode.Sequence {
```
stephencelis, line 19: Dunno if we wanna tag this to the progress or not to avoid strange comparisons, or if that's too much to bother, ha. Maybe nested tagged values are also weird?

mbrandonw: Actually not sure why I did it this way. The `episode_progresses` table has a primary key, so I should just use that. Will update.

stephencelis (approved):
Nice!

PR #833, opened 2023-03-01
Move public episodes to markdown files

mbrandonw:
This is a small step towards getting all episode transcripts converted to plain markdown files instead of Swift code. It's just the public episodes for now, which is about 20% of all of our episodes, and already I've noticed a decent reduction in compile times (about 10-15 seconds). Here's the code changes:

> 36,207 additions and 53,114 deletions.


Package.swift
```diff
@@ -566,3 +572,15 @@ for index in package.targets.indices {
...
+  let transcriptsDirectoryPath = URL(fileURLWithPath: #filePath)
+    .deletingLastPathComponent()
+    .appendingPathComponent("Sources")
+    .appendingPathComponent("Transcripts")
+    .appendingPathComponent("Resources")
+    .path
+
+  return try! FileManager.default.contentsOfDirectory(atPath: transcriptsDirectoryPath)
+    .map { .copy("Resources/\($0)") }
+}
```
mbrandonw, line 586: I was able to copy over just the markdown files that are in the `Resources` directory.

Sources/Transcripts/Episodes/0000-Introduction.swift
```diff
@@ -32,85 +32,6 @@ extension Episode {
       ),
       vimeoId: 354_215_017
     ),
-    transcriptBlocks: _privateTranscriptBlocks
+    transcriptBlocks: loadTranscriptBlocks(forSequence: 0)
```
mbrandonw, line 35: this is my hack for right now to load transcripts from a resource. I think eventually we will want to lazily load these on first access.

stephencelis (approved):

blueberrychopsticks:
Can paid members get access to private episode transcripts?

I'd love to throw together a TCA chatbot to help answer questions and direct devs towards relevant resources.

Let me know! cc @stephencelis @mbrandonw

mbrandonw:
Hi @blueberrychopsticks, sorry for the delayed response. We don't have any plans for that for right now, but we will keep it in mind.

PR #837, opened 2023-04-03
Swift 5.8 Support

stephencelis:
The Docker image is out now so we can get things up-to-date.


.github/workflows/ci.yml
```diff
@@ -14,24 +14,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
-    name: MacOS
-    runs-on: macos-12
```
stephencelis, line 19: Swift 5.8 is macOS 13-only, and GitHub Actions doesn't support that yet. Given that our main requirement is Linux, though, maybe that's OK?

mbrandonw (approved):
awesome!

PR #840, opened 2023-04-06
New admin tool for previewing emails

mbrandonw:
I wanted to give a slight refresh to some of our welcome emails, but realized it's a pain to preview the changes, so I built a little admin tool for it. I made some small updates to the welcome emails, but I want to do more soon now that we have this tool.


stephencelis (approved):
Looks good! We should also revisit the `WebPreview` module to see what it takes to get SwiftUI previews of pages working.

PR #847, opened 2023-05-31
Don't persist incomplete subscriptions

stephencelis:


Sources/PointFree/Subscribe.swift
```diff
@@ -116,7 +114,23 @@ private func subscribe(
           "requiresAction": true,
         ]
       )
+    } else if [.incomplete, .incompleteExpired].contains(stripeSubscription.status) {
+      Task {
```
mbrandonw, line 118: could do `@Dependency(\.fireAndForget)` here if we wanna get in the habit of doing things like that.

mbrandonw (approved):
nice!

Sources/PointFree/Subscribe.swift
mbrandonw: Coo coo, then let's skip!

PR #848, opened 2023-05-31
Better messaging for login errors

stephencelis:
A common login error occurs for users with multiple GitHub accounts associated with the same primary email address, either because they've deleted and recreated their GitHub account, or from work accounts clashing with home accounts, etc. This PR checks for this issue and surfaces a better error message to users.

### DIFF
--- a/Sources/PointFree/Auth.swift
+++ b/Sources/PointFree/Auth.swift
@@ -1,3 +1,4 @@
+import CustomDump
 import Dependencies
 import Either
 import Foundation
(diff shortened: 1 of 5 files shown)


Sources/PointFree/Auth.swift
```diff
@@ -82,94 +85,91 @@ public func fetchUser<A>(_ conn: Conn<StatusLineOpen, T2<Models.User.ID, A>>)
...
+      )
+    }
+  } catch {
+    await fireAndForget {
+      try await sendEmail(
+        to: adminEmails,
+        subject: "GitHub Auth Failed",
+        content: inj1(String(customDumping: error))
+      )
+    }
```
stephencelis, line 168: We'll no longer get admin emails for the duplicate email case, so it'll be on the user to reach out to us based off the flash message.

mbrandonw (approved):
awesome!

PR #858, opened 2023-07-31
Fix a typo in ep.243

Jager-yoo:
I corrected a simple typo. Look forward to the next episode! 😄


mbrandonw (approved):
Good catch, thanks!

PR #859, opened 2023-08-03
Fix logic around team invite code.

mbrandonw:


Tests/PointFreeTests/JoinMiddlewareTests.swift
```diff
@@ -258,7 +258,7 @@ class JoinMiddlewareIntegrationTests: LiveDatabaseTestCase {
       }
       $0.stripe.fetchSubscription = { _ in
         var stripeSubscription = Stripe.Subscription.mock
-        stripeSubscription.quantity = 2
+        stripeSubscription.quantity = 1
```
mbrandonw, line 261: This was wrong because we were starting in a state with 2 people already on the subscription. We want to start with just 1 and then add a 2nd person.

stephencelis (approved):

PR #866, opened 2023-09-05
Restore credit incrementing functionality

stephencelis:


Sources/PointFree/RunnerTasks/WelcomeEmails.swift
```diff
@@ -12,49 +12,60 @@ import Prelude
 import Styleguide
 import Views
 
-public func sendWelcomeEmails() -> EitherIO<Error, Prelude.Unit> {
+public func sendWelcomeEmails() async throws {
```
stephencelis, line 15: After getting a passing test I was finding that the complex `EitherIO`/`async` delay+retry hybrid logic when applied would occasionally double-increment the episode credit count. Rather than figure out why EitherIO was having this issue, I unwound the code to just be `async` and it fixed the test expectation.

Sources/PointFree/RunnerTasks/WelcomeEmails.swift
```diff
@@ -12,49 +12,60 @@ import Prelude
...
+          try await clock.sleep(for: .seconds(10))
+          try await sentEmails.append(send(email: email))
+        } catch {
+          do {
+            try await clock.sleep(for: .seconds(10))
+            try await sentEmails.append(send(email: email))
+          } catch {
+            try await clock.sleep(for: .seconds(10))
+            try await sentEmails.append(send(email: email))
+          }
```
stephencelis, line 46: This just recaptures the `retry` operator...we could prob write an `async`-specific one, but this was easy for now.

PR #889, opened 2023-12-19
Clean up blog posts and rss feed a bit.

mbrandonw:
Right now we show a "Tweet" button in blog posts, and I don't think it's really buying us much. So, getting rid of that, and simplify the RSS that is sent back to people for our blogs.


Sources/PointFree/Blog/BlogAtomFeed.swift
```diff
@@ -34,7 +35,7 @@ private func atomEntry(for post: BlogPost) -> AtomEntry {
   @Dependency(\.siteRouter) var siteRouter
 
   return AtomEntry(
-    content: blogPostContentView(post),
+    content: .p(.text(post.blurb)),
```
stephencelis, line 38: Should we render this as markdown?

PR #909, opened 2024-06-06
Allow accepting team invite when on inactive subscription

mbrandonw:
This came up in a support email today, but currently if one is on a "past due" subscription they are not allowed to accept a team invite.


Sources/PointFree/Account/Invite.swift
```diff
@@ -261,42 +261,26 @@ private func validateIsNot(currentUser: User) -> (User) -> EitherIO<Error, User>
...
 
-private func redirectCurrentSubscribers<A, B>(
-  _ middleware: @escaping M<T3<A, User?, B>>
-) -> M<T3<A, User?, B>> {
-  @Dependency(\.database) var database
-  @Dependency(\.stripe) var stripe
+func redirectCurrentSubscribers<A>(
+  _ middleware: @escaping M<A>
+) -> M<A> {
+  @Dependency(\.subscriberState) var subscriberState
```
mbrandonw, line 267: I think this helper was written before we had access to `subscriberState` in the dependency system, and so now the logic can be greatly simplified. No need to fetch subscriptions for the database/stripe since that has already been done for us.

Sources/PointFree/Enterprise.swift
```diff
@@ -357,41 +357,3 @@ extension EmailAddress {
     self.rawValue.lowercased().hasSuffix("@\(domain.rawValue.lowercased())")
   }
 }
-
-private func redirectCurrentSubscribers<Z>(
-  _ middleware: @escaping M<T2<User, Z>>
-) -> M<T2<User, Z>> {
```
mbrandonw, line 363: we had two versions of this helper that were slightly different for enterprise/non-enterprise flows, but I was able to consolidate them into one.

Tests/PointFreeTests/EnterpriseTests.swift
```diff
@@ -105,7 +105,6 @@ class EnterpriseTests: TestCase {
     await withDependencies {
       $0.database = .mock
       $0.database.fetchEnterpriseAccountForDomain = { _ in account }
-      $0.database.fetchSubscriptionById = { _ in throw unit }
```
mbrandonw, line 108: cleaned up some tests while I was here

Tests/PointFreeTests/EnterpriseTests.swift
```diff
@@ -300,7 +311,7 @@ class EnterpriseTests: TestCase {
       await assertSnapshot(matching: await siteMiddleware(conn), as: .conn)
     }
 
-    // todo: more verifications that subscription was linked
+    XCTAssertEqual(userAddedToSubscription.value, true)
```
mbrandonw, line 314: and added some assertions for todos we had.

stephencelis (approved):

PR #920, opened 2024-07-04
Update gifts index page.

mbrandonw:
This PR is based off of #919.

This updates just the index page for gifts. There are a few more gift related pages that will need to be updated, such as checkout, redemption, etc.


Sources/Views/Gifts/GiftsLanding.swift
```diff
@@ -6,78 +6,7 @@ import PointFreeDependencies
 import PointFreeRouter
 import Prelude
 
-public func giftsLanding(
```
mbrandonw, line 9: most of this file can be deleted, but there are still a few things that are used in other gifts-related pages. once we convert those we can delete all of this.

Sources/Views/GiftsV2.swift
```diff
@@ -0,0 +1,132 @@
+import Dependencies
+import StyleguideV2
+import TaggedMoney
+import Transcripts
+
+public struct GiftsV2: HTML {
```
mbrandonw, line 6: I'm using a lot from your `PricingV2` here, and I even moved a lot of the views from that page into the SharedComponents directory since they can be used directly in both screens.

Sources/Views/PricingV2.swift
```diff
@@ -120,118 +120,11 @@ public struct PricingV2: HTML {
     }
 
     WhatToExpect()
-    PricingFAQ()
+    FAQModule(faqs: .allFaqs)
```
mbrandonw, line 123: moved a bunch of stuff in here to the SharedComponents directory and generalized it a bit.

Sources/PointFree/Gifts/GiftsMiddleware.swift
```diff
@@ -17,8 +17,9 @@ public func giftsMiddleware(
       |> giftCreateMiddleware
 
   case .index:
-    return conn.map(const(()))
-      |> giftsIndexMiddleware
+    return IO {
```
stephencelis, line 20: Worth migrating the gifts middleware to be an `async` function?

Sources/Views/GiftsV2.swift
```diff
@@ -0,0 +1,128 @@
...
+          }
+          .attribute("href", siteRouter.path(for: .gifts(.plan(.sixMonths))))
+        }
+
+        Lane("1 year", annualPricePerMonth: 168) {
+          "One-time payment"
+        } features: {
+          li { "22% off the 3 and 6 month gift options" }
+            .color(.black)
+            .backgroundColor(.yellow)
```
stephencelis, line 50: Could make the highlight extend a bit and pop more:

```suggestion
            .backgroundColor(.yellow)
            .inlineStyle("margin", "-2px")
            .inlineStyle("padding", "2px")
```

stephencelis (approved):
Looks great!

PR #924, opened 2024-07-06
Update to clips pages to new styleguide

mbrandonw:
Haven't recorded snapshots yet, but it's mostly done.


Sources/Views/ClipsViews.swift
```diff
@@ -10,358 +10,149 @@ import PointFreePrelude
...
+    tag("pf-center-column") {
+      content
+        .inlineStyle("max-width", "1280px")
+        .inlineStyle("width", "100%")
+        .inlineStyle("margin", "0 auto")
+    }
+    .inlineStyle("width", "100%")
+    .inlineStyle("box-sizing", "border-box")
+    .inlineStyle("display", "block")
+  }
```
mbrandonw, line 109: Experimenting with a view component that does the minimal work to center some content in the screen with auto margins and max width 1280px. I think we could replace all instances of our hard coded 1280px with this.

Sources/Views/PageLayoutViewV2.swift
```diff
@@ -177,6 +177,7 @@ struct NavBar: HTML {
     .inlineStyle("width", "100%")
     .inlineStyle("position", "sticky", media: .mobile)
     .inlineStyle("top", "0", media: .mobile)
+    .inlineStyle("z-index", "9999")
```
mbrandonw, line 180: the video for clips (and episodes) is absolutely positioned and so we need this to make sure the nav renders over it.

stephencelis (approved):

PR #925, opened 2024-07-06
Use murmur hash to generate unique class names.

mbrandonw:
Taking inspiration from styled components in the react ecosystem where they generate unique class names using the encoded murmur hash of the style. This is used in development mode only so that class names don't shift around every time we added new styles to a page. Release will still use the shorter, index based naming convention.


Sources/StyleguideV2/HTML/HTMLInlineStyle.swift
```diff
@@ -1,4 +1,5 @@
 import ConcurrencyExtras
+import Darwin
```
stephencelis, line 2: This'll break on Linux, no?

mbrandonw: yeah, just have to figure out what to import to get access to `log` on linux.

stephencelis (approved):

PR #937, opened 2024-07-27
Fixed a typo 'momths' to 'months'

coenttb:
The page https://www.pointfree.co/gifts has a typo, which this PR fixes.


mbrandonw (approved):
Good catch, thank you @coenttb!

PR #947, opened 2024-09-26
Better handling of duplicate email when signing up

mbrandonw:
We haven't been detecting this situation correctly for awhile, so this fixes that, as well as does a few small refactors to set us up to better improve this area of the code base in the future.


Sources/PointFree/Auth.swift
```diff
@@ -122,7 +155,11 @@ private func registerUser(env: GitHubUserEnvelope) async throws -> Models.User {
       )
     }
     return user
-  } catch let PostgresError.server(error) where error.fields[.constraintName] == "users_email_key" {
+  } catch let error as PSQLError
+    where
+    error.serverInfo?[.constraintName] == "users_email_key"
+    && error.serverInfo?[.routine] == "_bt_check_unique"
+  {
```
mbrandonw, line 162: This is the correct way to detect the duplicate key SQL error.

stephencelis (approved):

PR #948, opened 2024-09-28
Update welcome emails to use new HTML library

mbrandonw:


Sources/PointFree/Admin/EmailPreview.swift
```diff
@@ -42,14 +42,13 @@ private func emailPreviewView(selectedTemplate: EmailTemplate?) -> Node {
         .fragment(options(selectedTemplate: selectedTemplate))
       )
     ),
-    .div(
+    .iframe(
```
mbrandonw, line 45: Better to render email previews in its own iframe so it's fully quarantined from the styles on this page.

Sources/PointFree/Emails/SendEmail.swift
```diff
@@ -91,6 +92,67 @@ public func prepareEmail(
   )
 }
 
+public func prepareEmailV2(
```
mbrandonw, line 95: Once we convert the rest of the emails to this new system we can get rid of `prepareEmail` in favor of this.

Sources/PointFree/Emails/WelcomeEmails.swift
```diff
@@ -0,0 +1,271 @@
+import Dependencies
+import Models
+import IssueReporting
+import PointFreeRouter
+import StyleguideV2
+
+struct WelcomeEmail<Content: HTML>: EmailDocument {
```
mbrandonw, line 7: This is a general "layout" for welcome emails, and could probably be generalized for all emails.

Sources/PointFree/Emails/WelcomeEmails.swift
```diff
@@ -0,0 +1,271 @@
...
+  ) {
+    self.content = content()
+    self.preheader = preheader
+    self.user = user
+  }
+
+  var body: some HTML {
+    span {
+      HTMLText(preheader)
+    }
```
mbrandonw, line 27: "Preheader" is a bit of hidden text you put at the top of emails that most email clients will render as the one-line preview in the inbox.

Sources/PointFree/RunnerTasks/SendWelcomeEmails.swift
```diff
@@ -0,0 +1,143 @@
+import Dependencies
```
mbrandonw, line 1: This file looks new in the diff but it's just me renaming an existing file.

Sources/StyleguideV2/HTMLHelpers/Link.swift
```diff
@@ -19,6 +19,7 @@ public struct Link<Label: HTML>: HTML {
   public var body: some HTML {
     a { label }
       .attribute("href", href)
+      .color(linkStyle.color)
```
mbrandonw, line 22: This is necessary to get Gmail to recognize link color, but it also causes a lot of churn in our snapshot tests. We could think about ways to limit this only to email html, but not sure how we want to do that.

Sources/PointFree/Admin/EmailPreview.swift
```diff
@@ -74,11 +73,32 @@ private func email(selectedTemplate: EmailTemplate) -> Node {
       newPricing: Pricing(billing: .yearly, quantity: 7)
     )
   case .welcomeEmail1:
-    return welcomeEmail1Content(user: blob)
+    return Node {
+      WelcomeEmail(
+        preheader: "This is the preheader",
```
stephencelis, line 78: - [ ] TODO

Sources/PointFree/Emails/SendEmail.swift
```diff
@@ -91,6 +92,67 @@ public func prepareEmail(
...
+  unsubscribeData
+    .map { userId, newsletter in
+      guard
+        let unsubEmail = mailgun.unsubscribeEmail(
+          fromUserId: userId, andNewsletter: newsletter),
+        let unsubUrl = (try? ExpressUnsubscribe().print((userId, newsletter)))
+          .flatMap({ Encrypted(String($0), with: envVars.appSecret) })
+          .map({ siteRouter.url(for: .expressUnsubscribe(payload: $0)) })
+      else {
+        logger.log(.error, "Failed to generate unsubscribe link for user \(userId)")
```
stephencelis, line 124: Should this use `reportIssue`?

Sources/StyleguideV2/HTML/HTMLPrinter.swift
```diff
@@ -31,19 +31,25 @@ public struct HTMLPrinter {
...
+          sheet.append("\(className){\(style)}")
+        }
         sheet.append(configuration.newline)
       }
     }
     return sheet
   }
 
   public struct Configuration {
+    let forceImportant: Bool
```
stephencelis, line 46: We could consider calling this `isHTMLEmail` or `isEmail` instead to document why, but we also know we wanna reconsider the HTML library design with more use and after the dust settles.

stephencelis (approved):

PR #949, opened 2024-09-30
Login failure landing

mbrandonw:
When we detect a GitHub login failure we now take the user to a dedicated page where they can fix the problem themselves, or contact us.


Sources/GitHub/Model.swift
```diff
@@ -2,19 +2,37 @@ import EmailAddress
 import Foundation
 import Tagged
 
-public struct AccessToken: Codable {
+public struct AccessToken: Codable, Equatable, RawRepresentable {
```
mbrandonw, line 5: -[] separate this into a tagged string (`AccessToken`) and a response payload (`AccessTokenResponse`).

Sources/PointFree/Auth.swift
```diff
@@ -39,48 +45,203 @@ func authMiddleware(
...
+      gitHubUserID: newGitHubUser.id,
+      githubAccessToken: accessToken
+    )
+    await fireAndForget {
+      let email = try await gitHub
+        .fetchEmails(accessToken: AccessToken(accessToken: existingAccessToken))
+        .first(where: \.primary)
+        .unwrap()
+        .email
+      let html = String(decoding: Data(), as: UTF8.self)
```
mbrandonw, line 93: - [ ] Use the actual email view.

Sources/PointFree/Auth.swift
```diff
@@ -39,48 +45,203 @@ func authMiddleware(
...
+            to: [email],
+            subject: "Your GitHub account has been updated",
+            text: html,
+            html: html,
+            domain: mgDomain
+          )
+        )
+      } catch {
+        print(error)
+        print("!!!")
```
mbrandonw, line 107: - [ ] use issue reporter

Sources/PointFree/Session.swift
```diff
@@ -87,11 +87,14 @@ extension URLRequest {
...
 
+import GitHub
+
 public struct Session: Equatable {
   public var flash: Flash?
+  public var gitHubAccessToken: AccessToken?
   public var user: User?
 
-  public static let empty = Session(flash: nil, user: nil)
+  public static let empty = Session()
```
stephencelis, line 97: Ha is this even used?

stephencelis (approved):
Looks good! Just that one TODO if you think it's worth knocking out before merging.

PR #950, opened 2024-10-11
Don't fail login if Stripe is having problems.

mbrandonw:
Yesterday log in was broken for awhile on Point-Free because the Stripe API was having problems, and we refresh the user's subscription state when they log in. Let's not prevent log in if Stripe is having problems. Also let's refresh the user's subscription state when they visit their account page.

### DIFF
--- a/Sources/PointFree/Account/Account.swift
+++ b/Sources/PointFree/Account/Account.swift
@@ -32,106 +32,73 @@ private func fetchAccountData<I>(
 ) -> IO<Conn<I, AccountData>> {
   @Dependency(\.database) var database
   @Dependency(\.stripe) var stripe
+  @Dependency(\.fireAndForget) var fireAndForget
 
   let (user, subscriberState) = lower(conn.data)
 
-  let subscription = EitherIO {
-    try await database.fetchSubscription(user: user)
-  }
-
-  let owner =
-    subscription
-    .flatMap { subscription in
-      EitherIO { try await database.fetchUser(id: subscription.userId) }
+  return IO { () -> AccountData in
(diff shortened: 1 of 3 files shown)


Sources/PointFree/Account/Account.swift
```diff
@@ -32,106 +32,73 @@ private func fetchAccountData<I>(
...
 
-  let subscription = EitherIO {
-    try await database.fetchSubscription(user: user)
-  }
-
-  let owner =
-    subscription
-    .flatMap { subscription in
-      EitherIO { try await database.fetchUser(id: subscription.userId) }
+  return IO { () -> AccountData in
```
mbrandonw, line 39: I refactored the internals of `fetchAccountData` to use async/await, but didn't refactor the signature of it to be modern because it's still plugged into a legacy thing.

Sources/PointFree/Auth.swift
```diff
@@ -359,7 +359,9 @@ private func gitHubAuthTokenMiddleware(
   do {
     let gitHubUser = try await gitHub.fetchUser(accessToken)
     let user = try await fetchOrRegisterUser(accessToken: accessToken, gitHubUser: gitHubUser)
-    try await refreshStripeSubscription(for: user)
+    await notifyError("GitHub Auth: Refresh stripe failed") {
+      try await refreshStripeSubscription(for: user)
+    }
```
mbrandonw, line 364: We will still notify us when there is a stripe problem on login, we just won't fail login.

stephencelis (approved):
Looks good! Should we get a test in for this?

PR #964, opened 2025-05-01
Restore Redacted Text in Transcription

antigluten:
The [episode page](https://www.pointfree.co/episodes/ep115-redacted-swiftui-the-problem) has sections that contain █████ (redacted blocks). I have restored these sections to their original, readable text.

Question: Was the original redaction in text intentional?


mbrandonw:
Hi @antigluten, yes this was intentional.

PR #1051, opened 2026-07-09
Add Vorwerk to the list of companies trusting PointFree.

Myrronth:
Added Vorwerk to the list of companies that are using PointFree's products.

Here's what the logo looks like:
Vorwerk


Myrronth:
From the bot explaining the failing job:
> The failing CI job looks unrelated to this change.
It fails in `.github/workflows/ci.yml` when checking out `pointfreeco/episode-transcripts` with `secrets.GH_PAT`, and that secret appears unavailable for forked PRs.
Since this PR comes from a fork, I don’t think I can fix that from my side. A possible fix is to guard that step for forked PRs or use a token that’s available in this context.

stephencelis (approved):
Thanks! And thanks for your support :smile:

PR #1054, opened 2026-07-11
Support logging into Point-Free with only email.

mbrandonw:


Sources/Database/Database.swift
```diff
@@ -175,8 +179,8 @@ public struct Client {
   }
 
   public func registerUser(
-    accessToken: GitHubAccessToken,
-    gitHubUser: GitHubUser,
+    accessToken: GitHubAccessToken?,
+    gitHubUser: GitHubUser?,
```
stephencelis, line 183: Should these be bundled up into single `GitHubResponse` type?

Looks like there is a `GitHub` type that was added to this PR, so could use it here?

Sources/PointFree/Auth.swift
```diff
@@ -26,23 +26,137 @@ func authMiddleware(
...
+          "A code was recently sent to this email. Please wait a minute before requesting another."
+        )
+      }
+    }
+    await fireAndForget {
+      let html = String(
+        decoding: LoginCodeEmail(loginCode: loginCode).render(),
+        as: UTF8.self
+      )
+      await fireAndForget {
```
stephencelis: Is this double-nesting needed?

PR #1058, opened 2026-08-05
Fix typos in Newsletter-220.md

loganmoseley:
Hey there. The JSON tools sound powerful. I noticed two typos while reading, so here ya are.


mbrandonw (approved):
Thanks!

